PR #68, opened 2013-06-11
Allow behaviors to be any control type

montegoulding:


montegoulding:
closing because the idea was rejected

PR #81, opened 2013-07-03
[[ Bug 11018 ]] prevent crash when opening a rotated referenced image wi...

livecodeian:
...th an invalid filename


livecodeian:
merged into release-6.1.1 branch

PR #91, opened 2013-07-26
Bugfix-11020: empty handler caused crash on server edition

livecodesebastien:


livecodeian:

PR #106, opened 2013-08-02
[[ Bugfix 11087 ]] don't try to create transformed image rep if there is...

livecodeian:
... no source rep


livecodeian:

PR #164, opened 2013-09-02
Refactor syntax unicode button menu

livecodefraser:


engine/src/button.cpp
```diff
@@ -3258,28 +3227,20 @@ class ButtonMenuCallback : public IParseMenuCallback
...
 #ifndef TARGET_PLATFORM_MACOS_X
-		const char *t_tabptr = p_menuitem->label.getstring();
-		uint4 t_labelsize = p_menuitem->label.getlength();
-		// replace any tabs, as these don't display in button menus
-		while (MCU_strchr(t_tabptr, t_labelsize, '\t', p_menuitem->is_unicode))
-		{
-			MCU_copychar(' ', (char*)t_tabptr, p_menuitem->is_unicode);
-		}
+		MCStringRef t_label;
+		/* UNCHECKED */ MCStringMutableCopy(p_menuitem.label, t_label);
```
livecodesebastien: Should be `p_menuitem -> label`

livecodesebastien:
The branch is broken for now - the menu items don't work anymore: clicking on them doesn't trigger anything. The error might come from refactor-syntax_unicode_button, which has glitches in the text width computation (obvious in the startup screen), and, on Linux and Windows, has menu items carrying extra characters.

livecodesebastien:
When exiting the IDE, the branch crashes on Linux, and an `MCAssert()` fails on Windows.
It doesn't compile on Android or iOS.

livecodesebastien:
Still crashes on Exit on Linux. The bug is likely to come from refactor-syntax_unicode_button_menu, as it works fine in refactor-syntax_unicode_button

PR #170, opened 2013-09-03
Refactor syntax unicode kupdown

livecodefraser:
Note: this branch depends on refactor-syntax_unicode_button_menu


livecodesebastien:

engine/src/stacklst.cpp
```diff
@@ -294,12 +295,19 @@ Boolean MCStacklist::doaccelerator(KeySym key)
...
+	if (key < 0x7f)
+		t_char = key;
+	else if ((key & XK_Class_mask) == XK_Class_codepoint)
+		t_char = key & XK_Codepoint_mask;
+
 	uint2 t_mod_mask = MS_CONTROL | MS_MAC_CONTROL | MS_MOD1;
 
-	if ((key >= 'a' && key <= 'z') || (key & 0xFF00) == 0xFF00)
+	// The shift state is important for lower-case letters and special keys
+	if (iswlower(t_char) || t_char == 0)
```
livecodesebastien, line 308: This doesn't compile on Linux - needs either to keep the previous comparison or add `#include <wctype.h>`

livecodesebastien:
Can you revert the change to the `.ico` files?

livecodesebastien:
The branch is currently buggy on Windows, Mac, and Linux as the message `keyDown` is never caught by scripts; if you create a new stack and set the stack's script to

```
on keyDown theKey
    answer theKey
end keyDown
```

nothing happens when you press a key

livecodesebastien:
Could you revert your last commit, and merge refactor-syntax_unicode in instead? It has the implementation of MCVariable::copyasvalueref()

livecodesebastien:
Can you merge again refactor-syntax_unicode in? There was a compilation issue on Windows which is now fixed

livecodesebastien:
There is also a compilation error on iOS in `mobileiphonedc.mm`

PR #191, opened 2013-09-06
Refactor syntax unicode regex

livecodepanos:


engine/src/exec-strings.cpp
```diff
@@ -380,8 +380,11 @@ void MCStringsEvalReplaceText(MCExecContext& ctxt, MCStringRef p_string, MCStrin
     
     uindex_t t_source_length = MCStringGetLength(p_string);
     uindex_t t_source_offset = 0;
+	MCStringRef t_head, t_tail;
+	t_head = nil;
+	t_tail = nil;
     
-    while (t_success && t_source_offset < t_source_length && MCR_exec(t_compiled, MCStringGetCString(p_string) + t_source_offset, t_source_length - t_source_offset))
+    while (t_success && t_source_offset < t_source_length && MCR_exec(t_compiled, t_tail, t_source_length - t_source_offset) && MCStringDivideAtIndex(p_string, t_source_offset, t_head, t_tail))
```
livecodesebastien: Attention, C++ boolean expressions are evaluated from left to right, so that `MCR_exec()` is executed before `MCStringDivideAtIndex()`. For now, using the Livecode function `replacetext()` crashes because `t_tail` is obviously a null pointer.
Also, it might be better to use `MCStringCopySubstring()` to create the substring, instead of `MCStringDivideAtIndex()` since the head is never used.

engine/src/exec-strings.cpp
```diff
@@ -381,8 +381,10 @@ void MCStringsEvalReplaceText(MCExecContext& ctxt, MCStringRef p_string, MCStrin
     
     uindex_t t_source_length = MCStringGetLength(p_string);
     uindex_t t_source_offset = 0;
+	MCStringRef t_substring;
+	t_substring = nil;
     
-    while (t_success && t_source_offset < t_source_length && MCR_exec(t_compiled, MCStringGetCString(p_string) + t_source_offset, t_source_length - t_source_offset))
+    while (t_success && t_source_offset < t_source_length && MCStringCopySubstring(p_string, MCRangeMake(t_source_offset + 1, MCStringGetLength(p_string) - (t_source_offset + 1)), t_substring) && MCR_exec(t_compiled, t_substring, t_source_length - t_source_offset))
```
livecodesebastien: Why `t_source_offset + 1`? It doesn't work, but do the job properly with `t_source_offset`

engine/src/regex.cpp
```diff
@@ -149,7 +149,7 @@
 *************************************************/
 
 size_t
-regerror(int errcode, const regex_t *preg, char *errbuf, size_t errbuf_size)
+regerror(int errcode, const regex_t *preg, MCStringRef errbuf, size_t errbuf_size)
```
livecodesebastien: If you update this method, you need to update as well `regexperror` to an `MCStringRef`. And be careful, `errbuf` is a return parameter. Plus, you can't use `MCStringAppendNativeChars()` on a non-mutable `MCStringRef`, which is the default when you create it.

engine/src/regex.cpp
```diff
@@ -327,7 +329,9 @@ int MCR_exec(regexp *prog, const char *string, uint4 len)
 		{
 			return (0);
 		}
-		regerror(status, NULL, regexperror, sizeof(regexperror));
+		MCAutoStringRef t_regexperror;
+		/* UNCHECKED */ MCStringCreateWithCString(regexperror, &t_regexperror);
+		regerror(status, NULL, *t_regexperror, sizeof(regexperror));
```
livecodesebastien: Here, I think you misunderstood how the function works, it creates an error message from the `status`, which is returned through the 3rd parameter. Also, instead of creating a temporary variable, you should release `regexperror` (updated to an `MCStringRef`) and pass it to `regerror()` as it was done before.
Also, you only need to pass the size of the 3rd param of `regerror` if it's an array. If you use an `MCStringRef`, you don't need it - `MCStringRef`s have no size when they are created.

PR #201, opened 2013-09-09
Refactor syntax unicode fetch

livecodefraser:


engine/src/deploy_linux.cpp
```diff
@@ -820,7 +820,7 @@ Exec_stat MCDeployToELF(const MCDeployParameters& p_params, bool p_is_android)
 	t_path = nil;
 #if defined(_MACOSX)
 	if (t_success)
-		t_success = MCCStringFromNative(p_params . output, t_path);
+		t_success = MCCStringFromNative((const char *)MCStringGetNativeCharPtr(p_params . output), t_path);
 #else
 	t_path = p_params . output;
```
livecodesebastien: Needs to be updated, only compile on Mac for now

engine/src/deploy.h
```diff
@@ -24,49 +26,104 @@ struct MCDeployParameters
...
 	// 4K into. It is designed for use on OS X to reduce standalone file size.
-	char *spill;
+	MCStringRef spill;
 
 	// The output path for the new executable.
-	char *output;
+	MCStringRef output;
+	
+	
+	MCDeployParameters()
```
livecodesebastien, line 74: In a general way, when there is no difference between being nil or empty for an `MC<type>Ref` variable, it should be initialized to `kMCEmpty<type>`, as it makes the release easier - and no need to test whether it's `nil` before checking anything else.
A function like `bool MCDeployParameters::InitWithArray(MCArrayRef)` might come handy to set all the struct fields.
This comment also applies for the other structs whose fields have been updated to `MCStringRef` in that branch

PR #204, opened 2013-09-09
Refactor syntax unicode srv multipart 2

livecodepanos:


engine/src/srvcgi.cpp
```diff
@@ -953,10 +957,10 @@ static bool cgi_multipart_get_boundary(char *&r_boundary)
 	{
 		for (uint32_t i = 0; i < t_param_count; i++)
 		{
-			if (MCCStringEqualCaseless(t_names[i], "boundary") && MCCStringLength(t_values[i]) > 0)
+			if (MCStringIsEqualToCString(t_names[i], "boundary", kMCCompareCaseless) && MCStringGetLength(t_values[i]) > 0)
 			{
-				r_boundary = t_values[i];
-				t_values[i] = NULL;
+				r_boundary = strdup(MCStringGetCString(t_values[i]));
+				t_values[i] = nil;
```
livecodesebastien: You can't nullify an `MCStringRef`, to delete it you want to use `MCValueRelease`. 
Anyway here, it was done like this to avoid a string copy: `r_boundary` is made to point to the memory of `t_values[i]`, and then `t_values[i]` is set to NULL so that it won't get `free`'d during the memory release in the final loop of the function.
You merely have to update the signature to having `MCStringRef` and using `MCValueRetain` there, or keep the `strdup(...)` and remove the `nil` assignment.

engine/src/srvmultipart.cpp
```diff
@@ -278,33 +279,38 @@ static inline bool is_qtext(char p_char)
 	return (p_char != '"') && (p_char != '\\') && (p_char != '\r');
 }
 
-bool read_quoted_string(const char *p_header, const char *&r_string_end, char *&r_string)
+bool read_quoted_string(MCStringRef p_header, MCStringRef &r_string_end, MCStringRef &r_string)
```
livecodesebastien: This function needs more thought: what the function does is reading character by character the input string, from a quote to another quote, building the output string (`r_string`), and returning the remaining part at the end. Actually what you need to do is merely to keep the same idea but using counter to keep your index in the input string. In the end you can create `r_string` from `t_string`, and `r_end_string` is the substring of `p_string`, from the counter + 1 to the end.

engine/src/srvmultipart.cpp
```diff
@@ -355,8 +361,9 @@ bool MCMultiPartParseHeaderParams(const char *p_params, char **&r_names, char **
 		if (t_success)
 		{
 			consume_whitespace(p_params);
-			t_has_value = p_params[0] == '=';
-			p_params++;
+			t_has_value = (MCStringGetNativeCharAtIndex(p_params, 0) == '=');
+			MCAutoStringRef t_head;
+			/* UNCHEKED */ MCStringDivideAtIndex(p_params, 0, &t_head, p_params);
```
livecodesebastien: You can't give the same `MCStringRef` to `MCStringDivideAtIndex()`, because in the end you would lose the reference to which `p_params` pointed beforehand; you need a temporary value, and `MCValueAssign()` it back after the call.

engine/src/srvmultipart.cpp
```diff
@@ -207,68 +207,69 @@ bool is_token_char(char p_char)
...
-		x_header += 3;
+		MCStringDivideAtIndex(x_header, 2, &t_head, x_header);
 		return true;
 	}
 	
 	return false;
 }
 
-void consume_whitespace(const char *&x_header)
+void consume_whitespace(MCStringRef &x_header)
```
livecodesebastien: This should either have two parameters (`in`, `out`) or use a temporary value, and `MCValueAssign()` that temporary value to the `in/out` parameter to avoid losing references to it.

engine/src/srvmultipart.cpp
```diff
@@ -207,68 +207,69 @@ bool is_token_char(char p_char)
...
-static bool skip_whitespace(const char *&x_header)
+static bool skip_whitespace(MCStringRef &x_header)
 {
-	if (is_whitespace(*x_header))
-	{
-		x_header++;
+	MCAutoStringRef t_head;
+	if (is_whitespace(MCStringGetNativeCharAtIndex(x_header, 0)))
+	{	
+		MCStringDivideAtIndex(x_header, 0, &t_head, x_header);
```
livecodesebastien: Same problem, you can't have the same `MCStringRef` as input and output for `MCStringDivideAtIndex()`, you need to use a temporary value. And same as for `consume_whitespace()`, the signature should change to an `in/out` param which is released and assigned at the end of the function, or `in` and `out` parameters.

engine/src/srvmultipart.cpp
```diff
@@ -278,33 +281,43 @@ static inline bool is_qtext(char p_char)
...
-	if (p_header[0] != '"')
+	if (MCStringGetNativeCharAtIndex(p_header, 0) != '"')
 		return false;
 	
-	p_header++;
-	while (t_success && p_header[0] != '"')
+	//MCAutoStringRef t_head;
+	//MCStringDivideAtIndex(p_header, 0, &t_head, p_header);
+	t_pos += 1;
+	while (t_success && MCStringGetNativeCharAtIndex(p_header, 0) != '"')
```
livecodesebastien: You need to check the char at `t_pos`, not at 0 (it was done like this because `p_header` is increased in each loop, currently you don't look at the current position but always the first character)

engine/src/srvmultipart.cpp
```diff
@@ -320,8 +333,9 @@ bool read_quoted_string(const char *p_header, const char *&r_string_end, char *&
...
 		if (t_string != NULL)
 			t_string[t_strlen] = '\0';
-		r_string = t_string;
-		r_string_end = p_header + 1;
+
+		/* UNCHECKED */ MCStringCreateWithCString(t_string, r_string);
+		/* UNCHECKED */ MCStringCopySubstring(p_header, MCRangeMake(t_pos +1 , MCStringGetLength(p_header) - (t_pos + 1)), r_string_end);
 	}
 	else
 		MCCStringFree(t_string);
```
livecodesebastien, line 329: You need anyway to release the `t_string`, as you copy it to `r_string`, and not just give the pointer as it was done before.

engine/src/srvmultipart.cpp
```diff
@@ -278,33 +278,40 @@ static inline bool is_qtext(char p_char)
...
+	while (t_success && MCStringGetNativeCharAtIndex(p_header, t_pos) != '"')
 	{
 		char t_char;
-		if (is_folded_lwsp(p_header))
+		if (is_folded_lwsp(p_header, t_pos))
 		{
-			t_char = p_header[2];
-			p_header += 3;
+			t_char = MCStringGetNativeCharAtIndex(p_header, t_pos + 2);
+			//MCAutoStringRef t_head;
```
livecodesebastien: Can you remove all the commented out code?

engine/src/srvmultipart.cpp
```diff
@@ -329,49 +337,47 @@ bool read_quoted_string(const char *p_header, const char *&r_string_end, char *&
 	return t_success;
 }
 
-bool MCMultiPartParseHeaderParams(const char *p_params, char **&r_names, char **&r_values, uint32_t &r_param_count)
+bool MCMultiPartParseHeaderParams(MCStringRef p_params, MCStringRef *&r_names, MCStringRef *&r_values, uint32_t &r_param_count)
 {
 	bool t_success = true;
 	
-	const char *t_next_param = NULL;
```
livecodesebastien, line 336: This is now useless

livecodesebastien, line 336: Wrong code line...

engine/src/srvmultipart.cpp
```diff
@@ -379,57 +385,62 @@ bool MCMultiPartParseHeaderParams(const char *p_params, char **&r_names, char **
...
 {
-	MCMultiPartHeader t_header = {NULL, NULL, 0, NULL, NULL};
+	MCMultiPartHeader t_header = {nil, nil, 0, nil, nil};
 	
 	bool t_success = true;
 	
 	uint32_t t_index = 0;
-	char *t_value = NULL;
-	char *t_name = NULL;
 	const char *t_next_param = NULL;
```
livecodesebastien: This one should be cleaned away.

engine/src/srvcgi.cpp
livecodesebastien: Could you remove this line? It's not needed anymore as now the value of `t_value[i]` is copied, and the MCStringRef will be released in the end.

PR #211, opened 2013-09-10
[[ Bug 10634 ]] Script editor "go to definition" does not find handlers ...

livecodemichael:
...in behaviors or library stacks.


engine/src/mode_development.cpp
```diff
@@ -630,80 +709,23 @@ Exec_stat MCObject::mode_getprop(uint4 parid, Properties which, MCExecPoint &ep,
 		}
 		else
 		{
+			// MM-2013-09-10: [[ Bug 10634 ]] Make we search both parent scripts and library stacks for handlers.
```
jacobkiers, line 712: Hi, it's a small thing, but shouldn't that comment start with: "Make **sure** we search..."?

PR #214, opened 2013-09-11
Refactor syntax unicode datetime

livecodefraser:


engine/src/date.cpp
```diff
@@ -1348,21 +1314,28 @@ bool MCD_convert_from_datetime(MCExecPoint &p_context, Convert_form p_primary_to
...
+			datetime_format(t_locale, *t_date_format, p_datetime, t_buffer_secondary);
+			
+			MCStringRef t_new;
+			/* UNCHECKED */ MCStringFormat(t_new, "%@ %@", t_buffer, t_buffer_secondary);
+			MCValueRelease(t_buffer);
+			MCValueRelease(t_buffer_secondary);
+			t_buffer = t_new;
 		}
+		
+		p_context . setvalueref(t_buffer);
```
livecodesebastien, line 1338: `t_buffer` should be released at the end of the function.

livecodesebastien:
Updating the `char*` to `MCStringRef` of `MCDateTime` also affects platform-specific files/functions:
- `cache_locale()` in sysunxdate.cpp
-  `windows_cache_locale()` in w32date.cpp
  which therefore need to be updated as well - Linux and Windows engine won't compile otherwise.
  Also, the update breaks the parsing: `put ("Wednesday, September 18 2013" is a date)` returns False in the new version, but used to return True.

PR #217, opened 2013-09-11
Refactor syntax unicode paragraph

livecodepanos:


engine/src/fieldf.cpp
```diff
@@ -2267,19 +2269,19 @@ void MCField::setupentry(MCButton *bptr, const MCString &s, Boolean isunicode)
...
 	if (MCactivefield == this)
 		unselect(False, True);
-	if (newtext.getlength() < MAX_PASTE_MESSAGES)
+	if (MCStringGetLength(newtext) < MAX_PASTE_MESSAGES)
 	{
 		char string[2];
 		string[1] = '\0';
-		char *sptr = (char *)newtext.getstring();
-		const char *eptr = sptr + newtext.getlength();
+		char *sptr = (char *)MCStringGetCString(newtext);
```
livecodesebastien: You can't cast the result of MCStringGetCString() as a char_: if it returns a const char_, it's because the string referenced by the MCStringRef shall never be modified but by MCString\* functions - which ensures the MCStringRef is mutable before executing any modification on it; otherwise, it would change the content of all the references to that string. What you want to do instead is to create a MCNativeCharArray from the MCStringRef, modify that array in the loop, and then create a new MCStringRef from it after the loop.

engine/src/fieldf.cpp
```diff
@@ -2310,7 +2312,7 @@ void MCField::typetext(const MCString &newtext)
 	}
 	else
 		updateparagraph(True, False);
-	delete (char *)newtext.getstring();
+	delete (char *)MCStringGetCString(newtext);
```
livecodesebastien: Same here, you should not delete the buffer of an MCStringRef.

engine/src/fieldf.cpp
livecodepanos: Is it ok if instead of the (char *) cast, I do :

char *sptr = strdup(MCStringGetCString(newtext))   ??

In that way, the string returned by MCStringGetCString() method will not be modified, right?

livecodesebastien: Sorry for the delayed feedback, so no, you can't do that, as modifying `sptr` in that context actually modifies `newtext` as well since it's a pointer to it's internal string. You really need to create a new `MCStringRef` and to use it later in the function to take that code in account.

engine/src/paragraf.cpp
```diff
@@ -1828,7 +1830,9 @@ Boolean MCParagraph::finsertnew(const MCString& p_text, bool p_is_unicode)
 		{
 			// We didn't find a line-break, so insert the string into the current paragraph and
 			// we must be done.
-			t_paragraph -> finsertnobreak(MCString(t_start, t_length), p_is_unicode);
+            MCAutoStringRef t_text;
+            /* UNCHECKED */ MCStringCreateWithNativeChars((const char_t *)t_start, t_finish - t_start, &t_text);
```
livecodesebastien: The bug which prevents any typing in the fields comes from here, the length of the string created should be `t_length` and not `t_finish - t_start` (which is always `0` in the `else` statement since `t_finish == t_start`)

livecodepanos:
In fact msort() doesn't work on Windows as well.

livecodesebastien:
Indeed, `mcsort()` is still broken. Now crashes on Mac on launch as well

livecodepanos:
Still buggy. revOnline Browser does not work properly (Click User Samples --> etc..)

PR #220, opened 2013-09-11
Refactor syntax unicode additem

livecodepanos:


engine/src/cmds.cpp
```diff
@@ -2420,23 +2420,23 @@ Exec_stat MCSort::sort_container(MCExecPoint &p_exec_point, Chunk_term p_type, S
...
 #endif /* MCSort::sort_container */
 
-void MCSort::additem(MCExecPoint &ep, MCSortnode *&items, uint4 &nitems, Sort_type form, MCString &s, MCExpression *by)
+void MCSort::additem(MCExecPoint &ep, MCSortnode *&items, uint4 &nitems, Sort_type form, MCStringRef s, MCExpression *by)
 {
 	if (by != NULL)
 	{
 		MCerrorlock++;
-		ep.setsvalue(s);
+		ep.setsvalue(MCStringGetOldString(s));
```
livecodesebastien: Use ep.setvalueref() instead.

engine/src/cmds.cpp
```diff
@@ -2420,23 +2420,23 @@ Exec_stat MCSort::sort_container(MCExecPoint &p_exec_point, Chunk_term p_type, S
...
 		MCeach->set(ep);
 		if (by->eval(ep) == ES_NORMAL)
-			s = ep.getsvalue();
+			ep . copyasstringref(s);
 		else
-			s = MCnullmcstring;
+			s = MCValueRetain(kMCEmptyString);
```
livecodesebastien: You should drop the reference with MCValueRelease before assigning a new value to an MCStringRef, so it would be better to create a temporary MCStringRef which either gets a value from  `<ep>` or kMCEmptyString if by != NULL, or gets a copy of the MCStringRef `<s>` otherwise.

engine/src/cmds.cpp
```diff
@@ -2463,32 +2463,33 @@ void MCSort::additem(MCExecPoint &ep, MCSortnode *&items, uint4 &nitems, Sort_ty
 				sptr++;
 				length--;
 			}
-			s.setlength(s.getlength() - length);
+            //Is the line below now needed? If yes, how to update s.setlength?
+			//s.setlength(s.getlength()) - length);
```
livecodesebastien: Here, you need to get the position of the number in the string. Use countersto get the boundaries of the number, and then with MCStringGetSubstring(), create the MCStringRef you'll pass to MCU_stor8

livecodesebastien:
For now the branch `additem` is buggy: the `sort` function is no longer working properly.
A simple script like the one following fails to produce the expected result if pasted and executed in the multiline message box (Ctrl+Enter to execute the script)
-- script
put "1,4,2" into t_array
sort items of t_array
put t_array
-- expected result: 1,2,4
-- end of script

livecodesebastien:
Actually the `sort` function seems to be broken as well on refactor-syntax_unicode, so there is no need to investigate on this problem in that branch.

livecodesebastien:
Actually there is more work to do on that one. The EP passed to `additem()` should definetely disappear. That can be done quite easily now that refactor-syntax_unicode_datetime has been merge in refactor-syntax_unicode, because it allows you to call `MCD_convert()` without EP, which is what you want!

PR #226, opened 2013-09-11
Refactor syntax unicode filter

livecodepanos:


engine/src/cmdsf.cpp
```diff
@@ -1787,16 +1800,20 @@ Boolean MCFilter::match(char *s, char *p, Boolean casesensitive)
...
-	while (*p == '*')
-		p++;
-	return *p == 0;
+	while (MCStringGetNativeCharAtIndex(p, 0) == '*')
+		/* UNCHECKED */ MCStringDivideAtIndex(p, i_p++, t_head_p, p);
+    
+    MCValueRelease(t_head_p);
+    MCValueRelease(t_head_s);
+    
+	return MCStringGetNativeCharAtIndex(p, 0) == 0;
```
livecodesebastien: The whole function needs more thoughts:
- you first don't want to divide the strings! The function just compare them
- the pointer arithmetic formerly used can't apply anymore, so you need to have counters to retain your position in the strings (i.e, `s_pos` and `p_pos`). You can then use MCStringGetNativeCharAtIndex() to retrieve the characters of the strings (and each time there is a `p++` or `s++`, instead of `*s` or `*p`. The appropriate counter shall be increased after the statement.
- `while (*s)`'s equivalent would be `while(s_pos < MCStringGetLength(s))`: you want to loop through the whole string, which ends with a NULL character (in which case, `*s` is NULL).
- for the recursive call of `match()` (l1785 of your file), you should create a substring from `p` and `s`, with a starting point at respectively `p_pos` and `s_pos`.

engine/src/cmdsf.cpp
```diff
@@ -1840,7 +1857,9 @@ char *MCFilter::filterlines(char *sstring, char *pstring,
 			*t_return = '\0';
 
 		line = t_string;
-		if (match(line, pstring, casesensitive) != out)
+        MCAutoStringRef t_line;
+        /* UNCHECKED */ MCStringCreateWithCString(line, &t_line);
+		if (match(*t_line, pstring, casesensitive) != out)
```
livecodesebastien: Same here, this is more subtle than it appears: `strchr()` should be replaced with `MCStringDivideAtChar(t_string)`, and then you should call `match()` on the head, and at the end of the loop, assign the tail to `t_string`. The loop then would break after tail is an empty string.

PR #228, opened 2013-09-12
Refactor syntax unicode chunk

livecodepanos:


engine/src/mbldialog.cpp
```diff
@@ -183,17 +183,17 @@ bool StringArrayToString (const_cstring_array_t *p_indexes, char *&r_result, MCC
     return t_success;
 }
 
-static bool contains_char(const char *p_string, char p_char)
+static bool contains_char(MCStringRef p_string, char p_char)
 {
-	return p_char != '\0' && strchr(p_string, p_char) != nil;
+	return p_char != '\0' && strchr(MCStringGetCString(p_string), p_char) != nil;
```
livecodesebastien: Use `MCStringFirstIndexOfChar()` instead of `strchr()`. In a general way, you should never use `strchr()` combined with `MCStringGetCString()`

engine/src/mbldialog.cpp
```diff
@@ -238,7 +238,11 @@ bool StringTokenize(const char *p_string, const char *p_delimiters, char**& r_el
 			// Add a new element to the array, and clone the substring
 			t_success = MCMemoryResizeArray(t_element_count + 1, t_elements, t_element_count);
 			if (t_success)
-				t_success = MCCStringCloneSubstring(t_token, t_ptr - t_token, t_elements[t_element_count - 1]);
+            {
+                MCAutoStringRef t_token_str;
+                /* UNCHECKED */ MCStringCreateWithCString(t_token, &t_token_str);
+				t_success = MCStringCopySubstring(*t_token_str, MCRangeMake(0, t_ptr - t_token), t_elements[t_element_count - 1]);
+            }
```
livecodesebastien, line 244: Same as for MCFilter::match.
You want better to use counters and `MCStringGetNativeCharAt()` instead of all the pointer arithmetic, and it will allow you to build properly the substrings - for now, you are always starting from the beginning of the string, but this is not correct: `t_token` stores the starting position of the token in the string, and `t_ptr` the ending point; `t_start` and `t_end` would replace them happily.

engine/src/mbldialog.cpp
```diff
@@ -292,7 +296,10 @@ bool SplitOptionListsByChunk(MCChunkType p_chunk_type, const_cstring_array_t *p_
         
 		if (p_option_lists->elements[t_element_count] != nil && t_success)
 		{
-            t_success = SplitStringByChunk(p_chunk_type, p_option_lists->elements[t_element_count], r_option_lists[t_element_count]->elements, r_option_lists[t_element_count]->length, p_itemdelimter);
+            MCAutoStringRef t_string;
+            /* UNCHECKED */ MCStringCreateWithCString(p_option_lists->elements[t_element_count], &t_string);
+            // I don't know how to update the 3rd arg of the method below!
+            t_success = SplitStringByChunk(p_chunk_type, *t_string, r_option_lists[t_element_count]->elements, r_option_lists[t_element_count]->length, p_itemdelimter);
```
livecodesebastien: Here, you have to create an array of `MCStringRef`s which will be passed to `SplitStringByChunk()`.
After that you have to allocate an array of `char*` for `r_option_lists[t_element_count]->elements` (the number of strings created by `SplitStringByChunks` is returned through the fourth param). You then have to copy as C-string (and release) all the `MCStringRef`s to this array - and set the field `r_option_lists[t_element_count]->length` appropriately.

livecodesebastien:
This branch actually only modifies dead code...

PR #244, opened 2013-09-16
Refactor syntax unicode extras5

livecodepanos:


engine/src/lnxans.cpp
```diff
@@ -779,25 +779,22 @@ MCPrinterDialogResult MCA_gtk_printer_setup ( PSPrinterSettings &p_settings )
...
 		{
-			char *t_data_out;
+			MCStringRef t_data_out;
 			uint32_t t_data_out_size;
-			t_data_out = nil;
-			t_data_out_size = 0;
+            t_data_out_size = 0;
 
-			MCRemotePrintSetupDialog(t_data_out, t_data_out_size, t_result, (const char *)t_data_in, t_data_in_size);
+			MCRemotePrintSetupDialog(&t_data_out, t_data_out_size, t_result, (MCStringRef)t_data_in, t_data_in_size);
```
livecodesebastien: You can't cast a void\* as MCStringRef - void\* is just plain data, whereas MCStringRef is a C++ object having more than on the data it contains. However, it's possible to create an MCDataRef from a <byte_t*> using MCDataCreateWithBytes and then to pass it to the MCRemote*Dialog function.
About the update itself, I would suggest to update 2 signatures like this instead:
    void MCRemotePrintSetupDialog(MCDataRef p_config_data, MCDataRef &r_reply_data, uint32_t &r_result);
    void MCRemotePageSetupDialog(MCDataRef p_config_data, MCDataRef &r_reply_data, uint32_t &r_result);
There is no need to specify the length of the data when MCDataRefs are used, since it's possible to query it using MCDataGetLength(). And to keep the consistency with the usual parameters order used elsewhere, it's better to put the return parameters at the end. Anyway there is no implementation of these functions for now, so there will only be signatures and calls to update.

engine/src/lnxans.cpp
```diff
@@ -945,16 +942,15 @@ MCPrinterDialogResult MCA_gtk_page_setup (PSPrinterSettings &p_settings)
...
-			t_data_out = nil;
 			t_data_out_size = 0;
 
-			MCRemotePageSetupDialog(t_data_out, t_data_out_size, t_result, (const char *)t_data_in, t_data_in_size);
+			MCRemotePageSetupDialog(&t_data_out, t_data_out_size, t_result, (MCStringRef)t_data_in, t_data_in_size);
 
 			if (t_result == PRINTER_DIALOG_RESULT_OKAY)
 			{
-				if (MCLinuxPageSetupDecode(t_data_out, t_data_out_size, t_setup))
+				if (MCLinuxPageSetupDecode(MCStringGetCString(*t_data_out), t_data_out_size, t_setup))
```
livecodesebastien: MCLinuxPageSetupDecode/Encode might be updated as well to use MCDataRef instead of a void\* and a length - same thing, there is no implementation (anymore) for these functions, so there won't be that many changes to cascade.

engine/src/mbldialog.cpp
```diff
@@ -166,20 +168,21 @@ bool StringArrayToString (const_cstring_array_t *p_indexes, char *&r_result, MCC
...
+                strcat (t_result, t_delimiter);
+                strcat (t_result, p_indexes->elements[i]);
             }
         }
+        /* UNCHECKED */ MCStringCreateWithCString(t_result, r_result);
     }
     if (!t_success)
     {
-        free (r_result);
+        MCValueRelease(r_result);
```
livecodesebastien: t_result should be free'd, and not r_result - the MCMemoryAllocate is called on t_result.

However - I'm not quite sure about Mark's policy about the way to build string steps by steps - I think using a MCAutoNativeCharArray might a cleaner way to to do the job (no char\* used then). There is a good example in MCS_loadtextfile in sysspec.cpp, but basically the idea is to instanciate a MCAutoNativeCharArray, allocate the memory using t_array.New and then passing the pointer - using t_array.Chars() - to the strcpy/strcat functions. At the end, a mere call to t_name.CreateString allows you to get the MCStringRef without having to care about releasing the memory since it is an auto pointer.

livecodesebastien:
Only needs refactor-syntax_unicode merged in.

livecodesebastien:
The branch doesn't compile on mobile: `m_name` of `MCNativeControl` has been updated to an `MCStringRef`, but none of the code using `m_name` has been updated.
EDIT: wrong branch, this was meant to be on refactor-syntax_unicode_NativeControl

livecodesebastien:
This branch doesn't compile on Linux.
`MCA_folder()` from `lnxans.cpp` must be updated, after the merging of `refactor-syntax_unicode` I guess: `MCS_resolvepath()`'s signature has changed, as well as `MCRemoteFolderDialog()` - both of them now use `MCStringRef`s.

livecodesebastien:
Still buggy on Linux: the script `answer file "Select" with type "Text|txt|TXT"` generates a windows without title, but works fine if no type is specified - the windows created by `answer file "Select" with "/home"` is as expected. The issue certainly comes from `MCA_file_with_types()`

PR #248, opened 2013-09-17
Refactor syntax unicode message with args

livecodefraser:
Warning:

This branch depends on a number of other branches and should not be merged until after them:

refactor-syntax_unicode_button
refactor-syntax_unicode_button_menu
refactor-syntax_unicode_kupdown
refactor-syntax_unicode_findstack
refactor-syntax_unicode_paragraph
refactor-syntax_unicode_sort


livecodesebastien:
Can you just revert the changes to the submodules (ide, thirdparty and prebuilt)?

engine/src/field.cpp
```diff
@@ -1107,15 +1107,17 @@ Boolean MCField::mup(uint2 which)
 				extend = extendwords = False;
 				// MW-2012-01-25: [[ FieldMetrics ]] Co-ordinates are now card-based.
 				setfocus(mx, my);
-				typetext(MCDataGetOldString(*t_text));
+                MCAutoStringRef t_text_str;
+                /* UNCHECKED */ MCStringCreateWithOldString(MCDataGetOldString(*t_text), &t_text_str);
+				typetext(*t_text_str);
```
livecodesebastien, line 1112: This should change to a proper conversion from `MCDataRef` to `MCStringRef` - the encoding used is however a bit confused to determine, see MCTransfer::Fetch - it seems that the encoding given as parameter isn't the encoding used in the returned data...

livecodesebastien, line 1112: So `MCTransferData::Fetch` (l.1105) is currently doing the wrong thing in two cases: it should obviously perform a conversion from `t_current_encoding` to `p_type`, but for now, the opposite is actually done l.407 and 416. If you could fix that while you're at updating this, that would be great!

livecodesebastien:
The script editor is broken: when creating a new function "on mouseUp" (for instance), the autocompletion is broken, it only generates a "e" instead of "end mouseUp".
The same thing is occuring when hitting TAB, instead of indenting the script, it breaks it - the more you press it, the more it breaks.

livecodesebastien:
The branch doesn't compile on Windows - there is still a call to `message_with_args(MCNameRef, MCString)` in w32icon.cpp

livecodesebastien:
iOS and Android versions don't compile as `message_with_args` calls haven't been updated for the mobile files and `#ifdef` blocks

livecodesebastien:
There is a bug on Mac: the combination `Fn + Delete` doesn't delete forward as expected but inserts a non-printable character instead. Note that it behaves correctly on Windows and Linux (i.e the character after the cursor gets deleted).

livecodefraser:
Fn + Delete bug is present in refactor-syntax_unicode; I will take a look and fix it in a different branch.

PR #263, opened 2013-09-23
Refactor syntax unicode execcontexts

livecodefraser:


livecodesebastien:
The branch doesn't compile on mobile (in exec-nativecontrol.cpp): `exec.h` must be included, and `MCEPptr` must update to `MCECptr`.

livecodesebastien:
`exec.h` is missing in `mbliphonevideo.mm`, not `exec-nativecontrol.cpp`

livecodesebastien:
Can you merge in refactor-syntax_unicode?

engine/src/deploy_sign.cpp
```diff
@@ -893,8 +894,8 @@ static bool MCDeploySignWindowsAddTimeStamp(const MCDeploySignParameters& p_para
 		t_success = i2d(i2d_SpcTimeStampRequest, t_request, t_request_data, t_request_length);
 
 	// Convert the request to base64
-	extern MCExecPoint *MCEPptr;
-	MCExecPoint ep(*MCEPptr);
+	extern MCExecContext *MCECptr;
+	MCExecPoint ep(MCECptr->GetEP());
```
livecodesebastien: Here you can update `MCU_base64encode()` and then use it further down to process the conversion instead of using the EP as a storage - behaviour which will be rendered impossible when the EP is completely removed from the ExecContext class

engine/src/dispatch.cpp
```diff
@@ -1559,7 +1559,7 @@ MCObject *MCDispatch::getobjname(Chunk_term type, MCStringRef s)
 		{
 			MCresult->clear(False);
 			MCExecPoint ep(MCdefaultstackptr, NULL, NULL);
-			MCExecPoint *epptr = MCEPptr == NULL ? &ep : MCEPptr;
+			MCExecPoint *epptr = MCECptr == NULL ? &ep : &MCECptr->GetEP();
```
livecodesebastien, line 1553: You can now use `MCU_geturl()` with an ExecContext and an `MCStringRef`, so you don't need the EP here, just check the return value.

engine/src/exec-debugging.cpp
```diff
@@ -60,24 +60,21 @@
...
 void MCDebuggingExecDebugDo(MCExecContext& ctxt, MCStringRef p_script, uinteger_t p_line, uinteger_t p_pos)
 {
-	MCExecPoint& ep = ctxt . GetEP();
-
 	Boolean added = False;
 	if (MCnexecutioncontexts < MAX_CONTEXTS)
 	{
-		ep.setline(p_line);
-		MCexecutioncontexts[MCnexecutioncontexts++] = &ep;
+		ctxt.GetEP().setline(p_line);
```
livecodesebastien, line 66: You can add a wrapper to `setline()` in `MCExecContext` since the line is stored for debugging purpose and is part of the context

livecodesebastien:

PR #268, opened 2013-09-24
Refactor syntax unicode get ep2

livecodepanos:
Removed GetEP() from methods 
`MCEngineSplitScriptIntoMessageAndParameters
MCFilesExecReadComplete
MCFilesExecWriteToFileOrDriver
MCFilesExecWriteToProcess
MCStoreGetPurchaseList`


engine/src/exec-engine.cpp
```diff
@@ -1155,8 +1155,11 @@ static void MCEngineSplitScriptIntoMessageAndParameters(MCExecContext& ctxt, MCS
 
 			// MW-2011-08-11: [[ Bug 9668 ]] Make sure we copy 'pdata' if we use it, since
 			//   mptr (into which it points) only lasts as long as this method call.
-			if (ctxt . GetEP() . gethandler() -> eval(ctxt . GetEP()) == ES_NORMAL)
-				newparam->set_argument(ctxt . GetEP());
+			if (ctxt . GetHandler() -> eval(ctxt . GetEP()) == ES_NORMAL)
```
livecodesebastien, line 1158: The Message Box bug comes from this: if `eval()` succeeds, then `newparam` must be assigned the value from `ep`, not `pdata`.

engine/src/exec-engine.cpp
```diff
@@ -1149,14 +1149,20 @@ static void MCEngineSplitScriptIntoMessageAndParameters(MCExecContext& ctxt, MCS
...
+			{  
+				MCAutoStringRef t_value;
+				ctxt.CopyAsStringRef(&t_value);
+				newparam->setvalueref_argument(*t_value);
+			}
```
livecodesebastien, line 1163: Same problem here, the eval method is for now used to evaluate the content of the EP - revert, this will be updated later.

livecodesebastien:
As for the other branches, unmerge it from ctxt_functions

PR #270, opened 2013-09-24
Refactor syntax unicode get ep3

livecodepanos:
Updates on 
`MCInterfaceExecCreateStack
 MCInterfaceExecCreateCard
MCInterfaceExecCreateControl
MCInterfaceExecClone`


engine/src/exec-interface.cpp
```diff
@@ -2896,13 +2896,13 @@ void MCInterfaceExecCreateStack(MCExecContext& ctxt, MCObject *p_object, MCStrin
 	{
 		MCAutoStringRef t_name;
 		p_object->names(P_NAME, &t_name);
-		ctxt . GetEP() . setvalueref(*t_name);
-		if (MCdefaultstackptr->setprop(0, P_MAIN_STACK, ctxt . GetEP(), False) != ES_NORMAL)
-		{
+		ctxt . SetValueRef(*t_name);
```
livecodesebastien: This is not needed, since now `setstringprop()` no doesn't use the ExecContext as a value bearer - and `SetValueRef()` is meant to disappear when the GetEP() is completely removed from the project

engine/src/exec-interface.cpp
```diff
@@ -2911,8 +2911,8 @@ void MCInterfaceExecCreateStack(MCExecContext& ctxt, MCObject *p_object, MCStrin
 
 	if (p_new_name != nil)
 	{
-		ctxt . GetEP() . setvalueref(p_new_name);
-		t_object->setprop(0, P_NAME, ctxt . GetEP(), False);
+		ctxt . SetValueRef(p_new_name);
```
livecodesebastien: Same here

engine/src/exec-interface.cpp
```diff
@@ -2943,8 +2943,8 @@ void MCInterfaceExecCreateCard(MCExecContext& ctxt, MCStringRef p_new_name, bool
...
```
livecodesebastien: Same here

engine/src/exec-interface.cpp
```diff
@@ -3011,8 +3011,8 @@ void MCInterfaceExecCreateControl(MCExecContext& ctxt, MCStringRef p_new_name, i
...
```
livecodesebastien: Same here - and for all the following...

engine/src/exec-interface.cpp
```diff
@@ -2916,13 +2916,12 @@ void MCInterfaceExecCreateStack(MCExecContext& ctxt, MCObject *p_object, MCStrin
 	{
 		MCAutoStringRef t_name;
 		p_object->names(P_NAME, &t_name);
-		ctxt . GetEP() . setvalueref(*t_name);
-		if (MCdefaultstackptr->setprop(0, P_MAIN_STACK, ctxt . GetEP(), False) != ES_NORMAL)
-		{
+		MCdefaultstackptr->setstringprop(ctxt, 0, P_MAIN_STACK, False, *t_name);
+		/*{
```
livecodesebastien: You should keep the execution as it was done in case `setprop()` fails

PR #273, opened 2013-09-25
Refactor syntax unicode get ep5

livecodepanos:


livecodesebastien:
Doesn't compile on mobile, `MCS_read_socket()` should be updated in `mblnetwork.cpp`

engine/src/opensslsocket.cpp
```diff
@@ -614,18 +616,19 @@ void MCS_close_socket(MCSocket *s)
 	s->closing = True;
 }
 
-void MCS_read_socket(MCSocket *s, MCExecPoint &ep, uint4 length, const char *until, MCNameRef mptr)
+void MCS_read_socket(MCSocket *s, MCExecContext &ctxt, uint4 length, const char *until, MCNameRef mptr)
```
livecodesebastien: This function shouldn't pass anymore the return value via `ctxt`, you need to add an `MCDataRef` return parameter.
Also, use `MCExecContext::GetObject()` instead of `GetEP().getobject()`

PR #292, opened 2013-10-01
Refactor syntax unicode mc session

livecodepanos:
I created this new pull request for this branch, because the previous one was closed, so I guess this was the reason it could not update itself with the most recent changes.


engine/src/srvoutput.cpp
```diff
@@ -486,43 +487,34 @@ bool MCSafeCStringEqual(const char *a, const char *b)
...
+		t_success = true;
+	
+	MCAutoStringRef t_encoded;
+	if (t_success && !MCStringIsEmpty(p_value))
 	{
-		// urlencode cookie value
 		MCExecPoint ep;
-		ep.setsvalue(p_value);
+		ep.setsvalue(MCStringGetCString(p_value));
 		MCU_urlencode(ep);
```
livecodesebastien: The EP here should obviously disappear, could you update `MCU_urlencode()` to MCU_urlencode(MCStringRef p_url, MCStringRef &r_encoded)`? There are only few places where it's used - in some of them, a temporary EP was used to comply with the signature...

engine/src/srvcgi.cpp
```diff
@@ -1733,14 +1733,14 @@ bool MCS_get_session_id(MCStringRef& r_id)
 	return MCStringCreateWithCString(MCsessionid, r_id);
 }
 
-bool MCServerGetSessionIdFromCookie(char *&r_id)
+bool MCServerGetSessionIdFromCookie(MCStringRef &r_id)
```
livecodesebastien, line 1737: I've pushed Fraser's new additional functions to `MCExecContext` allowing us to fetch properly values from an array, depending on the context, on refactor-syntax_unicode, which you can merge it in.
At the end of this function, there is a call to `fetcharrayelement_oldstring()`. You should get rid of it and use `MCExecContext::CopyElementAsString()` instead

livecodesebastien:
This just needs to add a `#include "exec.h"` in `srvcgi.cpp`

PR #305, opened 2013-10-07
Refactor syntax unicode get cor old string1

livecodepanos:
Updates on:

```
measure_stringref()
MCInterfaceMakeVisualEffectArgument()
MCInterfaceExecVisualEffect()
convert_options_array()
MCObject::SetProperties()
```


engine/src/lnxpsprinter.cpp
```diff
@@ -1402,12 +1402,13 @@ void MCPSMetaContext::setfont(MCFontStruct *font)
...
-	const char *fontname;
-	fontname = MCStringGetCString(MCNameGetString(fontname_n));
+	MCStringRef fontname;
+	fontname = nil;
+	fontname = MCValueRetain(MCNameGetString(fontname_n));
 
 	while (i < fontcount)
 	{
-		if (strcasecmp(fontname, defaultfonts[i].fontname) == 0 )
+		if (MCStringIsEqualToCString(fontname, defaultfonts[i].fontname, kMCCompareExact))
```
livecodesebastien: You can directly use `MCNameIsEqualToCString` here

engine/src/mbliphonedc.mm
```diff
@@ -362,7 +360,7 @@ static void MCScreenDCDoSetBeepSound(void *p_env)
 bool MCScreenDC::setbeepsound(MCStringRef p_beep_sound)
 {
 	MCScreenDCDoSetBeepSoundEnv t_env;
-	t_env . sound = MCStringGetCString(p_beep_sound);
+	t_env . sound = MCValueRetain(p_beep_sound);
```
livecodesebastien: No retain here, because `t_env.sound` is just read in the callback, and would never be released otherwise.

livecodesebastien:

PR #312, opened 2013-10-09
Refactor syntax unicode get cor old string2

livecodepanos:


engine/src/exec-printing.cpp
```diff
@@ -240,35 +240,103 @@ static void MCPrintingPrinterPageRangeParse(MCExecContext& ctxt, MCStringRef p_i
...
+			{
+				MCAutoStringRef t_substring;
+				/* UNCHECKED */ MCStringCopySubstring(p_input, MCRangeMake(t_pos, MCStringGetLength(p_input) - t_pos), &t_substring);
+				/* UNCHECKED */ MCStringToInteger(*t_substring, t_from);
+				t_to = t_from;
+				t_pos = MCStringGetLength(p_input);
+			}
+		}
+
 		if (!t_error)
```
livecodesebastien, line 307: I'm not really sure of what you are doing here... The idea here is to extract the page ranges contained in `p_input` (like "2,3-5,8-12")
The algorithm to be applied is:

loop while `t_pos` within string boundaries and `t_error == false`
   convert the string to an integer from the current position - this updates `t_pos` to the character after the number converted
   if the conversion failed:
      `t_error = true`
   else if the character after the number is a comma or the string is over // this range is just one page:
      `t_to` is the same as `t_from`
   else if the character after the number is a dash // it's a range (like "15-20"):
      try to convert the next characters to an unsigned integer - this updates `t_pos` to the character after the number converted

   if `t_error == false`:
     call `MCU_disjointrangeinclude` with `t_from` and `t_end`
     if there is a comma at `t_pos`: 
        offset `t_pos` by 1 to discard the comma
     else if the string is over:
        quit the loop // by setting `t_error` to `true`
end loop

PR #315, opened 2013-10-14
Refactor syntax unicode get cor old string3

livecodepanos:


engine/src/mbliphoneextra.mm
```diff
@@ -259,7 +259,10 @@ static void export_image(void *p_context)
 	ctxt = (export_image_t *)p_context;
 	
 	NSData *t_data;
-	t_data = [[NSData alloc] initWithBytes: (void *)MCStringGetCString(ctxt -> raw_data) length: MCStringGetLength(ctxt -> raw_data)];
+	char *temp;
+	/* UNCHECKED */ MCStringConvertToCString(ctxt -> raw_data, temp);
+	t_data = [[NSData alloc] initWithBytes: (void *)temp length: strlen(temp)];
+	delete temp;
```
livecodesebastien: This should be updated to use an `MCDataRef` instead of an `MCStringRef` as it is raw data (as should be all the "raw data" parameters). You need then to update further the structure `export_image_t` and the functions referencing it.

livecodesebastien:
There are few issues in this branch: mainly, when functions takes `const char*` parameters which are only compared to `const char*` stored in tables, there is no actual need to update them since they are used as they are

livecodesebastien:
The branch doesn't compile on Mac, you forgot to update `MCMacOSXOldJavaScriptEnvironment`'s `Run()` definition in osxscriptold.cpp

PR #317, opened 2013-10-18
Refactor syntax unicode remaining strings1

livecodepanos:


livecodesebastien:
Hey Panos, there is actually no need to update a function's signature taking a `const char*` when it merely compares it with a static C-string in its body (like it is done with `p_version` in numerous places). Could you revert those changes then?

livecodesebastien:
This branch mainly updates the `version` parameter of functions to a StringRef, but the version number has been updated to an integer with the file format. The branch can be dropped.

PR #322, opened 2013-10-22
Refactor syntax unicode fieldtext

livecodefraser:


livecodesebastien:
This branch breaks "File > Open stack" on Linux: the stack is never opened - no crash occurs though

PR #327, opened 2013-10-28
Refactor syntax unicode linux

livecodefraser:


livecodesebastien:
Can you revert the changes in the submodules?

PR #328, opened 2013-10-28
refactor-syntax_unicode_CStringMac_clean

livecodepanos:


engine/src/dskmac.cpp
```diff
@@ -928,21 +944,22 @@ static void configureSerialPort(int sRefNum)
...
-	//configure the serial output device
-	parseSerialControlStr(str,&theTermios);
+    for (int i = 0 ; i < nsettings ; i++)
+    {
+        // Note: 't_settings' is an array of strings
+        MCValueRef t_settingval = nil;
+        /* UNCHECKED */ MCArrayFetchValueAtIndex(*t_settings, i, t_settingval);
+        MCStringRef t_setting = (MCStringRef)t_settingval;
+        parseSerialControlStr(t_setting, &theTermios);
+        MCValueRelease(t_setting);
```
livecodesebastien: Don't release `t_settings` as it is the pointer to the value contained by the array - this is _not_ a copy

engine/src/dskmac.cpp
```diff
@@ -1520,7 +1541,11 @@ bool MCS_mac_FSSpec2path(FSSpec *fSpec, MCStringRef& r_path)
     
 	/* UNCHECKED */ MCS_pathfromnative(*t_filename_std, &t_filename);
     
-    t_char_ptr = strclone(MCStringGetCString(*t_filename));
+    char *temp;
+    /* UNCHECKED */ MCStringConvertToCString(*t_filename, temp);
+    
+    t_char_ptr = strclone(temp);
+    delete temp;
```
livecodesebastien: You can directly use `MCStringConvertToCString(*t_filename, t_char_ptr)`

engine/src/dskmac.cpp
```diff
@@ -1788,11 +1807,14 @@ OSErr MCS_path2FSSpec(MCStringRef p_filename, FSSpec *fspec)
...
+    char *temp;
+    /* UNCHECKED */ MCStringConvertToCString(*t_resolved_path, temp);
     
-	char *f2 = strrchr(MCStringGetCString(*t_resolved_path), '/');
-	if (f2 != NULL && f2 != (const char*)MCStringGetCString(*t_resolved_path))
+	char *f2 = strrchr(temp, '/');
+	if (f2 != NULL && f2 != (const char*)temp)
 		*f2++ = '\0';
 	char *fspecname = strclone(f2);
+    delete temp;
```
livecodesebastien: No quite right here (I was wrong when I refactored the code). The logic should be:
if the furthest right slash of `t_resolved_path` is not the first character then
    `t_resolved_path` is now the characters up to that slash (excluded)
    `fspecname` contains the characters after that slash.
else
    `t_resolved_path` remains unchanged
    `fspecname` is nil
end if

The function is only called by `MCPlayer::recordsound` - the other callers are never referenced - I guess that's why this mistake didn't cause any issue

engine/src/dskmac.cpp
```diff
@@ -3790,12 +3822,12 @@ struct MCMacSystemService: public MCMacSystemServiceInterface//, public MCMacDes
         delete[] replymessage;
         replylength = MCStringGetLength(p_message);
         replymessage = new char[replylength];
```
livecodesebastien: Remove the allocation - it's already done by `MCStringConvertToCString`

engine/src/dskmac.cpp
```diff
@@ -5700,23 +5734,28 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
...
         
         t_temp_file_chars = nil;        
-        t_success = MCStringCreateMutable(0, &t_temp_file_auto);
+        t_success = MCStringCreateMutable(0, t_temp_file_auto);
         
         if (t_success && FSFindFolder(kOnSystemDisk, kTemporaryFolderType, TRUE, &t_folder_ref) == noErr)
         {
             int t_fd;
-            t_success = MCS_mac_fsref_to_path(t_folder_ref, &t_temp_file_auto);
+            t_success = MCS_mac_fsref_to_path(t_folder_ref, t_temp_file_auto);
```
livecodesebastien, line 5758: What you should do here is rather creating a mutable copy of `t_temp_file_auto` after the call to `MCS_mac_fsref_to_path`.

engine/src/dskmac.cpp
```diff
@@ -7264,25 +7309,23 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
                     if (MCprocesses[index].pid > 0)
                         Kill(MCprocesses[index].pid, SIGKILL);
                     MCprocesses[index].pid = 0;
-                    MCeerror->add
-                    (EE_SHELL_BADCOMMAND, 0, 0, MCStringGetCString(p_command));
-                    return false;
+                    MCeerror->add(EE_SHELL_BADCOMMAND, 0, 0, p_command);
+                    return IO_ERROR;
```
livecodesebastien: `Shell` returns a boolean!

engine/src/dskmac.cpp
```diff
@@ -7264,25 +7309,23 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
...
+                return IO_ERROR;
             }
         }
         else
         {
-            MCeerror->add
-            (EE_SHELL_BADCOMMAND, 0, 0, MCStringGetCString(p_command));
-            return false;
+            MCeerror->add(EE_SHELL_BADCOMMAND, 0, 0, p_command);
+            return IO_ERROR;
```
livecodesebastien: Same here

engine/src/dskmac.cpp
```diff
@@ -7264,25 +7309,23 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
...
             }
             else
             {
                 close(tochild[0]);
                 close(tochild[1]);
-                MCeerror->add
-                (EE_SHELL_BADCOMMAND, 0, 0, MCStringGetCString(p_command));
-                return false;
+                MCeerror->add(EE_SHELL_BADCOMMAND, 0, 0, p_command);
+                return IO_ERROR;
```
livecodesebastien: Same here

livecodesebastien:

PR #330, opened 2013-10-29
Refactor syntax unicode mbl common

livecodepanos:


engine/src/mblad.cpp
```diff
@@ -88,7 +88,9 @@ MCAdType MCAdTypeFromCString(const char *p_string)
 
 const char *MCAdGetInneractiveKey(void)
 {
-    return MCStringGetCString(s_inneractive_ad_key);
+    MCAutoPointer<char> t_key;
+    /* UNCHECKED */ MCStringConvertToCString(s_inneractive_ad_key, &t_key);
+    return *t_key;
```
livecodesebastien: This doesn't work as you are returning a pointer to a `char*` which is destroyed at the end of the function - that the purpose of `MCAutoPointer`: automatically releasing the memory allocated when the function is exited.
Update the signature to `MCStringRef MCAdGetInneractiveKey(void)` and update the two calls in `mblandroidad.cpp` and `mbliphonead.mm` instead

engine/src/mblnetwork.cpp
```diff
@@ -51,7 +51,9 @@
 static bool MCS_mbl_hostNameToAddress(MCStringRef p_hostname, MCSystemHostResolveCallback p_callback, void *p_context)
 {
 	struct hostent *he;
-	he = gethostbyname(MCStringGetCString(p_hostname));
+    MCAutoPointer<char> t_hostname;
+    /* UNCHECKED */ MCStringConvertToCString(p_hostname, &t_hostname);
+	he = gethostbyname(*t_hostname);
```
livecodesebastien, line 56: `gethostbyname()` is a system call, it should use UTF-8

engine/src/mblspec.cpp
```diff
@@ -278,7 +278,9 @@ static bool MCS_geturl_callback(void *p_context, MCSystemUrlStatus p_status, con
 void MCS_geturl(MCObject *p_target, MCStringRef p_url)
 {
 	MCSGetUrlState t_state;
-	t_state . url = strclone(MCStringGetCString(p_url));
+    MCAutoPointer<char> t_url;
+    /* UNCHECKED */ MCStringConvertToCString(p_url, &t_url);
+	t_state . url = strclone(*t_url);
```
livecodesebastien: You can use `MCStringConvertToCString(p_url, t_state . url)` which will avoid you the hassle of creating a temporary value for nothing (on the top of doubling the allocation necessary).

engine/src/mblspec.cpp
```diff
@@ -555,7 +559,9 @@ void MCS_posttourl(MCObject *p_target, MCDataRef p_data, MCStringRef p_url)
 	
 	if (t_success)
 	{
-		t_state . url = strclone(MCStringGetCString(*t_processed));
+        MCAutoPointer<char> t_url;
+        /* UNCHECKED */ MCStringConvertToCString(*t_processed, &t_url);
+		t_state . url = strclone(*t_url);
```
livecodesebastien: Same as above

engine/src/mblspec.cpp
```diff
@@ -643,7 +649,9 @@ void MCS_putintourl(MCObject *p_target, MCDataRef p_data, MCStringRef p_url)
 	
 	if (t_success)
 	{
-		t_state.url = strclone(MCStringGetCString(*t_processed));
+        MCAutoPointer<char> t_url;
+        /* UNCHECKED */ MCStringConvertToCString(*t_processed, &t_url);
+		t_state.url = strclone(*t_url);
```
livecodesebastien: Same as above

engine/src/mblspec.cpp
```diff
@@ -725,7 +733,9 @@ void MCS_downloadurl(MCObject *p_target, MCStringRef p_url, MCStringRef p_file)
 	
 	if (t_success)
 	{
-		t_state . url = strclone(MCStringGetCString(*t_processed));
+        MCAutoPointer<char> t_url;
+        /* UNCHECKED */ MCStringConvertToCString(*t_processed, &t_url);
+		t_state . url = strclone(*t_url);
```
livecodesebastien: Same as above

livecodesebastien:
Also, your updates to `message_with_valueref_args` are overlapping Fraser's changes on the branch `message_with_args`, which I merged in unicode this afternoon. Could you merge refactor-syntax_unicode and fix the conflicts?

engine/src/mblhandlers.cpp
```diff
@@ -772,37 +772,37 @@ Exec_stat MCHandleSetAllowedOrientations(void *context, MCParameter *p_parameter
...
+            else if (MCStringIsEqualToCString(t_orientation, "landscape left", kMCCompareCaseless))
                 t_orientations_set |= ORIENTATION_LANDSCAPE_LEFT_BIT;
-            else if (MCCStringEqualCaseless(t_orientations_array[i], "face up"))
+            else if (MCStringIsEqualToCString(t_orientation, "face up", kMCCompareCaseless))
                 t_orientations_set |= ORIENTATION_FACE_UP_BIT;
-            else if (MCCStringEqualCaseless(t_orientations_array[i], "face down"))
+            else if (MCStringIsEqualToCString(t_orientation, "face down", kMCCompareCaseless))
                 t_orientations_set |= ORIENTATION_FACE_DOWN_BIT;
+            
+            MCValueRelease(t_orientation);
```
livecodesebastien: `t_orientation` should not be released as it is not a copied value, but a fetched value - releasing it will actually release what is stored in the array.

engine/src/mblspec.cpp
```diff
@@ -408,33 +393,26 @@ void MCUrlLoadEvent::Dispatch(void)
...
+			case kMCSystemUrlStatusFinished:
             {
-				MCAutoDataRef t_data;
-                MCAutoNumberRef t_num;
-                /* UNCHECKED */ MCDataCreateWithBytes((const byte_t*)m_data.bytes, m_data.size, &t_data);
-                /* UNCHECKED */ MCNumberCreateWithUnsignedInteger(m_data.size, &t_num);
-                t_object -> message_with_valueref_args(m_message, *t_url, MCSTR("downloaded"), *t_data, *t_num);
+                MCAutoStringRef t_data;
+                /* UNCHECKED */ MCStringCreateWithCString(static_cast<char*>(m_data.bytes), &t_data);
+				t_object -> message_with_valueref_args(m_message, m_url, MCSTR("downloaded"), *t_data);
```
livecodesebastien: `t_num` should be sent as it was before

engine/src/mblspec.cpp
```diff
@@ -452,27 +430,28 @@ static bool MCS_loadurl_callback(void *p_context, MCSystemUrlStatus p_status, co
...
 	if (p_status == kMCSystemUrlStatusError)
 		MCMemoryDelete(context->data.bytes);
 	else if (p_status == kMCSystemUrlStatusLoading)
 	{
-		const MCString *t_data = static_cast<const MCString*>(p_data);
-		MCMemoryReallocate(context->data.bytes, context->data.size + t_data->getlength(), context->data.bytes);
-		MCMemoryCopy(static_cast<uint8_t*>(context->data.bytes) + context->data.size, t_data->getstring(), t_data->getlength());
-		context->data.size += t_data->getlength();
+		MCMemoryReallocate(context->data.bytes, context->data.size + MCStringGetLength(*t_data), context->data.bytes);
+		MCMemoryCopy(static_cast<uint8_t*>(context->data.bytes) + context->data.size, *t_data, MCStringGetLength(*t_data));
```
livecodesebastien: You need to pass a pointer to an array of bytes to `MCMemoryCopy`, not a StringRef (you actually want to copy `p_data` here).

engine/src/mblspec.cpp
```diff
@@ -452,27 +430,28 @@ static bool MCS_loadurl_callback(void *p_context, MCSystemUrlStatus p_status, co
...
+		context->data.size += MCStringGetLength(*t_data);
 	}
-	
-	send_url_progress(context -> object, p_status, context -> url, context -> data . size, context -> total, (const char *)p_data);
+
+    send_url_progress(context -> object, p_status, context -> url, context -> data . size, context -> total, *t_data);
```
livecodesebastien, line 454: Careful here, `send_url_progress` takes a `const char*`, not a StringRef.

livecodesebastien:
There is an issue with `send_url_progress`: for now the function's parameter `p_data` may be either an `MCStringRef` or a `integer_t*`. You should ensure that all the calls to `send_url_progress` have been updated to pass an `MCStringRef` instead of a `char*`, and update function's code - it still considers that `p_data` is a `char *` for now. Also, `send_url_progress` should ensure that the last argument passed to `MCUrlProgressEvent::CreateUrlProgressEvent` is not `nil`.
Could you also merge refactor-syntax_unicode in?

livecodesebastien:
Can you revert the changes in engine/engine.xcodeproj/project.pbxproj? The merging broke it and the app no longer starts

PR #346, opened 2013-11-01
refactor-syntax_unicode_shims

livecodepanos:


engine/src/cmds.cpp
```diff
@@ -119,22 +119,11 @@ Exec_stat MCChoose::exec(MCExecPoint &ep)
...
-	MCAutoStringRef t_string;
-	/* UNCHECKED */ ep.copyasstringref(&t_string);
-	MCInterfaceExecChooseTool(ctxt, *t_string, littool);
-	
-	if (!ctxt . HasError())
-		return ES_NORMAL;
-
-	return ctxt . Catch(line, pos);
+    MCAutoStringRef t_string;
+    if (!ctxt . EvalOptionalExprAsNullableStringRef(etool, EE_CHOOSE_BADEXP, &t_string))
```
livecodesebastien: This is actually a case where the string is never `nil`, as `copyasstringref()` returns an empty string is `ep` is "empty", so no nullable string here

engine/src/cmdsc.cpp
```diff
@@ -2342,36 +2324,16 @@ Exec_stat MCLoad::exec(MCExecPoint &ep)
...
-	/* UNCHECKED */ ep . copyasstringref(&t_url);
-	MCNetworkExecLoadUrl(ctxt, *t_url, *t_message);
-
-	if (!ctxt . HasError())
-		return ES_NORMAL;
-
-	return ctxt . Catch(line, pos);
+    MCNewAutoNameRef t_message;
+    
+    if (!ctxt . EvalOptionalExprAsNullableNameRef(message, EE_LOAD_BADMESSAGEEXP, &t_message))
```
livecodesebastien: Not nullable here since it defaults to `kMCEmptyName` in case `message == nil`

engine/src/cmdsc.cpp
```diff
@@ -2503,30 +2453,16 @@ Exec_stat MCPost::exec(MCExecPoint &ep)
...
-	return ctxt . Catch(line, pos);
+    MCAutoDataRef t_data;
+    if (!ctxt . EvalExprAsDataRef(source, EE_POST_BADSOURCEEXP, &t_data))
+        return;
+    
+    MCAutoStringRef t_url;
+    if (!ctxt . EvalExprAsStringRef(dest, EE_POST_BADDESTEXP, &t_url))
+        return;
+    
+    ctxt . SetIt(it);
```
livecodesebastien, line 2464: This is not needed - the former code was just creating a temporary `MCExecContext`, not setting `It`

engine/src/cmds.cpp
```diff
@@ -995,21 +915,12 @@ Exec_stat MCGet::exec(MCExecPoint &ep)
...
-	MCEngineExecGet(ctxt, *t_value);
-	if (!ctxt . HasError())
-		return ES_NORMAL;
-	
-	return ctxt . Catch(line, pos);
+    MCAutoValueRef t_value;
+    if (!ctxt . EvalExprAsValueRef(value, EE_GET_BADEXP, &t_value))
+        return;
+    
+    ctxt . SetIt(it);
```
livecodesebastien, line 922: Same as below, the `SetIt` function is not needed - if you have a look at `MCEngineExecGet`, you'll see it exactly does this. It was needed before as a temporary `MCExecContext` was being used, and thus needed to have the `it` value set.

engine/src/cmdsc.cpp
livecodesebastien, line 2464: So this is actually needed as an `MCExecContext` might not have `m_it` set in some cases for the moment. Forget that then

engine/src/cmds.cpp
livecodesebastien, line 922: So same as below again, keep `SetIt` and ignore my previous comment

PR #359, opened 2013-11-06
Refactor syntax unicode win32

livecodefraser:
There are a few bits and pieces remaining but the bulk of the work is done so no point holding things up while I do the last fiddly wee bits!


engine/src/dskw32.cpp
```diff
@@ -1843,11 +2017,15 @@ struct MCWindowsDesktop: public MCSystemInterface, public MCWindowsSystemService
...
+		{
+			r_address = MCSTR("unknown");
+			return true;
+		}
         
         char *buffer = new char[MAXHOSTNAMELEN + 1];
         gethostname(buffer, MAXHOSTNAMELEN);
-        return MCStringFormat(r_address, "%s:&d", buffer, MCcmd);
+		buffer[MAXHOSTNAMELEN] = '\0';
+        return MCStringFormat(r_address, "%s:%s", buffer, MCcmd);
```
livecodesebastien: `MCcmd` is a StringRef, the format string should be `"%s:%@"`

engine/src/dskw32.cpp
```diff
@@ -2886,13 +3139,17 @@ struct MCWindowsDesktop: public MCSystemInterface, public MCWindowsSystemService
...
         
         MCAutoStringRef t_tmp_name;
-        return MCStringMutableCopy(*t_long_path, &t_tmp_name) &&
-			MCStringAppendFormat(*t_tmp_name, "/%s", t_ptr + 1) &&
-			MCStringCopy(*t_tmp_name, r_tmp_name);
+        if (!MCStringMutableCopy(*t_long_path, &t_tmp_name) &&
+			MCStringAppendFormat(*t_tmp_name, "/%s", t_ptr + 1))
+			return false;
+	
+		r_tmp_name = MCValueRetain(*t_tmp_name);
```
livecodesebastien: StringRefs return'd should always be immutable, so `MCStringCopy` must be used for that one

engine/src/w32ans.cpp
```diff
@@ -61,18 +63,18 @@ static void getfilter(MCStringRef p_filter, MCStringRef &r_filter)
...
 			MCStringAppend(*t_filterstring, p_filter);
 		}
 
-		/* UNCHECKED */ MCStringAppendNativeChar(*t_filterstring, '\0');
+		/* UNCHECKED */ MCStringAppendChar(*t_filterstring, '\0');
 		/* UNCHECKED */ MCStringFindAndReplaceChar(*t_filterstring, '\n', '\0', kMCStringOptionCompareCaseless);
 		/* UNCHECKED */ MCStringFindAndReplaceChar(*t_filterstring, ',', '\0', kMCStringOptionCompareCaseless);
 
-		/* UNCHECKED */ MCStringCopy(*t_filterstring, r_filter);
+		r_filter = MCValueRetain(*t_filterstring);
```
livecodesebastien: Copy here as well

engine/src/w32ans.cpp
```diff
@@ -167,30 +173,35 @@ static void build_paths(MCStringRef &r_path)
...
+		build_path(*t_std_path, *t_files, *t_path);
+	}
 
-	sg_chosen_files . clear();
-	sg_chosen_folder . clear();
+	s_chosen_files.Delete();
+	s_chosen_folder.Delete();
 
-	/* UNCHECKED */ MCStringCopy(*t_path, r_path);
+	r_path = MCValueRetain(*t_path);
```
livecodesebastien: Copy as well

engine/src/w32ans.cpp
```diff
@@ -698,7 +733,7 @@ static void get_new_filter(MCStringRef *p_types, uint4 p_type_count, MCStringRef
 	else
 	{
 		/* UNCHECKED */ MCStringAppendChar(*t_filters, '\0');
-		/* UNCHECKED */ MCStringCopy(*t_filters, r_filters);
+		r_filters = MCValueRetain(*t_filters);
```
livecodesebastien: Should be a copy

engine/src/dskw32.cpp
```diff
@@ -495,27 +535,38 @@ bool MCS_getlongfilepath(MCStringRef p_short_path, MCStringRef& r_long_path)
...
-	if (t_path_len - 1 != GetCurrentDirectoryA(t_path_len, (LPSTR)t_buffer.Chars()))
+	// NOTE:
+	// Get/SetCurrentDirectory are not supported by Windows in multithreaded environments
+	MCAutoArray<unichar_t> t_buffer;
+
+	// Retrieve the length of the current directory
+	DWORD t_path_len = GetCurrentDirectoryW(0, NULL);
+	/* UNCHECKED */ t_buffer.New(t_path_len);
+
+	DWORD t_result = GetCurrentDirectory(t_path_len, t_buffer.Ptr());
```
livecodesebastien: Even though `GetCurrentDirectory` and `GetCurrentDirectoryW` should be the same (returning TCHAR), it might be worth to explicitly use the W version here as well for consistency with the previous call

PR #387, opened 2013-11-15
Refactor syntax unicode shims eval2

livecodepanos:


engine/src/funcs.cpp
```diff
@@ -3503,18 +3337,17 @@ Exec_stat MCMatch::eval(MCExecPoint &ep)
...
     
     if (!ctxt.HasError())
     {
         for (uindex_t i = 0; i < t_result_count; i++)
         {
-            MCVariable *t_var = t_result_params->evalvar(ep);
+            MCVariable *t_var = t_result_params->evalvar(ctxt);
             if (t_var == nil)
             {
                 MCeerror->add(EE_MATCH_BADDEST, line, pos);
```
livecodesebastien: This should update to MCLegacyThrow(EE_MATCH_BADDEST).
Doesn't compile on Linux as the functions returns `ES_ERROR` despite it should return void.

PR #391, opened 2013-11-15
Refactor syntax unicode shims eval3

livecodepanos:


engine/src/funcsm.cpp
```diff
@@ -1179,147 +761,47 @@ Exec_stat MCLn1::eval(MCExecPoint &ep)
...
-	MCS_seterrno(0);
-	ep.setnvalue(log(ep.getnvalue()) / log(2.0));
-	if (MCS_geterrno() != 0 || MCS_isnan(ep.getnvalue()))
-	{
-		MCS_seterrno(0);
-		MCeerror->add(EE_LOG2_DOMAIN, line, pos);
-		return ES_ERROR;
-	}
-	return ES_NORMAL;
-#endif /* MCLog2 */
```
livecodesebastien, line 1241: This `ifdef` block should be kept

livecodesebastien, line 1241: Hm, it seems that the diff produced made it like you deleted it, but it actually still is just further in the file... Nevermind

PR #394, opened 2013-11-19
Refactor syntax unicode mc string get c string

livecodepanos:


engine/src/font.cpp
```diff
@@ -657,26 +657,29 @@ Exec_stat MCF_parsetextatts(Properties which, MCStringRef data,
...
+                        t_end_pos = MCStringGetLength(data);
+                    else
+                        t_end_pos = t_comma;
+                    
                     MCAutoStringRef tdata;
-					/* UNCHECKED */ MCStringCreateWithNativeChars((const char_t *)startptr, sptr - startptr, &tdata);
-					MCU_skip_char(sptr, l);
-					MCU_skip_spaces(sptr, l);
-					if (MCF_setweightstring(style, *tdata))
+                    /* UNCHECKED */ MCStringCopySubstring(data, MCRangeMake(t_start_pos, t_end_pos - t_start_pos), &tdata);
```
livecodesebastien, line 681: You need to increment `t_end_pos` by one after you have copied the substring (that's what `MCU_skip_char` does).

engine/src/iimport.cpp
```diff
@@ -272,7 +272,12 @@ bool MCImageImport(IO_handle p_stream, IO_handle p_mask_stream, MCPoint &r_hotsp
                 MCAutoStringRef t_name;
 				t_success = MCImageDecodeXWD(p_stream, &t_name, t_bitmap);
 				if (t_success)
-					r_name = strdup(MCStringGetCString(*t_name));
```
livecodesebastien, line 275: `r_name` is almost always discarded by the callers, update the function signature and the callers instead

engine/src/iquantization.cpp
```diff
@@ -361,52 +361,37 @@ bool MCImageParseColourList(MCStringRef p_input, uint32_t &r_ncolours, MCColor *
...
-			char *t_nextstring = strchr(t_currentstring, '\n');
-			if (t_nextstring)
-				t_currentstring = t_nextstring + 1;
-			else
-				t_currentstring += strlen(t_currentstring);
-		}
-
-		t_success = MCMemoryNewArray<MCColor>(t_ncolours, t_colours);
+		t_ncolours = 1 + MCStringCountChar(p_input, MCRangeMake(0, MCStringGetLength(p_input)), '\n', kMCCompareExact);
+        t_success = MCMemoryNewArray<MCColor>(t_ncolours, t_colours);
```
livecodesebastien: You can directly split the string and use ArrayGetCount to allocate the memory of the MCColor array

engine/src/objectprops.cpp
```diff
@@ -2027,7 +2027,9 @@ Exec_stat MCObject::setprops(uint32_t p_parid, MCExecPoint& ep)
 	MCValueRef t_value;
 	while(MCArrayIterate(*t_array, t_iterator, t_key, t_value))
 	{
-		MCScriptPoint sp(MCStringGetCString(MCNameGetString(t_key)));
+        MCAutoPointer<char> t_key_cstring;
+        /* UNCHECKED */ MCStringConvertToCString(MCNameGetString(t_key), &t_key_cstring);
+		MCScriptPoint sp(*t_key_cstring);
```
livecodesebastien: This is dead code

engine/src/opensslsocket.cpp
```diff
@@ -1014,10 +1018,15 @@ bool MCS_hostaddress(MCStringRef &r_host_address)
...
+    /* UNCHECKED */ MCStringConvertToCString(d, temp_d);
 	if (m != NULL)
-		buffer = strdup(MCStringGetCString(d));
+    {
+		buffer = strdup(temp_d);
+        delete temp_d;
+    }
 	else
-		buffer = (char *)MCStringGetCString(d);
+		buffer = temp_d;
```
livecodesebastien: No too sure about what the former code was intending, but it looks rather dangerous to keep a pointer to the internal string of an MCStringRe as its size might change or it can even be deallocated; in any case, since `ConvertToCString` allocates the memory anyway, there is no point in copying a string and deleting the original one right after - the `if` can probably be deleted.

engine/src/player.cpp
```diff
@@ -4482,8 +4494,10 @@ static void exportToSoundFile(const char *sourcefile, const char *destfile)
 	
 	if (t_success)
 	{
-		t_success = path_to_dataref(MCStringGetCString(*t_src_resolved_str), t_src_rec) &&
-			path_to_dataref(MCStringGetCString(*t_dst_resolved_str), t_dst_rec);
+		MCAutoPointer<char> t_src_res, t_dst_res;
+        /* UNCHECKED */ MCStringConvertToCString(*t_src_resolved_str, &t_src_res);
+        /* UNCHECKED */ MCStringConvertToCString(*t_dst_resolved_str, &t_dst_res);
+        t_success = path_to_dataref(*t_src_res, t_src_rec) && path_to_dataref(*t_dst_res, t_dst_rec);
```
livecodesebastien: You can update `path_to_dataref` to take an MCStringRef, and use `MCAutoStringRefAsCFString` to proceed the conversion within path_to_dataref

engine/src/player.cpp
```diff
@@ -4578,12 +4592,20 @@ void MCPlayer::stoprecording()
 		}
 #ifdef _WINDOWS
 		if (MCrecordformat == EX_MOVIE)
-			CopyFileA(MCStringGetCString(recordtempfile), MCStringGetCString(recordexportfile),False);
+        {
+            MCAutoStringRefAsUTF8String t_recordtempfile, t_recordexportfile;
+            /* UNCHECKED */ t_recordtempfile . Lock(recordtempfile);
+            /* UNCHECKED */ t_recordexportfile . Lock(recordexportfile);
+			CopyFileA(*t_recordtempfile, *t_recordexportfile, False);
```
livecodesebastien: Use `MCAutoStringRefAsWString`, and call `CopyFileW` instead

engine/src/socket_resolve.cpp
```diff
@@ -357,18 +357,19 @@ bool hostname_resolve_first_sockaddr_callback(void *p_context, bool p_resolved,
...
+	MCAutoPointer<char> t_name_in;
+    /* UNCHECKED */ MCStringConvertToCString(p_name_in, &t_name_in);
 	if (!MCS_init_sockets())
 		return false;
 
 	char *t_name;
-	t_name = strclone(p_name_in);
+	t_name = strclone(*t_name_in);
 
 	char *idstring = strchr(t_name, '|');
```
livecodesebastien: This can be done with substrings and `MCNumberParse`, which takes in consideration the unicodeness of the string.

engine/src/stack.cpp
```diff
@@ -2960,26 +2960,27 @@ bool MCStack::resolve_relative_path(MCStringRef p_path, MCStringRef& r_resolved)
...
 			else
-				t_leaf = MCStringGetCString(p_path);
+				/* UNCHECKED */ MCStringAppend(*t_new_filename, filename);
             
-			strcpy(t_filename + (t_last_separator - MCStringGetCString(t_stack_filename) + 1), t_leaf);
+            if (MCS_exists(*t_new_filename, True))
+            {
+                r_resolved = MCValueRetain(*t_new_filename);
+                return true;
+            }
```
livecodesebastien, line 2980: `t_new_filename` is a mutable, it must be copied to the return value - use `MCStringCopy` instead

engine/src/stack3.cpp
```diff
@@ -1831,22 +1831,31 @@ void MCStack::breakstring(MCStringRef source, MCStringRef*& dest, uint2 &nstring
...
+        {
+			// MW-2007-07-05: [[ Bug 110 ]] - Break string only breaks at spaces, rather than space charaters 
+			uint4 l;
+            l = 0;
+            uint4 remaining_chars;
+            remaining_chars = MCStringGetLength(source);
+			MCU_skip_spaces(source, l);
+            remaining_chars -= l;
+            uindex_t t_start;
+            t_start = 0;
```
livecodesebastien: `t_start` must start at `l`, otherwise the call to `MCU_skip_spaces` is useless.

engine/src/stack3.cpp
```diff
@@ -1831,22 +1831,31 @@ void MCStack::breakstring(MCStringRef source, MCStringRef*& dest, uint2 &nstring
...
+            
+			while(remaining_chars != 0)
 			{
-				while(l != 0 && !isspace(*sptr))
-					sptr += 1, l -= 1;
-
+				while(!isspace(MCStringGetNativeCharAtIndex(source, t_start)))
```
livecodesebastien: Actually your algorithm is correct - I didn't get that `l` is the start point and `t_start` is the length - but only if you change `t_start` to  `l` in this call

engine/src/util.cpp
```diff
@@ -1198,9 +1204,12 @@ static void msort(MCSortnode *b, uint4 n, MCSortnode *t, Sort_type form, Boolean
...
 		case ST_INTERNATIONAL:
 			{
+                char *t1, *t2;
+                /* UNCHECKED */ MCStringConvertToCString(b1->svalue, t1);
+                /* UNCHECKED */ MCStringConvertToCString(b2->svalue, t2);
 				const char *s1, *s2;
-				s1 = MCStringGetCString(b1->svalue);
-				s2 = MCStringGetCString(b2->svalue);
+				s1 = t1;
+				s2 = t2;
```
livecodesebastien, line 1212: You need to delete `t1` and `t2` in the end

engine/src/util.cpp
```diff
@@ -1221,9 +1230,12 @@ static void msort(MCSortnode *b, uint4 n, MCSortnode *t, Sort_type form, Boolean
...
 		case ST_TEXT:
 			{
+                char *t1, *t2;
+                /* UNCHECKED */ MCStringConvertToCString(b1->svalue, t1);
+                /* UNCHECKED */ MCStringConvertToCString(b2->svalue, t2);
 				const char *s1, *s2;
-				s1 = MCStringGetCString(b1->svalue);
-				s2 = MCStringGetCString(b2->svalue);
+				s1 = t1;
+				s2 = t2;
```
livecodesebastien, line 1239: Same here

engine/src/w32script.cpp
```diff
@@ -773,8 +773,11 @@ char *MCWindowsActiveScriptEnvironment::Call(const char *p_method, const char **
 
 MCScriptEnvironment *MCScreenDC::createscriptenvironment(MCStringRef p_language)
 {
+    MCAutoPointer<char> t_language;
+    /* UNCHECKED */ MCStringConvertToCString(p_language, &t_language);
 	LPOLESTR t_ole_language;
-	t_ole_language = ConvertUTF8ToOLESTR(MCStringGetCString(p_language));
+	t_ole_language = ConvertUTF8ToOLESTR(*t_language
```
livecodesebastien: This should probably use `MCAutoStringRefAsUTF8`

engine/src/stack3.cpp
```diff
@@ -1831,22 +1831,37 @@ void MCStack::breakstring(MCStringRef source, MCStringRef*& dest, uint2 &nstring
...
+            remaining_chars -= l;
+            uindex_t t_word_start;
+            t_word_start = l;
+            
+			while(remaining_chars != 0)
 			{
-				while(l != 0 && !isspace(*sptr))
-					sptr += 1, l -= 1;
-
+				while(!isspace(MCStringGetNativeCharAtIndex(source, l)))
```
livecodesebastien, line 1847: You should ensure that `remaining_chars > 0`

PR #398, opened 2013-11-20
refactor-syntax_unicode_MCNameGetOldString

livecodepanos:


engine/src/srvdbgvar.cpp
```diff
@@ -673,8 +673,10 @@ bool MCServerDebugListVariables(uint32_t p_start, uint32_t p_count, MCServerDebu
 			else
 			{
 				t_value = &t_owner -> variable -> getvalue();
-				t_name_ptr = MCNameGetOldString(t_owner -> variable -> getname()) . getstring();
-				t_name_length = MCNameGetOldString(t_owner -> variable -> getname()) . getlength();
+                MCAutoPointer<char> temp_name;
+                /* UNCHECKED */ MCStringConvertToCString(MCNameGetString(t_owner -> variable -> getname()), &temp_name);
+				t_name_ptr = *temp_name;
```
livecodesebastien, line 678: This doesn't work: `temp_name` is deleted at the end of the block, so that `t_name_ptr` then points to deallocated memory. You need to create `t_name` in the same scope as `t_name_ptr`

engine/src/util.cpp
```diff
@@ -1302,55 +1302,50 @@ void _dbg_MCU_realloc(char **data, uint4 osize, uint4 nsize, uint4 csize, const
...
+    MCStringFormat(&t_pattern, "%@ \"%@\"", MCSTR(nametable[type - CT_STACK]), MCNameGetString(name));
+    
+    if (MCStringContains(MCNameGetString(test), *t_pattern, kMCCompareCaseless))
+    {
+        uindex_t t_quotes;
+        /* UNCHECKED */ MCStringFirstIndexOfChar(MCNameGetString(test), '"', 0, kMCCompareExact, t_quotes);
+        if (MCStringGetLength(MCNameGetString(test)) - t_quotes == MCStringGetLength(MCNameGetString(name)) + 2)
+            match = true;
+        if (!match)
+            MCLog("[[ Bug 11068 ]] match name '%s' to '%@' attempted and failed due to better checking", MCNameGetCString(name), test);
```
livecodesebastien: You can use `%@` directly

PR #442, opened 2014-01-06
[[ Bug 11624 ]] Use stack loc transform method in MCClickCmd::exec

livecodeian:


livecodemichael:
Merged into release-6.5.2.

PR #449, opened 2014-01-14
[[ bugfix-11482 ]] iOS control scroller doesn't scroll on iOS7

livecodesebastien:


livecodemichael:
Merged into release-6.5.2.

PR #452, opened 2014-01-14
[[ Bug 11343 ]] Ensure behavior chain is searched for mouseWithin and mo...

livecodeali:
...useStillDown handlers


livecodemichael:
Merged into release-6.5.2.

PR #453, opened 2014-01-15
[[ Bug 11570 ]] Draw stippled background when referenced image file not ...

livecodeali:
...found


livecodemichael:
Merged into release-6.5.2.

PR #454, opened 2014-01-15
[[ Bug 11419 ]] Allow graphic drawing to begin in non-selectable regions

livecodeali:


livecodemichael:
Merged into release-6.5.2.

PR #456, opened 2014-01-16
[[ Bug 11678 ]] Arrow constant should be equivalent to 1

livecodeali:


livecodemichael:
Merged into release-6.5.2.

PR #457, opened 2014-01-16
[[ Bug 11675 ]] Fix infinite loop when skipping over comment in PBM file

livecodeian:


livecodemichael:
Merged into release-6.5.2.

PR #458, opened 2014-01-16
[[ Bug 11656 ]] Don't draw box around linux slider trough

livecodeali:


livecodemichael:
Merged into release-6.5.2.

PR #459, opened 2014-01-16
[[ Bug 11672 ]] If the locale doesn't use AM/PM, then always use 24-hour...

livecodeali:
... time


livecodemichael:
Merged into release-6.5.2.

PR #471, opened 2014-01-27
Rect points

mwieder:
Allows getting "the effective points" of graphic rectangles, round rectangles, and regular polygons.
This allows for some things that are difficult at present: setting the points of a freeform polygon to one of several regular polygon objects; moving an image around the points of a regular polygon, etc.

### DIFF
--- a/engine/src/graphic.cpp
+++ b/engine/src/graphic.cpp
@@ -411,6 +411,26 @@ void MCGraphic::setrect(const MCRectangle &nrect)
 		else
 			MCU_offset_points(realpoints, nrealpoints, nrect.x - rect.x, nrect.y - rect.y);
 	}
+	else if (getstyleint(flags) == F_G_RECTANGLE || getstyleint(flags) == F_ROUNDRECT)
+	{
+		nrealpoints = 4;
+		realpoints = new MCPoint[nrealpoints];
+		realpoints[0].x = nrect.x;
+		realpoints[0].y = nrect.y;
+		realpoints[1].x = nrect.x + nrect.width;
+		realpoints[1].y = nrect.y;
+		realpoints[2].x = nrect.x + nrect.width;
+		realpoints[2].y = nrect.y + nrect.height;
+		realpoints[3].x = nrect.x;
+		realpoints[3].y = nrect.y + nrect.height;
+		if (oldpoints == NULL)
+		{
+			oldpoints = new MCPoint[nrealpoints];
+			uint2 i = nrealpoints;
+			while (i--)
+				oldpoints[i] = realpoints[i];
+		}
+	}
 	if (m_fill_gradient != NULL)
 		setgradientrect(m_fill_gradient, nrect);
 	if (m_stroke_gradient != NULL)
--- a/engine/src/graphic.cpp
+++ b/engine/src/graphic.cpp
@@ -367,7 +367,28 @@ Boolean MCGraphic::doubleup(uint2 which)
 
 void MCGraphic::setrect(const MCRectangle &nrect)
 {
-	if (realpoints != NULL)
+	// MDW-2014-01-19: [[ feature_rect_points ]] set the points of rectangles and rounded rectangles
+	if (getstyleint(flags) == F_G_RECTANGLE || getstyleint(flags) == F_ROUNDRECT)
+	{
+		nrealpoints = 4;
+		realpoints = new MCPoint[nrealpoints];
+		realpoints[0].x = nrect.x;
+		realpoints[0].y = nrect.y;
+		realpoints[1].x = nrect.x + nrect.width;
+		realpoints[1].y = nrect.y;
+		realpoints[2].x = nrect.x + nrect.width;
+		realpoints[2].y = nrect.y + nrect.height;
+		realpoints[3].x = nrect.x;
+		realpoints[3].y = nrect.y + nrect.height;
+		if (oldpoints == NULL)
+		{
+			oldpoints = new MCPoint[nrealpoints];
+			uint2 i = nrealpoints;
+			while (i--)
+				oldpoints[i] = realpoints[i];
+		}
+	}
+	else if (realpoints != NULL)
 	{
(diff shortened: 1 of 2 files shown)


mwieder:
This pull request has been sitting around for a while. I just updated it to better code... any chance someone could take a look at it?

runrevmark:
Hi Mark - didn't quite get around to looking at it today, but should tomorrow! Thanks for your patience.

mwieder:
Thanks. Patience actually paid off because I found a better way of implementing it in the intervening months. Note: I haven't tackled the points of an oval, and I'm debating whether there's a good use case for that - the only thing I can think of is moving a graphic along the points of an oval.

runrevmark:
Hi Mark,

I've taken an initial look at this and it's looking good. My only suggestion would be to not use realpoints/points fields in the generators for the points for things that don't have points. Instead, just generate into a temporary buffer, unparse, then delete the temporary buffer. (This would side-step sorting out a couple of memory leaks which I think might be lurking - I've marked them in the code).

Indeed, you could have a method for each of the shape-types which don't have realpoints, that generate the points array - this would nicely refactor the code in ::draw which deals with the regular polygons (which could also be done on the fly) and mean that the 'points' field in MCGraphic becomes redundent. (Time taken to generate points for a regular polygon is vastly less than the time taken to render the resulting path to the context - so there's no issue about performance here). For example:

bool MCGraphic::getpointsforrect(MCPoint_& r_points, int& r_point_count) { ... }
bool MCGraphic::getpointsforregular(MCPoint_& r_points, int& r_point_count) { ... }

Thanks again,

Mark.

engine/src/graphic.cpp
```diff
@@ -546,14 +547,119 @@ Exec_stat MCGraphic::getprop(uint4 parid, Properties which, MCExecPoint& ep, Boo
...
+				nrealpoints = 0;
+				MCU_roundrect(realpoints, nrealpoints, geteffectiverect(), roundradius);
+				MCU_unparsepoints(realpoints, nrealpoints, ep);
+				break;
+			}
+			case F_G_RECTANGLE:
+			{
+				MCRectangle nrect = geteffectiverect();
+				nrealpoints = 4;
+				realpoints = new MCPoint[nrealpoints];
```
runrevmark: I think there's a memory leak here as its reallocating realpoints each time the points of a rectangle are fetched.

engine/src/graphic.cpp
```diff
@@ -546,14 +547,119 @@ Exec_stat MCGraphic::getprop(uint4 parid, Properties which, MCExecPoint& ep, Boo
 		ep.setint(nsides);
 		break;
 	case P_POINTS:
-		MCU_unparsepoints(realpoints, nrealpoints, ep);
+		// MDW-2014-06-14: [[ rect_points ]] allow effective points as read-only
+		p_flags = getstyleint(flags);
+		if (p_flags == F_ROUNDRECT || p_flags == F_G_RECTANGLE)
```
runrevmark: This should probably include F_REGULAR too - as you can't set the points of a regular graphic as they are defined by the other properties.

(You can set the points property, if style is regular, but I don't believe it has any effect until you set the style to polygon or curve - i.e. one of the styles that uses the realpoints).

mwieder: Quite right - I had F_REGULAR in there and then with sudden loss of clarity deleted it. I'll put it back forthwith. I don't think setting the points of rectangles, rounded rectangles, or regular polygons is a particularly good idea - there's too much validation involved to ensure that the angles and point placements are correct - that's why I enforce the "effective" modifer.

engine/src/graphic.cpp
mwieder: I think you're probably correct. The whole points/oldpoints/realpoints/markerpoints and delpoints()  thing is quite confusing and undocumented. I found that realpoints did what I wanted and stopped there. I'll see about patching the leak.

mwieder:
OK - refactored, cleaned up the memory leaks, pulled F_REGULAR back into the fold, and created new functions to return the points of rects, roundrects, and regular polygons. The regular polygon function is now integrated into the ::draw method as well. I'm at a loss as to how to determine the points of an oval - seems like squaring the circle, but the draw method seems capable of drawing a circle through the fillarc function.

engine/src/graphic.cpp
```diff
@@ -546,14 +550,97 @@ Exec_stat MCGraphic::getprop(uint4 parid, Properties which, MCExecPoint& ep, Boo
...
+		switch (graphic_type)
+		{
+			case F_ROUNDRECT:
+			{
+				fakepoints = NULL;
+				nfakepoints = 0;
+				get_points_for_roundrect(fakepoints, nfakepoints);
+				MCU_offset_points(fakepoints, nfakepoints, -trect.x, -trect.y);
+				MCU_unparsepoints(fakepoints, nfakepoints, ep);
+				break;
```
runrevmark, line 618: I think you need to delete fakepoints here too as MCU_round_rect I believe allocates a new array if the input array is nil.

runrevmark:
Hi Mark - this is almost there I think... There's a potential memory leak I've marked in the pull request, but apart from that it looks fine.

In terms of the oval / arc etc. then this is a little tricky, yes - I'll have a think about the best way to handle that (for completeness it would be useful to have).

Separating out the point computation methods has made the code a bit neater (as you mentioned above MCGraphic is not the prettiest of coding constructs!).

Thanks,

Mark.

mwieder:
Ah... good point. I guess it makes sense that the point array has to be created _somewhere_. I'll get that patched up tonight. Also - do you have a preference for coding style between CamelCase and snake_case? I notice that most of the functions are just stringwordstogetherstyle which I find hard to read.

runrevmark:
Hi Mark - thanks for getting that sorted quickly. I've merged this into develop via 'feature-effective_graphic_points' in the livecode repo. This should appear in the next dp of 6.7.

Thanks again,

Mark.

PR #486, opened 2014-02-12
[[ Bug 11783 ]] Only use ideal font layout on Windows when fullscreen

livecodeian:


livecodemichael:
Merged into release-6.5.3.

PR #502, opened 2014-02-26
[[ Bug 11753 ]] Avoid calling setLooping on MediaPlayer when not require...

livecodeian:
...d


livecodemichael:
Merged into release-6.5.3.

PR #514, opened 2014-03-06
Ensure showController of players is set before rectangle

montegoulding:
I found this one while working with lcVCS. showController needs precedence


livecodemichael:
Hi Monte. Thanks very much for this. It has been merged into develop and should be present in 6.6-rc-2.

PR #574, opened 2014-03-27
[[ Bug 12042 ]] Object chunk evaluation should include new chunk types

livecodeali:
Note this doesn't quite work at the moment due to the fact that the call to resolvechars expects to receive grapheme indices but chunk marking returns codeunit indices. I suspect that MCExecResolveCharsOfField can be safely made a no-op, but it needs a little testing lest the hangs of yesteryear resurface.


livecodeali:
Ok, MCExecResolveCharsOfField made a no-op. No problems in testing.

PR #582, opened 2014-03-31
Refactor syntax unicode text stream

livecodefraser:


livecodesebastien:
The branch is now also in runrev/refactor-syntax_unicode_text_stream.
I pushed a version with few issues cleaned up, but it still fails to return the correct indices in find, and can't replace in a string, as soon as the needle looked for contains a surrogate pair or a combining char.

livecodesebastien:
Replaced by the pull request from my repo, which includes the fixes for combining chars and surrogate pairs

PR #602, opened 2014-04-07
Bugfix 12078

livecodefraser:


livecodesebastien:
Merged with runrevsebastien/bugfix-12078 - strong isolate is used to find out the direction of each line

PR #620, opened 2014-04-09
[[ NumberExpectation ]] Added new property to ExecContext, in order to s...

livecodesebastien:
...pecify when we want to get a number

This avoids a useless number -> ValueRef -> number conversion


livecodesebastien:
Not useful enough since it's only applying to the literal numbers - conversion from string to double/int should be stored in the string itself, as it is often immutable, and that would speed up the number conversion in any place instead of one (small) part only.

PR #650, opened 2014-04-16
Bugfix 12075

livecodesebastien:
Buttons and tooltip drawing based on the MCUnicodeResolveDirection to split any label into the correct blocks of text, allowing RTL to be drawn as RTL. Different approach from the bugfix-12078


livecodesebastien:
Merged with bug fix 12078

PR #665, opened 2014-04-17
[[ Bug 11964 ]] Windows scaled text is spaced incorrectly.

livecodemichael:


livecodemichael:
Merged into release-6.6.2.

PR #669, opened 2014-04-22
[[ Bug 11904 ]] Italic text with underhang is clipped on Windows.

livecodemichael:


livecodemichael:
Merged into release-6.6.2.

PR #670, opened 2014-04-22
[[ Bug 12236 ]] Use logical -> screen conversion functions that return c...

livecodeian:
...onsistent width & height values


livecodemichael:
Merged into release-6.6.2.

PR #709, opened 2014-05-02
[[ Bug 12173 ]] Styling does not work for certain iOS fonts.

livecodemichael:
[[ Bug 12350 ]] The fontStyles is incorrect on iOS.


livecodemichael:
Merged into release-6.6.2.

PR #724, opened 2014-05-08
Bugfix 12409

livecodesebastien:


livecodesebastien:
Wrong bug number… Deleting the wrong branch to avoid confusion

PR #744, opened 2014-05-21
[[ Bug 12458 ]] Fix crash when reading past the end of a data stream whe...

livecodeian:
...n size larger than 1 byte


livecodemichael:
Merged into release-6.6.3.

PR #773, opened 2014-05-27
[[ Bug 12321 ]] Fix stack rect not being restored after unsetting fullsc...

livecodeian:
...reen


livecodemichael:
Merged into release-6.6.3.

PR #801, opened 2014-05-29
[[ Bug 12408 ]] Encryption commands don't work for iOS device builds.

livecodemichael:


livecodemichael:
Merged into release-6.6.2.

PR #859, opened 2014-07-02
[[ Bugfix ]] Fix crash in MCObject::sendmessage()

livecodestephen:
If you had a behaviour script containing a `before` message handler and opened the message watcher you would get a crash in `MCObject::sendmessage()`.  The problem was that it contained an array of strings used to convert a `Handler_type` enum value to a string, but this array didn't cover all possible values of the enum, so you'd end up dereferencing an invalid pointer.

I've added the missing strings to the array and added assertions that the enum is within the array.  I've also added a static assertion macro and use it to check the size of the array.


livecodefraser:
Closing as redundant (two identical pull requests).

PR #913, opened 2014-07-21
[[ Bug 12867 ]] Gradient colours display incorrectly on android

livecodeali:


livecodeian:
Hi Ali, it should be possible to replace all the #ifdefs with a single line:

t_ramp[i] . hw_color = MCGPixelToNative(kMCGPixelFormatBGRA, t_ramp[i] . color);

That should work for all cases.

livecodeali:
Ah yes, I thought that might be the case. Thanks!

PR #1068, opened 2014-08-20
[[ Bug ]] Fix potential crash in MCU_matchname caused by uninitialised globals

livecodeian:


livecodeian:
No bugzilla report for this one as it was encountered while building the engine with Visual Studio 2013, so may not apply to current builds.

PR #1156, opened 2014-09-11
[[ Bug 13404 ]] controlAtScreenLoc needs to use stack-relative loc rathe...

livecodeali:
...r than global


livecodefraser:
Duplicated bug report - was fixed by the bugfix for 13288.

PR #1213, opened 2014-09-25
image metadata property

montegoulding:


engine/src/image.cpp
```diff
@@ -1115,6 +1127,8 @@ Exec_stat MCImage::setprop(uint4 parid, Properties p, MCExecPoint &ep, Boolean e
 			MCPoint t_hotspot;
 			char *t_name = nil;
 			IO_handle t_stream = nil;
+            MCImageMetadata t_metadata;
+            MCMemoryClear(&t_metadata, sizeof(t_metadata));
```
livecodeian: Hi Monte, this change seems to be unnecessary as t_metadata isn't used here.

engine/src/imageloader.cpp
```diff
@@ -96,6 +96,19 @@ bool MCImageLoader::GetFrameCount(uint32_t &r_frame_count)
...
+{
+	if (!EnsureHeader())
+		return false;
+	
+	r_metadata = m_metadata;
+	
+	return true;
+    
+}
+
```
livecodeian, line 107: Note that this will replace the contents of r_metadata with the contents of m_metadata, so you should ensure that m_metadata is cleared in the MCImageLoader::MCImageLoader() constructor, otherwise it may contain uninitialised data.

engine/src/image.cpp
montegoulding: Sorry about that. Was an early change I forgot to cleanup after working out how things worked...

engine/src/imageloader.cpp
montegoulding, line 107: Ah, OK, thanks

trevordevore:
@runrevian What is the status of this pull request? Any chance of getting it incorporated into the engine? Currently on OS X we can't determine the proper dimensions to display retina images at when importing 3rd party image files into LiveCode.

engine/src/image.cpp
```diff
@@ -916,6 +917,17 @@ Exec_stat MCImage::getprop(uint4 parid, Properties which, MCExecPoint& ep, Boole
             else
                 ep . clear();
         break;
+        // MERG-2014-09-18: [[ ImageMetadata ]] return the image metadata as an array
+    case P_METADATA:
+        if (m_rep != nil)
+        {
+            MCImageMetadata t_metadata;
+            (static_cast<MCLoadableImageRep *>(m_rep))->GetMetadata(t_metadata);
```
livecodeian: This cast is dangerous, as the underlying type of m_rep can easily be something other than MCLoadableImageRep

livecodeian:
Hi Trevor, there is an outstanding issue with this pull request, and it will also require updating to be mergeable into the current development branch. Paging @montegoulding :)

montegoulding:
OK, I'll look at the merge conflicts today but I did resolve and push changes requested back in October. @runrevian am I merging into develop, develop6.7 or develop7.0? It looks like develop and develop6.7 are fairly close now.

engine/src/image.cpp
montegoulding: Ah, forgot to go back and change that. It should be cached rep.

montegoulding:
@runrevian This is my first play with the refactored property handling. It's simpler once you get your head around it but you may want to be extra diligent checking that I've done everything I needed to do given that if there's any docs I don't know where they are.

livecodeali:
The property side of this looks good to me - you managed to pick the most complicated property type to do @montegoulding, especially as you had to add a new macro ;-) 

I'll let @runrevian complete the review from the image object point of view.

engine/src/image.cpp
```diff
@@ -951,6 +953,17 @@ Exec_stat MCImage::getprop_legacy(uint4 parid, Properties which, MCExecPoint& ep
             else
                 ep . clear();
         break;
+        // MERG-2014-09-18: [[ ImageMetadata ]] return the image metadata as an array
+    case P_METADATA:
+        if (m_rep != nil)
+        {
+            MCImageMetadata t_metadata;
+            (static_cast<MCCachedImageRep *>(m_rep))->GetMetadata(t_metadata);
+            MCImageGetMetadata(ep, t_metadata);
```
livecodeian, line 962: This doesn't cover the case of MCMutableImageRep, which is not a cached image rep. Why not just make GetMetaData a pure virtual method of MCImageRep avoiding the need to cast altogether? The MCMutableImageRep implementation can return false, which you would need to check for and handle here.

montegoulding, line 962: OK, I'll take another stab... was sure I had all those classes covered but I guess not.

montegoulding, line 962: Ah I see, I missed it because it's in a different file.

montegoulding:
Thanks @livecodeali it was interesting trying to nut it out... I re-merged because for some reason I ended up committing a dirty ide in the last merge but it seemed to be only an untracked log file which you might want to add to gitignore or add ignore = untracked to gitmodules.

peter-b:
@montegoulding: We've ignored/removed that log file now. See:
- runrev/livecode-ide#11
- runrev/livecode-ide#27

montegoulding:
Guys there's now merge conflicts on this again which it would be nice if GitHub notified me of but it doesn't apparently. Can you give me a date when this is scheduled to be reviewed so I can ensure there's no conflicts on it?

livecodeian:
Hi Monte, I've merged this branch and resolved the remaining conflicts. Thanks.

PR #1260, opened 2014-10-13
[[ bugfix-13642 ]] Improve the documentation for the umask property.

peter-b:
When setting (or retrieving) the umask, it is convenient to use octal
representation, and this is what the documentation currently
recommends.

Unfortunately, LiveCode won't convert base-8 literal integers by
default, so at the moment following the instructions in the
documentation and writing "set the umask to 077" gives a real umask of
0115 (octal).

Update the documentation to clarify this a bit.


peter-b:
Closing this; I'll merge into #1265.

PR #1297, opened 2014-10-17
[[ feature-sendonidle ]] Initial implementation of "send ... when idle"

peter-b:
An naive implementation of the "send <message> to <object> when idle" feature.

Idle priority messages are only handled when there isn't any other work to do.  They are indicated internally by a message with an infinite delay time. If an idle priority message is scheduled while there is already an idle message with the same name and target priority queued, then the queued message is replaced.


peter-b:
Need to do some rework on this branch!

PR #1341, opened 2014-10-24
[[ Bug 13783 ]] Empty should not be among the items of empty

livecodeali:


livecodeali:
Merged into release-7.0.1

PR #1342, opened 2014-10-24
[[ Bug 13770 ]] Make sure every string does not contain the empty string

livecodeali:


livecodeali:
Merged into release-7.0.1

PR #1346, opened 2014-10-24
[[ Bug 13784 ]] Check we have a valid bitmap before exporting

livecodeian:


livecodeali:
Merged into release-7.0.1

PR #1348, opened 2014-10-24
[[ Bug 13776 ]] Make sure player is attached when setting the movie file...

livecodepanos:
...name


livecodeali:
Merged into release-7.0.1

PR #1351, opened 2014-10-24
[[ Bug 13780 ]] Android orientations enum should match system enum

livecodeali:


livecodeali:
Merged into release-7.0.1

PR #1353, opened 2014-10-24
[[ Bug 13350 ]] Fix iOS "snapshot at size" scale transform order

livecodeian:


livecodeali:
Merged into release-6.7.1

PR #1360, opened 2014-10-24
[[ Bug 13751 ]] Make sure volume selector is released on mup()

livecodepanos:


livecodeali:
Merged into release-7.0.1

PR #1361, opened 2014-10-27
[[ Bug 13809 ]] Allow 'width' as param in measureText

livecodeali:


livecodeali:
Merged into release-7.0.1

PR #1378, opened 2014-10-28
[[Bugfix 13840]] Stop MCS_savebinaryfile() from failing if MCresult non-empty.

peter-b:
Related to #1356 and #1374.


livecodeali:
Merged into release-7.0.1

PR #1422, opened 2014-10-31
[[ Bug 13853 ]] Update dragData docs

livecodesebastien:


livecodeali:
Hi Seb, can you convert the line endings to whatever they were before so the diff shows up properly? I'm not sure if it causes any problems, but probably worth doing anyway.

PR #1439, opened 2014-11-04
[[ Bug 13915 ]] Make sure we do not print if user cancels

livecodepanos:


engine/src/exec-printing.cpp
```diff
@@ -709,8 +709,16 @@ void MCPrintingExecOpenPrintingWithDialog(MCExecContext& ctxt, bool p_as_sheet)
 	MCAutoStringRef t_result;
 	if (MCprinter->ChoosePrinter(p_as_sheet, &t_result))
 	{
-		ctxt.SetTheResultToValue(*t_result);
-		MCPrintingExecOpenPrinting(ctxt);
+        // PM-2014-11-04: [[ Bug 13915 ]] Make sure we do not print if user cancels
+        ctxt.SetItToValue(*t_result);
```
livecodeali: Hi Panos, this line is setting the it variable rather than the result- I presume it should be SetTheResultToValue? Or in fact not be there at all, since the result is set either way afterward.

livecodepanos: Hi Ali, yes indeed this line is not needed. I will remove it and update the pull request.

PR #1441, opened 2014-11-04
[[ Bug 13934 ]] Laying out of a field after setting property refactored

livecodesebastien:


livecodeali:
This is a good idea :+1:

PR #1448, opened 2014-11-05
[[ ProperList ]] Pull proper list related changes from widgets branch

livecodeali:


peter-b:
I discussed this with @runrevmark again; I believe the conclusion was that it might be better to merge these changes with the other 8.0 `libfoundation` upgrades rather than pulling them out separately.

livecodeali:
No worries. It's here if you need to use it sooner rather than later anyway :-)

PR #1487, opened 2014-11-17
[[ Bug 14029 ]] Read server scripts even when not memory mapped.

peter-b:
In the original implementation of the server script loader, it was
assumed that:

1) if a file is memory mapped, MCSystemFileHandle::GetFilePointer()
   returns the address of the memory mapped buffer.

2) if a file was not successfully memory mapped,
   MCSystemFileHandle::GetFilePointer() returns NULL.

Assumption 1 is correct.  Assumption 2 is not; in that case,
GetFilePointer() returns a FILE *.

In almost all cases, a local file _can_ be memory mapped, and so the
fallback case is not ever used.  However, the Linux kernel (correctly)
refuses to map a zero-length file.

This patch changes the server script loader to use
MCS_loadbinaryfile(), since this is a clean way to obtain the raw
contents of a file.

### DIFF
--- a/docs/notes/bugfix-14029.md
+++ b/docs/notes/bugfix-14029.md
@@ -1,0 +1,1 @@
+# Read server scripts correctly when they cannot be memory-mapped.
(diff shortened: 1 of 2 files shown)


peter-b:
@runrevmark What do you think of this? The main downside is that we always perform at least one (and possibly two) memory-to-memory copies of the script, which isn't ideal.  Is there a better way to fix this?

livecodesebastien:
Merged in, even though a bit of improvement can be brought in a later step

PR #1498, opened 2014-11-18
Feature: Expanded stackfile low-level input and output

peter-b:
Adds the expanded stackfile file format specification and IO library to the engine.  It is currently only accessible via the special internal commands:

```
_internal stackdir save <state> to <filename>
_internal stackdir load <filename> into <variable>
```

It currently lacks source-info collection (i.e. the `r_source_info` array available via `MCStackdirIOGetState()` is non-useful).


peter-b:
This appears to be irrelevant now, per recent discussions.

montegoulding:
Which discussions? Did I miss something? I still think the spec and engine level stackdir format are a good idea. I'd prefer to see single files for all of an objects properties though because the blob of confusingly named files is complex enough... Lately I've been wondering if CSON is a good format. It can drop the trailing comma which causes unnecessary merge conflicts but it's still light weight and clear like pretty printed JSON. YAML is another one I'm looking at. The only reason I used JSON was because I had a relatively fast parser to use. One advantage to YAML is we could have multi-line text blocks for things like htmlText and graphic points etc. Somebody just needs to implement a blisteringly fast YAML -> Array and Array -> YAML.

PR #1517, opened 2014-11-21
avoid libcairo missing header problem

mwieder:
Avoids the missing header error when compiling libcairo

In file included from ./src/cairo-analysis-surface.c:37:0:
./src/cairoint.h:2827:22: fatal error: memcheck.h: No such file or directory
# include <memcheck.h>


peter-b:
Okay - I found the [forum thread](http://forums.livecode.com/viewtopic.php?f=66&t=21543&start=15#p113795) :+1:

PR #1521, opened 2014-11-23
Oval fix

mwieder:
Previously, non-opaque unclosed ovals would act the same as closed ovals. Thus rotating them would create a visible wedge. This fixes the problem by not creating extra points for the oval if it is not opaque and there is an open arc.

Note that this also includes the makefile patch so that it could be compiled on a 64-bit linux system.


peter-b:
I've rebased this onto the develop-7.0 branch because it seems like it's worth fixing in a 7.x update.

PR #1534, opened 2014-11-25
[[ Bug 13871 ]] Do not specify AI_ADDRCONFIG flag to getaddrinfo(3).

peter-b:
The AI_ADDRCONFIG flag allows getaddrinfo to return much faster in
some circumstances, but it will ignore any IP addresses configured on
the loopback interface.  This can be an issue when attempting to
connect to servers on the local machine without any network
interfaces configured.

See also https://bugzilla.redhat.com/show_bug.cgi?id=721350.


peter-b:
Resubmitting against develop-6.7.

PR #1539, opened 2014-11-27
[[feature-enumvalueref]] libfoundation: Add enumerated type definition.

peter-b:
Add the ability to define enumerated types.  An enumerated type specifies a number of distinct permitted values.  An enumerated value is an immutable and represents one particular permitted value of an enumerated type.


peter-b:
Amalgamated into #1568.

PR #1546, opened 2014-11-27
libfoundation: Add initial implementation of MCProperSet.

peter-b:
This is a very naive implementation of a proper set; an `MCProperSetRef` _is_ an `MCProperListRef` instance that just happens to contain sorted, interned values.


peter-b:
Amalgamated into #1568.

PR #1561, opened 2014-12-03
[[ Bug 14013 ]] Iterate over the keys of the RHS not the LHS in MCMathAr...

livecodestephen:
...rayApplyOperationWithArray()

If you have two arrays `A` and `B`and do

```
multiply A by B
```

then in 6.7 `A` can contain keys that are not present in `B`, they are just left unchanged, but in 7.0 this was raising an error.  The problem was that `MCMathArrayApplyOperationWithArray()` was iterating over the keys of the left hand argument, eg `A`, and not the keys of the right hand side.


livecodestephen:
Ah, this doesn't quite work! :confused:

PR #1568, opened 2014-12-04
🍄 [[feature-typeinfo]] libfoundation: New types and type infos from VCS project.

peter-b:


jameshale:
Ok, I'll bite.
What does the mushroom signify'?
Does it mean this was all some drug induced fantasy'?
Perhaps this conversation is to be kept in the dark as it has just been fed on bulls$%@t?
Or we were really away with the fairies on this but have now come home'?
Intrigued.

PR #1583, opened 2014-12-09
[[ Bug 14176 ]] Make sure sound does not play every other time, but *eve...

livecodepanos:
...ry\* time

### DIFF
--- a/engine/src/aclip.cpp
+++ b/engine/src/aclip.cpp
@@ -867,6 +867,11 @@ void MCAudioClip::stop(Boolean abort)
 #endif
 }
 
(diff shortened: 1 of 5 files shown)


engine/src/aclip.cpp
```diff
@@ -867,6 +867,11 @@ void MCAudioClip::stop(Boolean abort)
 #endif
 }
 
+bool MCAudioClip::isPlaying()
+{
+    return MCPlatformSoundIsPlaying(s_current_sound);
```
livecodesebastien, line 872: That leaves an error when FEATURE_PLATFORM_AUDIO is not defined

PR #1584, opened 2014-12-09
Export all libfoundation symbols defined in foundation.h

livecodefraser:
Conflicts:
    libfoundation/include/foundation.h


peter-b:
Can we call this macro some function of `DLL_?EXPORT` please? "Export" is already a fairly heavily-overloaded verb in LiveCode and it would be nice make things as clear as possible.  Also, other codebases usually use a symbol based on "dllexport", so it using `MCEXPORT` kind-of violates the principle of least surprise...

peter-b:
These changes are already in develop (because they were merged as part of the MLC tech preview merge).

PR #1587, opened 2014-12-10
Feature pseudo object props

livecodeali:
See http://livecode.com/blog/2014/05/28/extending-the-refactored-engine-properties/

Allows definition of properties for non-MCObject objects such as screen, file etc

Implemented in this branch
the rect of screen 
the openmode of file | process 
the name of socket


montegoulding:
This looks interesting @livecodeali

livecodeali:
@montegoulding I think we've ended up ignoring this for now and waiting for open language! The implementations might still come in handy at that point though.

PR #1615, opened 2014-12-17
[[ Bug 14251 ]] Prevent a crash when saving in Legacy format

livecodepanos:


livecodesebastien:
The bug 14251 is a duplicate of the bug 14227, and that fix isn't really right (that is due to uninitialised segment, not an absence of parent).
I'll close the pull request and update the bug report 14251.

PR #1652, opened 2015-01-12
[[ LC Builder ]] Implement various things in encoding module

livecodeali:


libfoundation/src/foundation-string.cpp
```diff
@@ -501,6 +501,98 @@ bool MCStringCreateMutable(uindex_t p_initial_capacity, MCStringRef& r_string)
...
 
+struct MCStringEncodingName
+{
+    const char *m_name;
+    MCStringEncoding m_encoding;
+};
+
+// These names must match the formatting of character encoding names as output
+// by the algorithm described in section 1.4 of the Unicode Consortium TR22.
+static MCStringEncodingName MCStringEncodingNames[] =
```
peter-b, line 754: Do you think it would be a good idea to add the "official" hyphenated versions of the encoding names as synonyms, e.g.:

```
iso-8859-1
utf-8
```

etc.?

runrevmark, line 754: I think this covered by the standard name processing that occurs - Fraser can confirm :)

livecodeali:
This currently crashes when concatenating a list element with a string literal and passing it through to a handler.

peter-b:
Is this a good time to address http://quality.runrev.com/show_bug.cgi?id=13892 by making the builder encoding/decoding syntax return kMCNull if a codec error occurs? I.e.

```
encode "∞" using "iso-8859-1"      -- Result is undefined, not "?"
```

This would be consistent with the behaviour agreed on for `parse`... :wink:

livecodeali:
We need to decide what happens when a replacement character is specified but the data is invalid due to length, eg decoding 3 bytes from utf32 or 1 from utf16. Currently these are decoded as empty.

runrevmark:
I think those count the same as if you had an invalid sequence in utf-8. When decoding, if you specify a replacement char then it is by definition for invalid byte sequences (notionally our internal rep - unicode - is a superset of all encodings so the non-encodable character case doesn't come up).

I'm still on the fence about utf-16 encoded as utf-8 and isolated surrogates in utf-16. Some systems (Java for example) do generate utf-8 with surrogates which is (strictly speaking) incorrect but there's the interoperability aspect to consider (perhaps this case could be considered as a different encoding?)

On the utf-16 front - since we allow code unit access, LC handles isolated surrogates fine so there isn't necessarily a problem. Again, maybe this could be handled via a separate encoding. (The same is true of utf-32)

libfoundation/include/foundation.h
```diff
@@ -1748,15 +1756,28 @@ MC_DLLEXPORT bool MCStringCreateMutable(uindex_t initial_capacity, MCStringRef&
...
+// Evaluate the text encoding from the given string.
+MC_DLLEXPORT bool MCStringEvalTextEncoding(MCStringRef encoding, MCStringEncoding& r_encoding);
+    
 // Encode the given string with the specified encoding. Characters which cannot
 // be represented in the target encoding are replaced by '?'.
 MC_DLLEXPORT bool MCStringEncode(MCStringRef string, MCStringEncoding encoding, bool is_external_rep, MCDataRef& r_data);
 MC_DLLEXPORT bool MCStringEncodeAndRelease(MCStringRef string, MCStringEncoding encoding, bool is_external_rep, MCDataRef& r_data);
+    
+// Encode the given string with the specified encoding. Characters which cannot
+// be represented in the target encoding are replaced by the replacement byte, or,
```
peter-b: "replacement bytes"

libfoundation/include/foundation.h
```diff
@@ -1748,15 +1756,28 @@ MC_DLLEXPORT bool MCStringCreateMutable(uindex_t initial_capacity, MCStringRef&
...
+// if p_replacement is nil, the function returns false.
+MC_DLLEXPORT bool MCStringEncodeWithReplacement(MCStringRef p_string, MCStringEncoding p_encoding, bool p_is_external_rep, MCDataRef p_replacement, MCDataRef& r_data);
 
 // Decode the given data, intepreting in the given encoding.
 MC_DLLEXPORT bool MCStringDecode(MCDataRef data, MCStringEncoding encoding, bool is_external_rep, MCStringRef& r_string);
 MC_DLLEXPORT bool MCStringDecodeAndRelease(MCDataRef data, MCStringEncoding encoding, bool is_external_rep, MCStringRef& r_string);
 
+// Decode the given data, intepreting in the given encoding. Invalid bytes for the
+// target encoding are replaced by the replacement byte, or, if p_replacement is nil,
```
peter-b: "replacement string"

libfoundation/src/foundation-string.cpp
```diff
@@ -1111,28 +1442,46 @@ uindex_t MCStringGetChars(MCStringRef self, MCRange p_range, unichar_t *p_chars)
...
+            t_native_char = self -> native_chars[i];
         else
-            p_chars[i - p_range . offset] = MCStrCharMapToNative(self -> chars[i]);
-
-		t_count += 1;
-	}
+        {
+            if (!MCUnicodeCharMapToNative(self -> chars[i], t_native_char))
+            {
+                if (p_replacement)
```
peter-b: Having `p_replacement` passed in as a pointer suggests that `p_replacement` could be a multi-byte sequence (e.g. `BAD`).

If `*p_replacement` is `\0`, consider abandoning the loop early.

peter-b:
Still need to look into decoding with multibyte replacements.

peter-b:
The story continues elsewhere...

PR #1658, opened 2015-01-13
[[ Builder ]] Basic file I/O support

peter-b:
Add some basic file I/O support to MLC.  Currently supported are:
1. Obtaining standard streams (stdout, stderr, stdin)
2. Getting/setting the entire contents of files
3. Opening read, write and read-write streams for files
4. Writing to streams and reading from them

There are a number of outstanding issues to be addressed in the future (search the source code for `FIXME`).
- `MCStreamIsReadable()` currently succeeds for write-only file streams (and vice versa)
- The only way to read from/write to streams without loss of data is to do so a byte at a time
- There is no way to explicitly close & flush a stream

There are also some Windows-specific issues:
- The Windows implementation has been written, but not compiled or tested.
- There is no long filename support for Windows yet.  Using `/` in paths on Windows is currently forbidden so that we can support long filenames in the future without regressions.
- On Windows, `__MCFileSetContents()` doesn't try to create a file on the same partition as the target path, which means that moving the new file into place is unlikely to be atomic.


peter-b:
@runrevali, @runrevfraser, @runrevben:

As discussed yesterday, I'm going to rework this pull request and remove most of the syntax for working with files and file streams. I'm definitely going to keep:

```
the contents of file tPath
```

However, I am pondering whether it might be a good idea to keep the following items:

```
the output stream
write tData to stream tStream
```

because I think they might be necessary for making unit tests work. Any thoughts?

peter-b:
Closing and resubmitting.

PR #1660, opened 2015-01-13
[[feature-random]] Generic random data services in libfoundation

peter-b:
General rework for providing random numbers in libfoundation.
- Add a new internal API function to libfoundation, `__MCRandomByes()`, which uses platform-specific interfaces to obtain high quality random numbers.
- Port `MCMathRandom()` to use it, and add `MCDataCreateRandom()` for generating random data arrays.
- Remove `MCU_random_bytes()`, and convert all API consumers to use `MCDataCreateRandom()` (because it has an identical prototype).


peter-b:
Hmm -- I still need to fix up the OS X build before this is mergeable, I think

peter-b:
I'm going to rebase this branch against #1658 once I've figured out the final form for the streams support in LC 8 DP 1.

PR #1668, opened 2015-01-16
[RFC] [[feature-filestreamio]] Add more complete file streams

peter-b:
Add some additional syntax to MLC for working with files and streams:

```
-- stdin and stderr
the input stream
the output stream

-- stream reading
read tCount bytes from stream tStream

-- file streams
open read only stream for file tPath
open read write stream for file tPath
open write only stream for file tPath
```

N.b. this is based on #1667, but isn't ready to be merged yet because the stream API needs some work before it's useful (particularly error handling and partial reads).

### DIFF
--- a/libscript/src/file.mlc
+++ b/libscript/src/file.mlc
@@ -4,10 +4,18 @@ This module specifies the syntax definitions and bindings for file i/o operation
 
 module com.livecode.file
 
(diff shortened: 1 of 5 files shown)


peter-b:
If we iterate on file streams I want to implement them in LCB-only code.

PR #1674, opened 2015-01-19
[[ Builder ]] Change lc-compile command-line option format

peter-b:
Previously, `lc-compile` has used long-form command line options with the `-foo` format (one leading dash).  

Many command line tools use a single leading dash to denote a single-dash "short" option, with a corresponding double-dash "long" option.  For example, the `-o` option might correspond to the `--output` option, and so on.

Although lc-compile currently doesn't understand any short options other than `-h`, switching the long options to the double-dash format (e.g. `--modulepath` instead of `-modulepath`) will enable us to add short options in the future.


peter-b:
@runrevali, could you please update the IDE stacks and merge this before the next release? :grinning:

PR #1681, opened 2015-01-19
[[ Bug 14401 ]] mobileStorePurchaseError no longer returns empty on LC 7...

livecodepanos:
....0

### DIFF
--- a/engine/src/exec-store.cpp
+++ b/engine/src/exec-store.cpp
@@ -155,8 +155,9 @@ void MCStoreGetPurchaseError(MCExecContext& ctxt, int p_id, MCStringRef& r_error
     
     t_success = MCPurchaseFindById(p_id, t_purchase);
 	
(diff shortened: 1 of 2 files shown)


engine/src/exec-store.cpp
```diff
@@ -155,8 +155,9 @@ void MCStoreGetPurchaseError(MCExecContext& ctxt, int p_id, MCStringRef& r_error
     
     t_success = MCPurchaseFindById(p_id, t_purchase);
 	
+    // PM-2015-01-19: [[ Bug 14401 ]] Fixed mismerge issue that caused mobileStorePurchaseError to return empty
```
peter-b, line 158: Which merge commit introduced the bug? Could you add the information to the commit message, please?

livecodepanos, line 158: Sorry for the confusion, "mismerge" may not be the correct word. I mean that the original code which worked on 6.7, was not refactored properly so as to work on 7.0.

PR #1689, opened 2015-01-21
Reinstate "module-helper.cpp".

peter-b:
This file is used when compiling the lc-test program (a basic test
runner) to ensure that all necessary objects get linked into lc-test.
Some object files contain symbols that are only accessed at runtime
(via dlsym(3)).  The linker assumes that since it can't detect any
references to the object's symbols, the objects aren't needed, and
prunes them.  The "module-helper.cpp" file references at least one
symbol from each of the problematic objects in order to help the
linker out.

The file was erroneously removed in 116ac1379207, breaking lc-test
(and thus the "make check" target on Linux.  Admittedly,
"module-helper.cpp" is a horrible hack and we should generate it at
build time rather than maintaining it manually.  Unfortunately we
don't have time to do that yet.

### DIFF
--- a/toolchain/lc-compile/src/module-helper.cpp
+++ b/toolchain/lc-compile/src/module-helper.cpp
@@ -1,0 +1,120 @@
+/* Copyright (C) 2003-2013 Runtime Revolution Ltd.
+ 
+ This file is part of LiveCode.
+ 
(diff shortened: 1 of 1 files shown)


peter-b:
Even worse, we have a second copy of this file (differing in subtle ways) in the engine tree. Hurrah.

PR #1711, opened 2015-01-26
[[ Bug 14437 ]] Make sure setting the player filename to empty does not ...

livecodepanos:
...crash LC


peter-b:
Reviewed-by: Peter TB Brett peter.brett@livecode.com

PR #1713, opened 2015-01-26
Builder char as grapheme

livecodeali:
This changes the interpretation of the char chunk from codeunit to grapheme in LiveCode Builder.


peter-b:
Could you please add a "codeunit" chunk, then?  As discussed previously, there are occasions in which you really do want to navigate by codeunit rather than by grapheme -- and since the code was already written to support codeunit chunks, all that would be involved in continuing to provide it would be a search-and-replace...

PR #1715, opened 2015-01-26
[[ CI ]] Don't do multithreaded Travis CI builds.

peter-b:
We've been having some intermittent race conditions on Travis where
make(1) tries to use a file before the rule that creates it has finished
running.

Remove `-j2` from make(1) invocation to prevent this problem from
occurring.  Builds will take a bit longer but will hopefully stop
randomly failing...


peter-b:
Note that this seems to pretty much double the Travis CI build time from about 6 minutes to about 12 minutes.

PR #1722, opened 2015-01-27
[[ Bug 14450 ]] Remove an alert dialog in case of an in-app purchase err...

livecodepanos:
...or (Google Play Store), since this information is present in purchaseStateUpdate message


peter-b:
Reviewed-by: Peter TB Brett peter@livecode.com

I don't think it was really necessary to add the comment there, though! :smiley:

PR #1723, opened 2015-01-27
libscript: Silence "Loading builtin module" message.

peter-b:
Now that builtin modules are working quite well, the messages are more
noisy than useful.


livecodefraser:
I have very thoroughly reviewed your 2 line code-removal patch.

PR #1727, opened 2015-01-28
[CVE-2015-0235] Only use gethostbyname(3) if inet_aton(3) failed. (7.0)

peter-b:
Protects against exploitable buffer overflow in glibc's gethostbyname(3).

http://www.openwall.com/lists/oss-security/2015/01/27/9


livecodefraser:
Fixes the CVE issue but we should investigate why we aren't just using getaddrinfo instead.

PR #1741, opened 2015-01-28
Fix project files

livecodefraser:


engine/Android.mk
```diff
@@ -2,7 +2,7 @@ LOCAL_PATH := $(call my-dir)
 
 include $(CLEAR_VARS)
 
-DERIVED_SRC = ../_cache/mac/Debug/kernel.build/DerivedSources
+DERIVED_SRC = ../_cache/mac/Release/kernel.build/DerivedSources
```
peter-b, line 5: This change doesn't look very relevant...?

livecodefraser, line 5: Well, some of the Android makefiles referenced the Release sources and some the Debug...

peter-b, line 5: Ouch

peter-b:
I'm fairly surprised that there wasn't more horrible brokenness than that in my Windows code, given that I hadn't even compiled it! :stuck_out_tongue_winking_eye:

Reviewed-by: Peter TB Brett peter@livecode.com

PR #1750, opened 2015-01-30
[[ Scriptify revSaveAsStandalone ]] Move revSaveAsStandalone and revSBLi...

livecodeali:
...brary substacks out of revStandaloneSettings stackfile and into script-only stackfiles


livecodeali:
Closing because this has a couple of errors and will be superseded.

PR #1754, opened 2015-01-30
🍄 [[ Builder ]] Further work on encoding/decoding module

peter-b:
This pull request is for continued work on encoding/decoding support for Builder, based on @livecodeali's progress in #1652.

In particular, I have:
- Added support for multi-byte replacements when encoding (only ASCII and native encodings affected)
- Added support for multi-character replacements when decoding ASCII
### Not done yet
- Support for multi-character replacements when decoding (but I can't deal with this without resolving the issues listed below).
### Unresolved issues

There are still some unresolved issues when decoding UTF-8, UTF-16 and UTF-32.
1. When decoding UTF-8, we still use `UTF8ToUnicode()`, which detects and ignores errors with helpful comments like:
   
   ```
   foundation-legacy.cpp:401: // This is an error
   ```
   
   Should there be a `MCStringCreateWithUTF8CharsAndReplacement()` function?  My current feeling is that `MCStringCreateWithBytesAndReplacement()` should be able to detect UTF-8 encoding errors and thus a replacement for `UTF8ToUnicode()` is needed.
2. When decoding UTF-16 and UTF-32, what should be done with trailing bytes? Should they be ignored, should they cause a replacement string to be inserted, or should they always cause an error?  @livecodeali  and I have discussed it, and have had difficulty coming to a solid conclusion.
3. When decoding UTF-16, what should be done about unpaired surrogates? Given that we can safely round-trip an unpaired surrogate through LiveCode's internal representation, I don't think that it makes sense to consider them to be a encoding error.
4. What should be done about UTF-16 encoded as UTF-8? @runrevmark brought this up in the previous pull request.

Several of these issues could be solved by allowing flags to be passed to the codec functions, at the cost of making the API even **more** complicated.


peter-b:
Discussion with @runrevmark today (offline).

About the unresolved issues: the thing that determines a decoding error is if we encounter something while decoding that we can't represent internally. That means:
- For UTF-8, incomplete or too-long sequences get replaced.  But we allow short sequences (as in Modified UTF-8) or surrogates (as in UTF-16-encoded-as-UTF-8). We'll have to add a new e.g. `MCStringCreateWithUtf8BytesAndReplacement()` function.
- For UTF-16 & UTF-32, trailing bytes get replaced.
- Unpaired surrogates get accepted silently.

A couple of additional things were brought up:
- We should convert by grapheme rather than by `unichar_t`.  This may require changes to the way we do ASCII and "native" encoding.  Unfortunately this is likely to be a massive amount of work.
- We need to clean up the `MCString*` API and hide as many of the encoding-specific functions as possible. Also, get rid of the `external_rep` flag wherever feasible.

We need to decide how far we want to go with all of the above before the next release.

PR #1762, opened 2015-02-03
[Bug 14483] Unlock image bitmap to prevent a crash when setting alphaDat...

livecodepanos:
...a

### DIFF
--- a/engine/src/exec-interface-image.cpp
+++ b/engine/src/exec-interface-image.cpp
@@ -617,6 +617,8 @@ void MCImage::SetTransparencyData(MCExecContext &ctxt, bool p_flatten, MCDataRef
 		if (t_success)
 		{
 			MCImageSetMask(t_copy, (uint8_t*)MCDataGetBytePtr(p_data), t_length, !p_flatten);
+            // PM-2015-02-03: [[ Bug 14483 ]] Unlock bitmap to prevent a crash/hang when repeatedly setting the alphaData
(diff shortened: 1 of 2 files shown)


engine/src/exec-interface-image.cpp
```diff
@@ -617,6 +617,8 @@ void MCImage::SetTransparencyData(MCExecContext &ctxt, bool p_flatten, MCDataRef
 		if (t_success)
 		{
 			MCImageSetMask(t_copy, (uint8_t*)MCDataGetBytePtr(p_data), t_length, !p_flatten);
+            // PM-2015-02-03: [[ Bug 14483 ]] Unlock bitmap to prevent a crash/hang when repeatedly setting the alphaData
+            unlockbitmap(t_copy);
 			setbitmap(t_copy, 1.0);
 		}
```
livecodeian, line 623: There's a couple of issues here - you shouldn't attempt to modify a locked bitmap, or use it after unlocking.
This bug seems to be a result of bugfix-13938 which changed the previous copybitmap() call to a lockbitmap() call.

You really shouldn't be using lockbitmap() here. I suggest you revert the patch for bug 13938 and fix the bug in MCImage::copybitmap() that was resulting in it always returning premultiplied image data.

livecodepanos, line 623: Thanks Ian. I will close this pull request, and submit another one against develop-6.7, which will include the updates you suggest.

PR #1765, opened 2015-02-04
com.livecode.unittest: Add new unit testing handlers.

peter-b:
Add a new unit test library that outputs results on stdout using the
Test Anything Protocol http://testanything.org/.

Examples:

```
public handler TestSelf()
    plan 10 tests

    test diagnostic "Normal tests"
    test 2 > 1
    test "Basic test" when true

    test diagnostic "Skipped tests"
    skip test
    skip test "Skipped 2"
    skip test because "Not implemented"
    skip test "Skipped 4" because "Not supported on this platform"

    test diagnostic "Tests which are expected to fail"
    broken test false
    broken test "Failed 2" when false
    broken test false because "broken"
    broken test "Failed 4" when false because "really broken"
end handler
```


peter-b:
Modified the library to be handler-only, and moved it out of the standard library for now.  Should be easy enough to move back in eventually!

PR #1767, opened 2015-02-04
contrib: Add basic Emacs mode for LiveCode builder.

peter-b:
Currently support some basic syntax highlighting.

### DIFF
--- a/contrib/emacs/livecode-mode.el
+++ b/contrib/emacs/livecode-mode.el
@@ -1,0 +1,88 @@
+;; Copyright (C) 2015 Runtime Revolution Ltd.
+
+;; This file is part of LiveCode.
+
+;; LiveCode is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License v3 as published
+;; by the Free Software Foundation.
+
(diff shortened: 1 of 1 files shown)


livecodestephen:
Perhaps it would be better to call this livecode-builder-mode, as you'd want a different mode for livecode scripts.

peter-b:
I was going to add logic to the mode to automatically go into the correct submode depending on the file type, like Emacs `cc-mode` does.

peter-b:
Anyway, aren't we expecting people to use the IDE for editing scripts?

livecodestephen:
Doesn't the same go for LCB scripts!

You can't currently edit CGI scripts in the IDE (at least I don't know how to) and anyway on the server you don't have access to the IDE.

peter-b:
At the moment I don't really have time to fiddle around with highlighting for LC Script; if you'd like to do a bit more work on this and propose an alternative implementation that would be great. :smiley:

peter-b:
Just merging this because I doubt many other people are likely to be using Emacs! @livecodestephen, please send patches. :wink:

PR #1785, opened 2015-02-06
Updated TextWrangler plist so that it matches 'private handler' refs.

trevordevore:
This updated regex will cause private handlers to show up in the BBEdit/TextWrangler function list.


livecodeali:
Hi Trevor, not sure what's happened here but it looks from the diff like nothing has changed. It might be Git playing up though.

livecodeali:
Ahh, no worries, it's because of the line endings. Great, thanks for this!

PR #1797, opened 2015-02-09
Add `server-all` and `server-install` Makefile targets

livecodestephen:
- `server-install` installs livecode server in /opt/livecode/server
- `server-all`  is just the server targets from `all`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,7 @@ lc-test-check: lc-compile lc-test
 ###############################################################################
 # All Targets
 
-.PHONY: all bootstrap thirdparty clean
+.PHONY: all server-all server-install bootstrap thirdparty clean
 .DEFAULT_GOAL := all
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -235,6 +235,14 @@ all: development standalone installer server
 all: revpdfprinter revandroid
 all: lc-run lc-test
 
+server-all: server
```
peter-b, line 230: Could you please add "Server Targets" section of the file for the server-specific targets?  Also, please make `all` require `server-all` instead of listing all the bits of the server stuff twice. :-)

livecodestephen:
As suggested I've created a "Server Targets" section - I'm not sure how far you want to take this but you could factor out some of the other components as well, e.g. you could have a `development-all` target.

peter-b:
Sure, this looks reasonable.

PR #1812, opened 2015-02-10
Feature standalone builder inclusions 7.0

livecodeali:
This is the refactored engine equivalent of https://github.com/runrev/livecode/pull/1805


livecodeali:
In order to make things work with the IDE, this has the min_os_version changes to standalone building included (bug 14422)

PR #1829, opened 2015-02-11
[[ Bug 14587 ]] De-activate the stdout/stderr buffering on Linux

livecodesebastien:


peter-b:
You should only disable the buffering if stdout is a TTY (same for stderr). Use isatty(3) to check. :-)

peter-b:

PR #1836, opened 2015-02-13
[[ Bug 14599 ]] Use exact string comparison for text sort, for consistency with other string operations

livecodeali:
Until we are using context variables for case and form sensitivity in string comparison, all comparison in LCB should be case sensitive.


peter-b:

PR #1850, opened 2015-02-16
[Bug 14489] Use global ref for s_array_list_class

livecodepanos:
to ensure it will be valid next time we use it.


engine/src/mblandroidjava.cpp
```diff
@@ -439,6 +453,7 @@ bool MCJavaInitList(JNIEnv *env, jobject &r_list)
 bool MCJavaFreeList(JNIEnv *env, jobject p_list)
 {
     env->DeleteLocalRef(p_list);
+    free_arraylist_class(env);
```
runrevmark: You don't want to free the arraylist class here - we want to keep the globalref around so we don't have to fetch it each time we create a list.

runrevmark:
Whilst we are fixing this issue we should really make the statics in this file work correctly - Android does not reinit statics on restart of an app which means they can contain stale data.

We need a MCJavaInitialize / MCJavaFinalize method in this file - Initialize should call all the 'class' init functions (init___class), and Finalize the 'free___class' functions). The Initialize/Finalize pair should be added to the android engine startup (in mobile_main in mblandroiddc.cpp - after we attach to the thread).

engine/src/mblandroiddc.cpp
```diff
@@ -1184,6 +1184,10 @@ static void *mobile_main(void *arg)
 		return (void *)1;
 	}
 
+    // PM-2015-02-19: [[ Bug 14489 ]] Init statics on restart of an app
+    if (!MCJavaInitialize(s_java_env))
```
runrevmark, line 1188: If you fail to JavaInitialize, then you should exit (return (void *)1). (Copy the same as the failure clause above).

engine/src/mblandroidjava.cpp
```diff
@@ -36,7 +36,11 @@ static jmethodID s_integer_constructor = nil;
 static bool init_integer_class(JNIEnv *env)
 {
     if (s_integer_class == nil)
```
runrevmark: Don't check for nil in any of this init_\* functions. Make sure that they are only ever called once on startup - so you can guarantee you need to fetch refs again.

runrevmark:
Reviewed-by: @runrevmark.

PR #1851, opened 2015-02-16
[[ Scriptify IDE ]] Restructure libraries - move ide libs to ide submodule

livecodeali:
and move scriptified lib from ide-support/ide-libraries to ide-support.


livecodeali:
I'm going to merge this unreviewed, because there aren't actually any substantive changes here - this is simply a bunch of files moved around so that the IDE works.

PR #1861, opened 2015-02-18
[[ SB Inclusions ]] A couple more tweaks to make things work.

livecodeali:


livecodeali:
Travis CI revealed that revdb would not compile on non-MacOS platforms due to the per-platform DATABASEREC struct definition. I've added the callback setting field to the structs on Linux and Windows.

I've also noticed the library c-string array was not being freed from the deploy parameters.

peter-b:

PR #1864, opened 2015-02-18
🍄 [feature-traits] Adding traits to LiveCode Builder

peter-b:
There's a need in Builder to be able to define operations that are specialized by type.

Traits allow methods to be "added on" to other types and allow in turn allow those other types to be used anywhere where the trait may be used.

Traits provide a reasonably straightforward and clean way to provide extensible standard library syntax without the need for generic handlers and (mostly) preserving the ability to perform compile-time type checking.

[Rendered version](https://github.com/peter-b/livecode/blob/feature-traits/docs/specs/lcb-traits.md)

### DIFF
--- a/docs/specs/lcb-traits.md
+++ b/docs/specs/lcb-traits.md
@@ -97,27 +97,63 @@ implementing the trait for a specific storage type.
 ### Defining traits
 
     public trait type CanFormatAsString
-        method type FormatAsString(in me) as string
+        in method type FormatAsString() as string
     end type
 
-A trait definition is a block containing a list of zero or more `method type`
-declarations.  `method type` declarations take the same form as existing
-`handler type` declarations.
+    public trait type CanParseFromString
+        handler type ParseFromString(in pString as String) as MyType
+    end handler
 
-Two additional symbols are defined within a trait definition:
+A trait definition is a block containing a list of one or more
+`method type` or `handler type` declarations.
 
-1. `MyType` is always equivalent to the type of the value that the method is
-   being called for.
-2. `me` must always be the first parameter a `method` definition or
-   `method type` declaration; it is equivalent to writing `me as MyType`.  The
-   `me` parameter must always be `in` or `inout`.
+An additional symbol, `MyType`, is defined within a trait definition.
+It is always equivalent to the concrete type for which the trait is
+implemented.
+
+#### Trait methods
+
+    public trait type AtomicArithmeticOps
+        inout method type AtomicIncrement() as MyType
+        inout method type AtomicDecrement() as MyType
+    end type
+
+A trait body may contain `method type` declarations.  `method type`
+declarations are similar to `handler type` declarations.
+
+Methods take an implicit argument named `me` of type `MyType`.  This
+is the value for which the method is being called.
+
+The `method type` declaration may have an `in` or `inout` qualifier
+that controls the mutability of the `me` special argument.  If neither
+`in` nor `inout` are specified, `me` is immutable.
+
+Methods may only be invoked for values of concrete types for which the
+trait is implemented.
+
+#### Trait handlers
+
+A trait body may contain `handler type` declarations.  These are
+procedures that are never invoked for a value.  For example, they may
+be used for construction.
+
+Trait handlers must always fulfil the **Relevance Rule**.
+
+#### Relevance rule
+
+The **Relevance Rule** is required to ensure that implementations of a
+trait handler are always distinct.
+
+Trait handler prototypes must therefore mention `MyType`.  This means
+that the prototype must either return `MyType`, or accept a `MyType`
+parameter.
 
 ### Implementing traits
 
     public implementation of CanFormatAsString for T
-        method FormatAsString(in me) as string
+        method FormatAsString() as string
             variable tFormatted as string
-            -- Format me parameter
+            -- Format implicit "me" parameter
             return tFormatted
         end method
     end implementation
@@ -174,7 +210,7 @@ calls trait methods.  Consider the following partially-typed module:
     end handler
 
 Coupled with another module that contains a call:
-`LeakInstanceOfT().FormatAsString()`.  Requiring any implementation of a
+`LeakInstanceOfT():FormatAsString()`.  Requiring any implementation of a
 `public` trait to also be `public` allows this behaviour to be intuitively
 correct.
 
@@ -182,16 +218,19 @@ It may be possible to relax this rule if each module has a different list of
 trait implementations attached to the `T` type.  However, the rule simplifies
 things to start off with!
 
-### Method calls
+### Trait method calls
 
     public handler MCStringEvalFormatAsString(in pValue as CanFormatAsString,
-                                              out rFormatted as string)
-        pValue.FormatAsString() into rFormatted
+                                              out rFormatted as String)
+        pValue:FormatAsString() into rFormatted
     end handler
 
-Selecting an appropriate method implementation for a value is a type of
-namespace operation.  Currently, `.` is used for module-based namespace
-disambiguation.  It is therefore logical to reuse it for method calls.
+The `.` character is reserved for namespace qualification in Builder.
+Method invocation is subtly different because unlike a namespace
+qualification the result cannot be determined entirely statically.
+It's also important to be able to distinguish between calling a method
+on a type object and calling a handler of a type object. The `:`
+character is therefore used for method invocation syntax.
 
     syntax FormattedAsString is postfix operator with precedence 1
         <Operand: Expression> "formatted" "as" "string"
@@ -202,30 +241,130 @@ disambiguation.  It is therefore logical to reuse it for method calls.
 Handler-based `syntax` definitions can be used to provide type-generic syntax
 by using trait types.  Methods do not need to directly used in `syntax` blocks.
 
-### Derived traits
+### Trait handler calls
 
-    public trait type Circle derived from Shape
-        method type Radius(in me) as number
+    public handler MCLogicEvalStringParsedAsBool(in pValue as String,
+                                                 out rParsed as Boolean)
+        put Boolean.ParseFromString(pValue) into rParsed
+    end handler
+
+Trait handler calls are pure namespace operations, and use the `.`
+character.
+
+In the absence of generic functions, trait handlers are most useful as
+a convenience mnemonic for frequently-used constructors.
+
+    syntax StringParsedAsBoolean is postfix operator with precedence 1
+        <Operand: Expression> "parsed" "as" "boolean"
+    begin
+        MCLogicEvalStringParsedAsBool(Operand, output)
+    end syntax
+
+Trait handlers do not need to be directly used in `syntax` blocks.
+
+## Trait type conversion
+
+In general, explicit casting from one type to another is performed
+using the same syntax as for type, parameter and variable
+declarations: `as U`, where `U` is the target type.
+
+### Concrete to trait cast in expression context
+
+    trait type U
+        method type Func()
     end type
+    implementation of U for Number
+        method Func()
+        end method
+    end implementation
 
-One trait can be derived from other traits.  A derived trait inherits all of
-the `method type` declarations.
+    handler DoFunc(in pParam as U)
+        U:Func()
+    end handler
 
-    public trait type Circle derived from (Shape, CanBeEqual)
-        method type Radius(in me) as number
-        method type Centre(in me) as Point
+    public handler main()
+        DoFunc(5)
+    end handler
+
+Implementation-based implicit type conversion occurs wherever
+necessary and unambiguous.
+
+In this case, `DoFunc` requires an parameter of type `U`.  The
+parameter supplied is of type `Number`.  Since `U` is implemented for
+`Number`, the `Number` can be passed to `DoFunc` without an explicit
+type cast.
+
+    trait type U
+        method type Func() as Number
     end type
+    implementation of U for Number
+        method Func()
+            return me + 1
+        end method
+    end implementation
 
-When declaring a derived trait, the **Uniqueness Rule** must be observed.
+    trait type V
+        method type Func() as Number
+    end type
+    implementation of V for Number
+        method Func()
+            return me - 1
+        end method
+    end implementation
 
-#### Uniqueness rule
+    public handler main() as Number
+        return (5 as U):Func() -- returns 6
+    end handler
 
-For a trait `A` and its supertraits `B, ..., k`, no pair of methods declared in
-`A, B, ..., k` may have the same name but a different prototype.
+It is possible for a concrete type to have multiple implementations in
+scope with conflicting method names, leading to ambiguity.  If so, an
+explicit instance cast is required to remove the ambiguity.
 
-This may be relaxed if support for polymorphic dispatch is added to Builder.
+In this case, both the `U` and `V` traits declare a `Func()` method,
+and both are implemented for `Number`.  It is therefore an error to
+invoke `5:Func()` directly; an explicit instance cast to either `U` or
+`V` is required.
 
-## Further work
+### Concrete to trait cast in type expression context
+
+    trait type U
+        handler type Func() as MyType
+    end type
+    implementation of U for Number
+        handler Func()
+            return 1
+        end handler
+    end implementation
+
+    trait type V
+        handler type Func() as MyType
+    end type
+    implementation of V for Number
+        handler Func()
+            return -1
+        end handler
+    end implementation
+
+    public handler main() as Number
+        return (Number as U).Func() -- returns 1
+    end handler
+
+It is possible for a concrete type to have multiple implementations in
+scope with conflicting handler names, leading to ambiguity.  If so, an
+explicit type cast is required to remove the ambiguity.
+
+In this case, both the `U` and `V` traits declare a `Func()` handler,
+and both are implemented for `Number`.  It's therefore an error to
+invoke `Number.Func()` directly; an explicit type cast to either `U`
+or `V` is required.
+
+## Conclusions
+
+Traits provide a reasonably straightforward and clean way to provide extensible
+standard library syntax without the need for generic handlers and (mostly)
+preserving the ability to perform compile-time type checking.
+
+## Appendices
 
 ### Parameterized types
 
@@ -235,7 +374,7 @@ formatting polymorphic compound types such as `list`.  Consider the following
 naïve approach:
 
     public implementation of CanFormatAsString for list
-        method FormatAsString(in me) as string
+        method FormatAsString() as string
             variable tString as string
             put "(" into tString
 
@@ -263,14 +402,6 @@ Similar arguments apply for the `array` type.
 To get full generality and type safety of the `formatted as string` trait
 Builder therefore also needs type specialization (e.g. `list of
 CanFormatAsString`).
-
-## Conclusions
-
-Traits provide a reasonably straightforward and clean way to provide extensible
-standard library syntax without the need for generic handlers and (mostly)
-preserving the ability to perform compile-time type checking.
-
-## Appendices
 
 ### Trait implementations vs. versioning
 
@@ -327,7 +458,7 @@ partially-typed code.
         variable tVar as any
         put pIn into tVar
 
-        tVar.f()
+        tVar:f()
     end handler
 
 If `U` is the only trait in this module and its dependency modules that defines
--- a/docs/specs/lcb-traits.md
+++ b/docs/specs/lcb-traits.md
@@ -98,23 +98,11 @@ implementing the trait for a specific storage type.
 
     public trait type CanFormatAsString
         method type FormatAsString(in me) as string
-    end trait
+    end type
 
-A trait definition is a block containing a list zero or more `method`
-definitions or `method type` declarations.  `method type` declarations take the
-same form as existing `handler type` declarations.  `method` definitions take
-the same form as existing `handler` declarations.
-
-    public trait type CanFormatAsString
-        method FormatAsString(in me) as string
-            return "<unknown>"
-        end method
-    end trait
-
-`method` definitions allow a trait's definition to include default
-implementations of the trait's methods.  Default implementations can call other
-methods of the trait or perform any operations valid for the generic `Self`
-type.
+A trait definition is a block containing a list of zero or more `method type`
+declarations.  `method type` declarations take the same form as existing
+`handler type` declarations.
 
 Two additional symbols are defined within a trait definition:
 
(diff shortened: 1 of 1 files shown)


peter-b:
Obviously, this is a very minimal trait system, and omits neat things like:
- Traits with type parameters
- Associated types
- Generic handlers bounded by traits

etc.

docs/specs/lcb-traits.md
```diff
@@ -0,0 +1,302 @@
...
+            return "<unknown>"
+        end method
+    end trait
+
+`method` definitions allow a trait's definition to include default
+implementations of the trait's methods.  Default implementations can call other
+methods of the trait or perform any operations valid for the generic `Self`
+type.
+
+Two additional symbols are defined within a trait definition:
```
runrevmark: Is it necessary to have an explicit first argument 'me'?

It could potentially be eliminated in favour of 'mutable method' and 'method' - expressing whether 'me' is mutatable or not in the handler. (Which would imply the appropriate mode of the first argument).

docs/specs/lcb-traits.md
```diff
@@ -0,0 +1,302 @@
...
+            return tFormatted
+        end method
+    end implementation
+
+The `me` and `MyType` symbols are also defined inside implementation blocks.
+
+In order for an `implementation of U for T` to be valid, the following criteria
+must be fulfilled:
+
+1. Either `U` or `T` must be local, i.e. defined in the current module.  This
```
runrevmark: This does seem to capture what we want... Just to sanity check this - does the following work (I'm assuming versioning):

Person A creates type FooBar@1.
Person B creates trait Faff@1 and implements for FooBar@1 in the Faff module.
Person A sees trait Faff@1 and sees benefit in it being a standard thing that the FooBar module should implement.
Person A releases type FooBar@2 with implementation of Faff@1 in the FooBar module.

So:

Faff@1 module has a dependence on FooBar@1 module.
FooBar@2 module has a dependence on Faff@1 module.

I think this avoids a circular dependency as long as module imports are fine grained - i.e. you never import a module, only ever a versioned symbol in a module. However, I'm wondering if there might be implementation issues with this...

(The situation I'm modelling here is one where over-time, well-established traits will likely want to move to be implemented in well-established type modules.)

peter-b: Is it currently permissible for the module dependency graph to contain cycles?  I.e. is the following allowed?

```
module A
use B
public handler foo(in pParam)
    if pParam then
        bar()
    end if
end handler
end module

module B
use A
public handler bar()
    foo(false)
end handler
end module
```

docs/specs/lcb-traits.md
peter-b: You might want to add `mutable` to the normal list of parameter modes.  For example, an in-place swap trait:

```
trait type Swap
    method type Swap(mutable me, mutable pOther as MyType)
end trait
```

In a later version of the trait implementation, it may be interesting to permit the `me` parameter to be absent.  For example (probably don't want to call static methods with a `.`, but bear with me):

```
trait type CanParseFromString
    method type ParseFromString(in pString as string) as MyType
end trait

-- Implement CanParseFromString for integer etc.

public handler main()
    variable tInt as integer
    integer.ParseFromString(element 1 of the command arguments) into tInt
end handler
```

runrevmark: That wasn't quite what I meant :)

I was suggesting that trait methods don't necessarily need an explicit 'me' - we could use the same formalism as object methods in other languages where me/self/this is implicit.

If we did add the equivalent of 'static' methods it would probably make more sense to have it as an explicit type of method rather than implicit from the argument list.

peter-b: That's certainly an option. I'm not sure it makes more sense, but I'm not particularly devoted to inferring static-ness from the lack of a `me` argument.

I have two other arguments in favour of the explicit `me` parameter:
1. It makes it clearer when reading the code that the `me` is effectively a handler parameter -- i.e. you can see where it comes from and what its scope is. It makes it easier to move from writing handlers to writing methods and it increases commonality in syntax and concepts.
2. As mentioned above, it provides a convenient place to add kind-of-slot modifiers (as in the `Swap` example above).

I'm not totally wedded to the idea of explicit `me` parameters, but I've found it works very intuitively in Python and (more recently) in Rust.

docs/specs/lcb-traits.md
runrevmark: Cycles aren't currently allowed, no - the reason I bring this up is that it will influence how versioning needs to work as this seems like a potentially reasonable thing to try and accommodate.

Traits combined with the strict module namespace essentially de-centralize policy, but the rule about placement of traits, types and implementations of types means that it would be difficult for a type producer to adopt a trait natively that had previously been implemented externally (which they may want to do if they can implement it more efficiently - I'm assuming here that types will be opaque outside of the module which produces them, and thus only the defining module would have access to its internals).

If you imagine a module being composed of separate parts (source-files) and versions can be specified per-part (rather than per-definition), then the compiler would compile the module parts in version order - first producing the version 1 interface, then the version 2 interface etc. The idea here being that when it compiles version 2 it can use other modules which only depend on version 1.

Perhaps there's another way to achieve the above - or maybe I'm overthinking things...

peter-b: > it would be difficult for a type producer to adopt a trait natively that had previously been implemented externally

I don't think so... given that all the method types of a trait _must_ be exported in order for the trait to be usable, there's nothing stopping the type producer from just copying the definition into her module.

Another way to think about it: it's exactly as difficult to adopt a third-party trait as it is to adopt a third-party function.

docs/specs/lcb-traits.md
```diff
@@ -0,0 +1,302 @@
...
+    end handler
+
+    public handler main()
+        variable tVar as any
+        put 15 into tVar        -- contents of tVar is an integer
+        Confuse(tVar) into tVar -- contents of tVar is a T trait object
+        add 1 to tVar           -- error?
+    end handler
+
+It is not clear what the correct behaviour is here.
```
runrevmark: What the 'correct' behavior here does depend on how we view a trait T.

One view (distinct) would be that implementing a trait T for type U creates a new type (T, U) where (T, U) conforms to any other type (T, V) and U bridges to (T, U). With this point of view the above would give an error (unless 'add' had an overload for T). This would require some sort of indirection to implement (e.g. a trait object) which means that an implicit (bridging) operation is needed when going from U to (T, U).

Another view (interface) would be that implementing a trait T for type U makes U conform to T. WIth this point of view the above would not give an error as the runtime type of tVar would remain 'integer'. This would mean that the method invocation operator '.' would actually be a typed operation 'call method on value for trait T'.

Advantages of the 'distinct' view is that it gives a stronger interface guarantee - however, the presence of an implicit operation which is based on a type means that you would get different behavior if you removed type annotations from slots (inserting explicit 'as' operators as needed).

Advantages of the 'interface' view is that you can remove type annotations from slots (inserting explicit 'as' operators as needed) and you will still get the same behavior - however this is at the potential expense of subtle binary interface changes. (Java has a similar issue with the presence of interfaces - subclasses can have extra interface implementations, so the set of interfaces implemented by an object returned by a method can change even though the signature of the method does not).

docs/specs/lcb-traits.md
runrevmark: > I don't think so... given that all the method types of a trait must be exported in order for the trait to be usable, there's nothing stopping the type producer from just copying the definition into her module. 

Wouldn't that create an entirely separate trait though? i.e. Any consumers anywhere of the trait implemented in Faff for FooBar, would be relying on the trait defined in Faff. (I'm assuming that the implementation would essentially need to add each trait implementation to the list implemented by a given storage type - otherwise generic 'format list as string operations wouldn't work...).

docs/specs/lcb-traits.md
```diff
@@ -0,0 +1,302 @@
...
+    public trait type T
+        method F(in me)
+    end trait
+
+    public handler F(in pIn as T) -- pIn is a T trait object
+        variable tVar as any
+        put pIn into tVar         -- tVar is a ?? (out of lexical scope)
+
+        tVar.F()                  -- error?
+    end module
```
runrevmark: Applying the two points of view to the above example would mean that in the 'distinct' view - tVar.F() would work; but in  the 'interface' view you'd actually get a compile-time error as there is no way to infer where F() should be looked for.

docs/specs/lcb-traits.md
peter-b: Note that this is analogous to an downcast in traditional OO systems.

```
class X
    method f()
class Y extends X
    method g()

function h(X p) -> X

var a -> X
h(a).f()  -- Always
h(a).g()  -- Downcast to Y is required.
```

In compile-time typing world, this shouldn't successfully compile.  In duck-typing world, there are no type restrictions at all, so this will work fine as long as `h()` returns _any_ value with a `g()` method. We're somewhere in between.

docs/specs/lcb-traits.md
peter-b: > Wouldn't that create an entirely separate trait though? i.e. Any consumers anywhere of the trait implemented in Faff for FooBar, would be relying on the trait defined in Faff.

Yes, just like if there'd been a useful handler function of `FooBar` copied into the original module. It'd be an entirely different handler, and any consumers of the handler implemented in `Faff` would be relying on it to be defined in `Faff`. But that's okay because they'd just use the earlier version of the `FooBar` module API which doesn't have the colliding symbol, right?

> I'm assuming that the implementation would essentially need to add each trait implementation to the list implemented by a given storage type - otherwise generic 'format list as string operations wouldn't work...

Sometimes it's useful to have private implementations.  For example, you might want to implement `CanFormatAsString` for a `public type T` for debugging purposes internally to your module but also want to prevent users of your module from from formatting `T` (maybe because your debugging output is a. hideous and b. exposes implementation details you'd rather not exhibit).

docs/specs/lcb-traits.md
peter-b: This example is why I personally favour the 'distinct' view. :wink:

docs/specs/lcb-traits.md
peter-b: Maybe detailed discussion of type versioning issues (including trait type versioning issues) should be postponed until there's some sort of concrete specification of how versioning is going to work.

peter-b:
@runrevmark I've been thinking about the issues we discussed last night over breakfast. I think that the way traits are proposed to work is going to have to change in order to enable duck-typed code to use strictly-typed code.  However, this is going to pose its own problems.

I suspect that the syntax should be `trait type ... end type` rather than the current `trait type ... end trait` and I'll update the spec to match.

### Storage typeinfos must collect all implemented traits.

Firstly, it will be necessary to attach a list of implementations to the value typeinfo (and thus to the value).  Consider the following code, which is duck-typed by virtue of its use of unparameterized `list`:

```
public trait type U
    -- ...
end type

public handler FizzList(in pList as list)
    variable tElement as U
    repeat for each element tElement in pList
        -- ...
    end repeat
end handler
```

Unless each value in `pList` has a list of all traits implemented by its type attached to the corresponding typeinfo, it is impossible to determine whether any particular element can conform to `U`

### Method name collisions are forbidden

Secondly, with traits that support duck typing, name collisions must be forbidden.  Consider the following fully-typed code:

```
trait type U
    method f(in me)
    end method
end type

trait type V
    method f(in me)
    end method
end type

type T

implementation of U for T
end implementation

implementation of V for T
end implementation
```

This is fine; although there is an ambiguity if you call `t.f()` for some `variable t as T`, this can be detected at compile time and worked-around by simply casting e.g. `variable tU as U`.  This is pretty similar to the current dynamic handler call restriction.

Unfortunately, in duck-typed code it will always be ambiguous to declare a `variable t`, store a value of type `T` in it, and subsequently call `t.f()`.

`implementation of U for T` should therefore be a compilation error if the preexisting methods of `T` have any names that collide with method names of `U`.

### `implementation` access bounded by trait type and concrete type access.

Consider the following partially-typed module:

```
public trait type U
    method f(in me)
    end method
end type

type T

implementation of U for T
end implementation

public handler MakeT() as any
    variable t as T
    -- Construct t
    return t
end handler
```

According to duck-typing rules, external code should be able to call `MakeT().f()`.  Therefore the implementation of a `public` trait for a `private` type must be `public`.

The main question is whether the converse is also true:

```
trait type U
    method f(in me)
    end method
end type

public type T

implementation of U for T
end implementation

public handler MakeU() as any
    variable u as U
    -- Construct value of type T and store in u
    return u
end handler
```

This clearly leaks a value of type `U`, so it should be possible to call `MakeU().f()` in duck-typed code.

I propose that an `implementation of U for T` must be `public` if either `U` or `T` is `public`.

**Note:** I'm not sure that it even makes sense to have private types (either trait or concrete), especially since types are to be considered opaque outside the defining module.

### Some fully-typed code may require duck-typed rules

Consider the following apparently-fully-typed module:

```
trait type U
    method f(in me) as undefined
    end method
end type

trait type V
    method g(in me) as undefined
    end method
end type

type T as integer

implementation of U for T
end implementation

implementation of V for T
end implementation

handler RequiresDuckTyping(in pParam as U) as undefined
    pParam.g()
end handler
```

The `RequiresDuckTyping()` handler is valid with duck-typing rules but invalid with strict typing rules.  Therefore it's not possible to infer whether a piece of code should have strict typing rules applied to it just by checking whether every slot is typed.

runrevmark:
Just comments on the above (after internal discussion)...

Storage typeinfos will definitely need to keep a list of the traits which are implemented for it - this is a global thing that would happen as modules load into the environment.

I don't see a problem with method name collisions because I think it is better that trait access is explicit (at a low-level at least) - i.e. the underlying access operator would be (v as T).Method() (note that I'm not saying this is the syntax, just the VM operation that would invoke a method of a trait).

In terms of access control...

In order to implement a public trait privately, it would mean that the trait list attached to the type T would have to be different locally in the module compared to globally. However, this would still mean that the private trait could leak if the 'distinct' view is taken (but then this perhaps does not matter - you can leak private subclasses of public classes in Java I believe - which is perfectly reasonable for specialized implementation). Obviously, it is easy enough to restrict public traits to having public implementations in the first instance.

There's no real implementation issue with private traits of a public type since only public things have a 'name' outside of the module. Attaching a private trait to a public type would cause a 'uniquely named' trait to be attached to type's global typeinfo list which no other module would be able to reference. Of course, if the above scheme is implemented, then private traits would be attached to the per-module type trait list which strengthens this even further.

docs/specs/lcb-traits.md
runrevmark: Well developing a concrete specification of how versioning needs to work depends on knowing what it needs to account for ;)

A module import is a triple (module, version, symbol) - so a compiled module has a list of these.

Perhaps I should clarify my example above as I'm not sure your proposed resolution works in this case (from my understanding of what you wrote it would mean that any modules which use the specified trait would need to be updated to also use the trait that had been copied into the type's module):
- Miss Foo creates a type org.example.foo.Type@1
- Mr Bar creates a trait org.example.bar.Trait@1 and implements it for all popular types including org.example.foo.Type@1
- Mr Bar's trait becomes widely popular so lots of people use it
- Miss Foo sees this and realizes that she can implement the trait much more efficiently in the module implementing Type@1 but obviously she cannot do so because either a trait has to implemented next to the type or next to the trait (you can't do both because it would result in a cyclic dependency).

So I'm wondering if a suitably designed way of compiling versioned modules solves this problem: Miss Foo creates org.example.foo.Type@2 and implements org.example.bar.Trait@1.

Thus:
- org.example.foo@1 does not depend on org.example.bar@1
- org.example.bar@1 depends on org.example.foo.Type@1
- org.example.foo@2 depends on org.example.bar.Trait@1
- org.example.bar@1 does _not_ depend on org.example.foo@2 as it was compiled before the latter was created

Now, when Mr Bar comes to update his module he will get a cyclic dependency error - this will force him to version his module. All he has to do is move org.example.bar.Trait into a 'version 1' part which uses org.example.foo@1 - thus removing it from version 2.

peter-b: @runrevmark I just don't think that this use-case can be permitted.

```
module org.example.baz
use org.example.foo@2
use org.example.baz@1

handler insanity(in p as Trait)
    p.f()   -- Error: conflicting implementations of Trait for type Type
end handler
end module
```

**However**, there is an alternative. Suppose in `org.example.foo@2`, Miss Foo adds an `org.example.foo.TraitPrime@2` trait and implements it for `org.example.foo.Type@1`.  She can then contact Mr Bar and persuade him to release `org.example.bar@2` with the implementation of `Trait` for `Type` updated to be a thin wrapper around the implementation of `TraitPrime` for `Type`.  The Locality Rule is respected, and the resulting code is pretty efficient (especially with an optimising VM).

runrevmark: As discussed offlline - not doing anything clever with versioning initially doesn't do any harm - it will just mean that there will be times distinct authors will have to communicate... It would be an interesting possibility to explore at some point though (in a world of de-centralized policy, communication between parties cannot always be guaranteed).

docs/specs/lcb-traits.md
runrevmark: It occurs to me that the distinct view and interface view are not incompatible - at a VM implementation level I think they come down to how fine-grained the operations are.

For the interface view, you just need an (Value as Trait).Method() operation.

For the distinct view, you need two operations 'Value as Trait' and 'Value.Method()' (the point here being that for 'Value as Trait' to actually have the desired effect it has to produce something which has a different runtime type attached to it than Value).

peter-b:
I've updated and greatly revised the traits spec to reflect today's very interesting discussions!

docs/specs/lcb-traits.md
```diff
@@ -0,0 +1,364 @@
...
+
+* Module `A` defines and type `T`
+* Module `B` implements `CanFormatAsString` for `T`
+* Module `C` implements `CanFormatAsString` for `T` (differently)
+* Module `D` uses modules `B` and `C` and attempts to format an instance of `T`
+  as a string.  Which implementation should be used?
+
+By requiring either the trait type or the concrete type to be defined locally
+to the implementation, this diamond problem can never occur.
+
```
livecodestephen, line 194: I'm not convinced that this isn't overly restrictive.  Consider the following:
- Alice writes a library containing a type `T`.  From Alice's point of view there is no generic way of representing a `T`
  as a string, so she cannot provide an implementation of `CanFormatAsString`.
- Bob is writing an app that uses Alice's library.  From Bob's point of view there is a clear way of representing a `T` as a string in his app and so he want's to implement `CanFormatAsString`.

The Locality rule prevents Bob from providing this implementation.  Clearly Bob can use a function to do the formatting, but this adds to the complexity of Bob's code.  I'm open to being convinced that this isn't a big deal, but from Bob's point of view the ability to provide an implementation of an external trait for an external type would be nice.

livecodefraser, line 194: The issue is that both Bob and Carol might implement the trait for T. Which implementation gets used when Dave's module tries to format a T as a string?

livecodestephen, line 194: That's why I said that Bob was writing an app, so there would be no one else importing his modules.

livecodestephen, line 194: There's a balance to be made here between theoretical simplicity of the language (avoiding the diamond inheritance problem is definitely a worthwhile goal) and the practicalities for the end user.

livecodestephen, line 194: Thinking about this overnight, I suppose the solution would be for Bob to define a derived type that doesn't actually directly add anything to Alice's `T`, he could then add his own implementation of `CanFormatAsString`.  You could even have some syntactic sugar to do this, eg `import com.example.Alice.T`.

runrevmark, line 194: @livecodestephen: I'm not sure this problem is resolved by derivation - as the derived type BobT will be distinct from the original AliceT. The result is that any modules which consume / produce AliceT will still be producing AliceT's - there would need to be explicit code in the all the modules wanting to get the CanFormatAsString effect on AliceT's to turn AliceT's into BobT's (recursively through any compound types too).

Stepping back slightly, the problem here comes down to the definition of 'world'.

Modules as we've currently been thinking about them have to interact with a world they do not know - this makes the locality rules required as it prevents (ahead of time) any potential conflicts arising in this most general context (which is what we have in the context of LCB-as-an-extension-language).

In the app example though, the person producing the app has complete control over the 'world'. The result of this is that only 'modules which might be used in an unknown world' have to conform to the locality rules - if you are compiling a module for a world you control then there is no problem... At the 'link' (or 'run' step) you might get an error if you've provided conflicting implementations of a trait - just as you would if you have an implementation of function twice in two object files and try to link them together.

Upshot: in becomes a property of the 'scope' of the module which determines whether you can define an implementation for a public trait and public type you do not define, and modules which don't have an appropriate scope cannot be used in a wider context (i.e. used by a third-party).

(Of course, there is a generalization here... It is not beyond the realms of possibility that I might want to provide a module which is intended for use by third parties but within its 'internal world of operation' you want to adapt/specialise third-party types and traits but that's perhaps something for later consideration!)

livecodestephen, line 194: Is it not the case that a `BobT` is an `AliceT` and, therefore, can be used where ever an `AliceT` can be used?  If a user wanted to use Bob's implementation of `CanFormatAsString` then they have to specify that somehow and using `BobT`s instead of `AliceT`s seems a reasonable way to do that.

A related question is how do traits work for derived types, are they "virtual"?  So if you had a base type and a derived type both implementing a trait will code that's implemented using the base type always use the base types implementation.

I suppose if traits are not virtual then you probably can't use a derived type to solve my problem.

runrevmark, line 194: Yes - for a 'derived' type (which we still need to define exactly what that means in terms of traits) a BobT 'will be' an AliceT (in some sense). However, the use-case is using a third-party type and a third-party trait and joining them by your implementation of the trait for the type... This means that the values of type AliceT will be coming from somewhere outwith your module - and thus will be AliceT's and not BobT's. Thus, if you want your trait to work whenever you get an AliceT you need to 'do something' to make it a BobT. (It would probably have to be aggregation rather than derivation).

livecodestephen, line 194: Ah, I was think of Bob creating `BobT`s and sending them into Alice's code, but of course they can flow in the other direction as well.

peter-b:
@runrevmark @livecodestephen @livecodefraser: I've updated the spec again to reflect some discussions over the last couple of weeks.  For example, I've removed `derived` traits (for now), removed explicit `me` arguments, added static methods ("trait handlers") and changed the method call syntax from `tVar.f()` to `tVar:f()`.

I think the main issue still to resolve for the initial version of traits is whether casting from a trait type to a concrete type should be permitted, and if so, when.

PR #1875, opened 2015-02-20
libfoundation: Correct implementation of isinf() on Windows.

peter-b:
Previously, the fallback implementation of isinf(3) on Windows platforms
would return:
- 1 if the argument is an infinity or NaN.
- 0 otherwise

This conformed to neither C++, C99 nor POSIX specification:
- C++ requires `isinf(x) == 0` when `x` is NaN.
- C99 and POSIX require `isinf(x) == 0` when `x` is NaN and `isinf(x) == -1`
  when `x` is negative infinity.

This patch alters the behaviour of the fallback implementation of isinf(3) to
match POSIX behaviour (and thus also conform to C++ and C99).

### DIFF
--- a/libfoundation/include/foundation-math.h
+++ b/libfoundation/include/foundation-math.h
@@ -30,7 +30,9 @@
 #    define NAN ((float)(INFINITY*0.));
 #  endif
 #  include <float.h>
-#  define isinf(x) (!_finite(x))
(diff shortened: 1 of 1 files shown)


livecodefraser:
Reviewed-by: @runrevfraser

PR #1894, opened 2015-02-24
Merge 6.7.4 rc 1

livecodesebastien:


engine/src/mblstore.cpp
```diff
@@ -74,8 +74,10 @@ static struct {const char *name; MCPurchaseState state;} s_purchase_states[] =
     {"unverified", kMCPurchaseStateUnverified},
 };
 
-// we maintain here a list of known pending purchases
+// we maintain here a list of known pending purchases, and a list of completed purchases
 static MCPurchase *s_purchases = nil;
+#error "NEed to finish"
```
livecodesebastien: Still need to add a cleanup function for s_completed_purchases...
Not suitable for merging as for now

livecodesebastien:
Issues fixed, the branch is ready to be reviewed.

engine/src/util.cpp
```diff
@@ -3182,6 +3186,110 @@ bool MCU_random_bytes(size_t p_bytecount, MCDataRef& r_bytes)
...
+            return nil;
+    }
+    else
+    {
+        uindex_t t_last_slash_index;
+        if (!MCStringLastIndexOfChar(*t_path, '/', UINDEX_MAX, kMCStringOptionCompareExact, t_last_slash_index))
+            t_last_slash_index = MCStringGetLength(*t_path);
+
+        if (!MCStringFormat(&t_filename, "%*@/%s", MCRangeMake(0, t_last_slash_index), MCcmd, *t_path))
+            return nil;
```
livecodeali, line 3261: There are a couple of issues here: MCStringLastIndexOfChar should be called on MCcmd (and MCStringGetLength).
The format string for MCStringFormat is wrong - the %s should be a %@. Also the MCRange parameter for MCStringFormat has to be a pointer to a range rather than a range.

engine/src/mblstore.cpp
```diff
@@ -347,9 +354,22 @@ void MCPurchaseNotifyUpdate(MCPurchase *p_purchase)
...
+    if (s_completed_purchases == NULL)
+        /* UNCHECKED */ MCListCreateMutable('\n', s_completed_purchases);
+
+    MCListAppend(s_completed_purchases, p_purchase -> prod_id);
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 
+void MCPurchaseClearList(void)
+{
```
livecodepanos: I don't think we need this method, since it is not used anywhere. We have MCPurchaseClearPurchaseList() that does the same job.

engine/src/mblstore.h
```diff
@@ -172,4 +173,6 @@ bool MCStoreRequestProductDetails(MCStringRef p_product_id);
 
 void MCPurchaseNotifyUpdate(MCPurchase *p_purchase);
 
-#endif //MBLSTORE_H
\ No newline at end of file
+void MCPurchaseClearList(void);
+
```
livecodepanos: Same here :)

engine/src/mblstore.cpp
```diff
@@ -141,14 +154,8 @@ bool MCPurchaseStateToString(MCPurchaseState p_state, const char *&r_string)
...
-{   
-	MCAutoListRef t_list;
-	if (!MCListCreateMutable('\n', &t_list))
-		return false;
-	for (MCPurchase *t_purchase = MCStoreGetPurchases(); t_purchase != NULL; t_purchase = t_purchase->next)
-        MCListAppendFormat(*t_list,"%@", t_purchase -> prod_id); // we want a list of product IDs
-	
-	return MCListCopyAsString(*t_list, r_string);
+{
+    return MCListCopyAsString(s_completed_purchases, r_string);
```
livecodepanos: This will crash if s_completed_purchases == nil (i.e. if the user has not made/restored any purchases). In that case, we want to return true and assign empty to r_string

PR #1901, opened 2015-02-25
[Coverity] Add Unreachable annotations to the blend-mode switch statements

livecodefraser:
All switch statements are exhaustive over the valid input parameters, but this cannot be proven by static analysers like Coverity. Add default statements with an MCUnreachable() annotation.

Fixes Coverity defects 15998 16125 16265 16375 16481 16499 16586.


peter-b:

PR #1912, opened 2015-02-25
Merge 7.0.4 rc 1

livecodesebastien:


livecodesebastien:
Hi Monte - thanks a lot for the quick feedback, and sorry for having left these issues. That will get fixed in develop-7.0 as soon as possible, and be included in LC 7.0.4-rc-1.

PR #1915, opened 2015-02-26
Fix to the CEF external that broke for Win32 after merging 7.0

livecodefraser:


livecodesebastien:
Reviewed-By: sebastien.nouat@livecode.com

PR #1919, opened 2015-02-26
[Build] Linux: Compare stripped compilers for bootstrap verification.

peter-b:
After going to the the trouble of stripping the stage 3 and stage 4
compilers for comparison, it might make sense to actually compare the
stripped binaries rather than the unstripped ones.


livecodefraser:
Reviewed-by: Fraser Gordon

PR #1928, opened 2015-02-27
Fix incorrect use of TARGET_OS_IPHONE macro

livecodefraser:
This also revealed a bug where we don't include the correct headers on OSX.


peter-b:
Although this is also wrong in develop-7.0, we decided that since it wasn't actually breaking anything in 7.x it would be better to target it to develop.

Also, all the best fixes mostly remove code.

PR #1937, opened 2015-03-02
[[ Bug 14720 ]] Ensure 'converted' out parameter is always set when necessary when converting types for lcb handler calls

livecodeali:


livecodefraser:
Reviewed-By: Fraser Gordon

PR #1943, opened 2015-03-02
[Bug 14677] Rework "last offset" string operation for native strings.

peter-b:
Rework "the last offset" syntax implementation for native strings, fixing a number of issues.
- Avoid `uindex_t` underflow errors when searching for 1 character string in 1 character string.
- Start searching at the last position at which the needle will fit into the range, rather than at the end of the haystack.  For example, when searching for "xxxx" in the string "xxxx.", start searching at offset 2 rather than offset 5.
- Limit the underlying "shared prefix" operation to looking at characters in the specified range of the haystack string, rather than allowing the needle to be matched against characters beyond the end of the range.


livecodeali:
Reviewed by me.

PR #1948, opened 2015-03-02
[Bug 14630] com.livecode.list: Add "offset" and "index" syntax for lists

peter-b:
Add a number of new list manipulation syntax definitions and corresponding implementations.

Finding elements in lists:

```
index of _ in _
first index of _ in _
last index of _ in _
index of _ before _ in _
first index of _ before _ in _
last index of _ before _ in _
index of _ after _ in _
first index of _ after _ in _
last index of _ after _ in _
```

Finding subsequences in lists:

```
offset of _ in _
first offset of _ in _
last offset of _ in _
offset of _ before _ in _
first offset of _ before _ in _
last offset of _ before _ in _
offset of _ after _ in _
first offset of _ after _ in _
last offset of _ after _ in _
```

Since strings/data are self-flattening sequences of chars/bytes, the `index` operations are added to strings and data, identical to the corresponding `offset` operations.  The `offset of chars` and `offset of bytes` syntax is removed; it's the default chunk type for and should be implicit.

In order to make `index` available as a syntax keyword, the deprecated `index` and `uindex` types are removed, and `LCIndex` and `LCUindex` should be used instead.

### DIFF
--- a/libfoundation/include/foundation.h
+++ b/libfoundation/include/foundation.h
@@ -2881,6 +2881,7 @@ MC_DLLEXPORT bool MCProperListLastIndexOfElementInRange(MCProperListRef list, MC
 
 MC_DLLEXPORT bool MCProperListFirstOffsetOfList(MCProperListRef list, MCProperListRef p_needle, uindex_t p_after, uindex_t& r_offset);
 MC_DLLEXPORT bool MCProperListFirstOffsetOfListInRange(MCProperListRef list, MCProperListRef p_needle, MCRange p_range, uindex_t & r_offset);
+MC_DLLEXPORT bool MCProperListLastOffsetOfListInRange(MCProperListRef list, MCProperListRef p_needle, MCRange p_range, uindex_t & r_offset);
 
 MC_DLLEXPORT bool MCProperListIsEqualTo(MCProperListRef list, MCProperListRef p_other);
 
--- a/libfoundation/include/foundation.h
+++ b/libfoundation/include/foundation.h
@@ -2877,6 +2877,8 @@ MC_DLLEXPORT bool MCProperListRemoveElements(MCProperListRef list, uindex_t p_st
 MC_DLLEXPORT bool MCProperListFirstIndexOfElement(MCProperListRef list, MCValueRef p_needle, uindex_t p_after, uindex_t& r_offset);
 MC_DLLEXPORT bool MCProperListFirstIndexOfElementInRange(MCProperListRef list, MCValueRef p_needle, MCRange p_range, uindex_t& r_offset);
 
+MC_DLLEXPORT bool MCProperListLastIndexOfElementInRange(MCProperListRef list, MCValueRef p_needle, MCRange p_range, uindex_t & r_offset);
+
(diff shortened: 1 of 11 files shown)


runrevmark:
(Replicated from BZ to make it more visible)

I'm not sure I'm 100% happy with 'the offset of list' - as it is slightly confusing.

Could we perhaps consider using 'index' and 'offset' as the two distinct operations?

INDEX

the index of [ <chunk> ] tValue in tSequence:
  returns the first (1-based) index of tValue in tSequence

Here tSequence is considered to be a sequence of <chunk> and tValue should be compatible with <chunk>. It returns 0 if tValue is not found in tSequence.

OFFSET

the offset of [ <chunk> ] tValue in tSequence:
  returns where the first instance of tValue starts in tSequence

Here tSequence is considered to be a sequence of <chunk> - it does a subsequence match.

So:

the index of [ char ] <character> in <string>
the index of [ byte ] <byte> in <data>
the index of [ element ] <any value> in <list>

And:

the offset of [ chars ] <string> in <string>
the offset of [ bytes ] <data> in <data>
the offset of [ elements ] <list> in <list>

I think this captures the difference between element-wise match and substring match. For strings 'offset' and 'index' are almost the same thing (strings being self-flattening sequences of chars, chars being strings ;)) - except that we can be much more strict with 'index'.

runrevmark:
Indeed, both offset and list are target type driven - so I wonder if 'char', 'byte' and 'element' are needed at all:

If the haystack is a string, then the needle must be a char (string)
If the haystack is a data, then the needle must be a byte (data)
If the haystack is a list, then needle can be any value (index) or must be a list (offset)

I think we might have got slightly muddled up with the string-chunk index/offset operation - where you _do_ need the clarifying 'chunk' type as chunks (items, lines etc.) are all strings.

runrevmark:
@peter-b: Semantic / syntax decisions aside - I think I've figured out the problem with this branch... When it merges the two syntax rules, its not remarking the 'marks' correctly. Upshot: the list of applicable methods for the syntax at a call-site is empty for the provided parameters - hence 'non-evaluatable'.

peter-b:
@runrevmark I think the `index` and `offset` distinctions are clear.  I'm not sure what the exact implications for the `string`/`char` syntax are though.

I suggest that I should implement `index` for lists, and then figure out from that what the semantics should be for strings...?

runrevmark:
@peter-b: Yes, so to clarify...

I think instead of 'the offset of list x in y' should be just:

```
  the offset of x in y
```

And what we currently have as 'the offset of element x in y' should be:

```
  the index of x in y
```

Additionally, I don't think the optional modifiers are necessary as the type of index/offset operation can be determined by the type of the target.

Adding string char and byte 'index' operations is perhaps a separate thing :)

peter-b:
Okay. What about removing "offset of chars" variants from strings?

runrevmark:
The 'offset of chars' variant isn't needed, and 'offset of bytes' should be just 'offset of' I think.

peter-b:
@runrevmark I think that should be sufficiently comprehensive.

peter-b:
Apparently not; I need to add `index` for `Data`.

livecodefraser:
Everyone happy with this now?

peter-b:
@livecodefraser I'm happy if @runrevmark is. :grinning:

runrevmark:
Reviewed-by: @runrevmark.

PR #1957, opened 2015-03-03
[Bug 14689] Cast the last nil sentinel to prevent EXC_BAD_ACCESS

livecodepanos:
The crash (because of EXC_BAD_ACCESS) happened only in release mode, and only when tested against 64 bit iOS devices (physical or simulated)


livecodesebastien:
Reviewed-By: sebastien.nouat@livecode.com

PR #1987, opened 2015-03-06
Fontawesome inclusion

livecodefraser:


livecodefraser:
Reviewed-By: Fraser Gordon

PR #1993, opened 2015-03-06
[[ Canvas ]] Add useful MCRange functions to libfoundation header

livecodeian:
[[ Canvas ]] Fix some relative svg commands not offsetting their point params
[[ Canvas ]] Parse command string as native chars throughout. svg command data should be entirely composed of ASCII chars
[[ Canvas ]] Fix real number parsing - use custom function to retrun real value directly rather than as MCNumber.


livecodeian:
Closing to split changes into multiple commits

PR #2016, opened 2015-03-10
Quick-fix to the Header widget so action setting works

livecodefraser:
- Don't use iOS 8 theme (Windows doesn't have the Helvetica Neue font)
- Different approach to parsing actions property


livecodeali:
These changes have been made slightly differently in the header widget.

PR #2025, opened 2015-03-10
[Bugfix 14846] Fix and re-enable offset/index before/after ops

peter-b:


livecodeali:
Reviewed.

peter-b:
Requires a release note.

PR #2029, opened 2015-03-11
[Bug 14846] Fix and re-enable offset/index before/after ops for Data

peter-b:


libscript/src/binary.mlc
```diff
@@ -203,7 +203,7 @@ Tags: Binary
 syntax EmptyData is expression
 	"the" "empty" "data"
 begin
-	MCBinaryEvalEmpty(output)
+	MCDataEvalEmpty(output)
 end syntax
```
livecodeali, line 207: It's probably not important, but wouldn't it be better to keep the prefix consistent and correct the name in the module-binary.cpp file to be MCBinaryEvalEmpty?

peter-b:
@runrevali I'm super undecided. In the interests of strictness my gut feeling is that either we should accept both `after 0` and `before 0` degenerate cars, or neither …

livecodeali:
Fair enough :-)

PR #2030, opened 2015-03-11
libfoundation: MCAssert(): Add a newline to message output.

peter-b:
Makes assertion crashes slightly more ergonomic on platforms where
libc prints a message on assertion failure.


livecodeali:
Reviewed.

PR #2073, opened 2015-03-16
[[ Bug 14937 ]] Use correct variable name for standalone settings

livecodeali:


livecodeali:
Fixed already here:
https://github.com/runrev/livecode/commit/07c8417a44160accd1552ba40646816d495dfdd6

PR #2084, opened 2015-03-16
[[ Bug 14897 ]] $0 should return the self positional parameter on server

livecodesebastien:


livecodestephen:
This is identical to #1405

livecodesebastien:
As stated by @livecodestephen , that's identical to #1405

PR #2085, opened 2015-03-17
Added iOS 8 and future proofed to iOS 9... not sure how to redesign this...

montegoulding:
... so it doesn't need updating


livecodesebastien:
Hi Monte, thanks for the update. No problem to get that merged in, and that will be back ported to develop-6.7 and develop-7.0.

PR #2093, opened 2015-03-18
[[ Release Notes ]] Remove reference to the Font Museo Sans Bold, which ...

livecodesebastien:
...appears to be missing from the server.


livecodesebastien:
The real need is simply to follow the website change, and the new font they are using.
Closing this pull request.

PR #2100, opened 2015-03-19
Feature ide widgets

runrevben:
Problem with canvas drawing text left aligned


livecodeali:
This pull request consists of all the changes to extensions for the current develop IDE.
- All widgets using icons now use the iconsvg library, and icons are referenced by name in the corresponding places in script
- Much of the documentation has been updated
- New widget 'Tree View', which is used in the property inspector for array views (including custom properties) and in the script editor variable viewer window
- New widget 'Palette Actions' which is used for the headers and footers of all IDE palettes.
- WIP widgets navbar and icon picker, the former using the latter as a popup widget.

livecodeali:
There's too much here to practically review, however we have been using this branch consistently for the past couple of weeks, so it is well tested.

PR #2110, opened 2015-03-23
[[ Bug 15045 ]] Clamp range correctly when calculating nested char chunk extents

livecodeali:


livecodesebastien:
Could you add the release notes? That looks fine otherwise.

PR #2119, opened 2015-03-25
[[ Bug 14326 ]] Ensure that all the symbols from external.c are passed t...

livecodesebastien:
...o iOS external (including setExternalInterfaceVersion)


livecodesebastien:
No bug notes, as the bug is only visible in 7.0.
See the bug report http://quality.runrev.com/show_bug.cgi?id=14326

PR #2120, opened 2015-03-25
[[ Bug 14326 ]] Ensure that mobiles rev externals can set the external interface version

livecodesebastien:


livecodesebastien:
7.0 counterpart of #2119 (with the bugfix note)

PR #2130, opened 2015-03-27
[[ Bug 14287 ]] Use correct length when creating port name

livecodeali:


livecodesebastien:
Reviewed-by: Sebastien Nouat

That will wait for the 7.0.5 cycle to be merged in though.

PR #2147, opened 2015-04-01
Tabs rects are now calculated so that the boundary fall on 0.5 pixel bou...

runrevben:


extensions/widgets/paletteactions/paletteactions.lcb
```diff
@@ -281,13 +281,16 @@ private handler drawTab(in pIndex as Integer)
 	
 	if pIndex is mSelectedNavItem then
 		variable tPath as Path
-        put getPathOfTab(tRect) into tPath
-		set the paint of this canvas to getPaint("line")
-		stroke tPath on this canvas
-		
+		put getPathOfTab(tRect) into tPath
+		log tPath
```
livecodeali: If we do 
variable tRectPath as Path
put tPath into tRectPath

then we don't have to call getPathOfTab again.

PR #2163, opened 2015-04-09
Gypify the build process for 6.7

livecodefraser:


peter-b:
Shouldn't the gyp source code live in `thirdparty`?

PR #2184, opened 2015-04-20
[[ Merge 6.7.5 rc 1 ]] Merge 6.7.5 RC 1 in develop-7.0

livecodesebastien:


engine/src/coretextfonts.cpp
```diff
@@ -32,6 +32,48 @@
...
+
+    uindex_t t_separator;
+    if (!MCStringFirstIndexOfChar(*t_mapping, '=', 0, kMCStringOptionCompareExact, t_separator))
+        return;
+    
+    MCAutoStringRef t_from, t_to;
+    MCNewAutoNameRef t_from_as_name;
+    if (!MCStringCopySubstring(*t_mapping, MCRangeMake(0, t_separator), &t_from)
+            || !MCNameCreate(*t_from, &t_from_as_name)
+            || !MCStringCopySubstring(*t_mapping, MCRangeMake(t_separator, UINDEX_MAX), &t_to))
```
livecodeali: I think this includes the '=' in the t_to string doesn't it? I presume that is not wanted.

engine/src/exec-files.cpp
```diff
@@ -254,9 +255,18 @@ void MCFilesEvalSpecialFolderPath(MCExecContext& ctxt, MCStringRef p_folder, MCS
...
     MCNewAutoNameRef t_path;
+    t_error = false;
     MCNameCreate(p_folder, &t_path);
-	if (MCS_getspecialfolder(*t_path, r_path))
+    // We have a special, mode-specific resource folder
+    if (MCStringIsEqualToCString(p_folder, "resources", kMCStringOptionCompareCaseless))
+        MCModeGetResourcesFolder(r_path);
+    if (!MCS_getspecialfolder(*t_path, r_path))
+        t_error = true;
+
```
livecodeali, line 267: Doesn't this just overwrite the returned r_path from MCModeGetResourcesFolder with the one returned from MCS_getspecialfolder?

PR #2185, opened 2015-04-20
Bugfix 15241

livecodeian:


engine/src/exec-pasteboard.cpp
```diff
@@ -639,28 +639,80 @@ void MCPasteboardSetClipboardOrDragData(MCExecContext& ctxt, MCNameRef p_index,
...
+		{
+			case TRANSFER_TYPE_TEXT:
+			case TRANSFER_TYPE_FILES:
+			{
+				// convert to MCStringRef
+				MCStringRef t_string;
+				t_success = ctxt.ConvertToString(p_data, t_string);
+				if (t_success)
+					t_value = t_string;
+
```
livecodeali, line 662: Assignment to an Auto class via = increments the reference count, so I think t_string here (and the other ones below) should be MCAutoStringRefs / MCAutoDataRefs.

livecodeali:
Reviewed.

PR #2189, opened 2015-04-21
[[ CI ]] Call "apt-get update" to refresh package list before attempting...

livecodeian:
... install


peter-b:
@runrevian Should work just as well to put those two commands on separate lines without chaining them...

livecodeian:
@peter-b Sure, if we need to add any more commands there then you can split them up.

peter-b:
Reviewed-By: @peter-b

PR #2195, opened 2015-04-22
[[ Bug 15243 ]] Do not use the uninitialised return value of GetParagrag...

livecodesebastien:
...PropOfCharChunk if the result is mixed


livecodeali:
Reviewed.

PR #2213, opened 2015-04-24
[[ TemplateAppMetadata ]] Regression fix to TemplateManifest for android...

montegoulding:
... manifests and implementation of template plists for iOS

It looks like I got mixed up at some point between the tManifestFile variable and the tManifest one... sorry. Should be fixed now and also implemented for iOS if you have your own Info.plist in copy files it will use that instead of the default plist.


montegoulding:
Closing due to unintended files in the commit and some whitespace issues

PR #2216, opened 2015-04-24
Fix compiler warning in external.h

livecodestephen:
Compiling with gcc and `-Wpedantic` you get a warning about an extra ';'


livecodestephen:
Closing as I've noticed a couple of other related bugs.

PR #2237, opened 2015-04-30
Improve the errors when the shell command fails

livecodestephen:
When a `shell` command fails there are several system calls that could have failed and there's no indication in the error message which one actually failed, so I've added the system error to the error stack.

(There was also a leak of the `toparent[1]` file handle in the case of an error, which I've fixed.)


peter-b:
@livecode-vulcan review ok 55a3f95

livecode-vulcan:
💙 review by @peter-b ok 55a3f9549262becbd8c7e82202e5d0097bc7d5c3

PR #2280, opened 2015-05-15
[Bug 15372] wait with messages cause crash on Mac OS

livecodepanos:


engine/src/desktop.cpp
```diff
@@ -867,8 +867,12 @@ void MCPlatformHandleTextInputInsertText(MCPlatformWindowRef p_window, unichar_t
...
                     
                     // SN-2014-11-03: [[ Bug 13832 ]] Enqueue the event, instead of firing it now (we are still in the NSApplication's keyDown).
-                    MCKeyMessageAppend(s_pending_key_up, s_pending_key_down -> key_code, s_pending_key_down -> mapped_codepoint, s_pending_key_down -> unmapped_codepoint);
-                    MCKeyMessageNext(s_pending_key_down);
+                    // PM-2015-05-15: [[ Bug 15372]] s_pending_key_down might become nil in the meanwhile, if 'wait with messages' is used
+                    if (s_pending_key_down != nil)
+                    {
+                        MCKeyMessageAppend(s_pending_key_up, s_pending_key_down -> key_code, s_pending_key_down -> mapped_codepoint, s_pending_key_down -> unmapped_codepoint);
+                        MCKeyMessageNext(s_pending_key_down);
+                    }
```
livecodesebastien: That's not the right fix, as we want the KeyUp message to be fired. The call to MCKeyMessageAppend should simply be made before calling wkdown (MCKeyMessageNext can be left where it is, it does nothing if the message queue is NULL).

livecodesebastien:
@livecode-vulcan review ok f478b2f

livecode-vulcan:
💙 review by @livecodesebastien ok f478b2f1e5e5aa23198c55d50c2ef4a996fc247f

PR #2292, opened 2015-05-19
[[ Bug 15208 ]] Fix for the crash caused when using the character vie…

livecodesebastien:
…wer in a field with selected text

### DIFF
--- a/docs/notes/bugfix-15208.md
+++ b/docs/notes/bugfix-15208.md
@@ -1,0 +1,1 @@
+# LiveCode crashes when using System Character Viewer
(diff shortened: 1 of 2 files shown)


livecodesebastien:
This pull request only fixes the crash, but using the Character Viewer (Ctrl + Cmd + Space) is still useless if no text is selected in the field when opened.

livecodesebastien:
The issue is not the fact that MCactivefield becomes nil; it is that the focus is not kept when the utility window that the Character Viewer is, is open.

PR #2361, opened 2015-06-04
[[ Merge 6.7.6 rc 1 ]]

livecodesebastien:
The only thing refactored is the new dontuseqt player property.


livecodeali:
@livecode-vulcan review ok 266c212

livecode-vulcan:
💙 review by @livecodeali ok 266c2129670358fb2fd16ecaa265decc1b89c280

engine/src/player-platform.cpp
```diff
@@ -879,6 +879,9 @@ MCPlayer::MCPlayer()
     m_is_attached = false;
     m_should_attach = false;
     m_should_recreate = false;
+    
+    dontuseqt = false;
+    MCdontuseQT = false;
 }
```
livecodepanos, line 885: This should be dontuseqt = False and usingqt = False :)

livecodesebastien, line 885: Good catch.

livecodeali, line 885: Oops. Yes, thanks @runrevpanos.

PR #2420, opened 2015-06-16
Make bootstrap

mwieder:
modified the Makefile to include bootstrap when making all, because I can never remember to do this on my own.


peter-b:
Hi @mwieder,

Thank you for your patch.  @livecodefraser has been working very hard over the last couple of months to modernise (replace) the entire build system used on all platforms, which will address your problem.  I'm therefore not going to merge your changes, because in the next couple of days they won't be relevant any more!  If you want to try out the new build system, try out the `develop-gyp` branch.

When you're submitting a change like this one in future, could you please make it just a single commit to be added to our repository?  If I was going to merge this, I would have to clean it up to remove all of the unnecessary merge commits.

Many thanks again,

Peter

PR #2434, opened 2015-06-18
Bugfix linux dnd modal loop

livecodefraser:
This PR looks bigger than it is because it depends on bugfix-15482


livecodefraser:
Already merged as part of another branch

PR #2467, opened 2015-06-23
[CI] Disable Travis builds.

peter-b:
There are some problems with gyp and Travis CI that are making
builds on Travis reliably fail.

Disable Travis CI builds until they can be resolved.


livecodesebastien:
@livecode-vulcan review ok ee15cf3

PR #2502, opened 2015-06-29
[Bug 15552] Send textChanged any time the contents of the fld are updated

livecodepanos:
regardless of the focus. This allows textChanged msg to be sent in cases like:
1. set the text of fld 1 to "foo"
2. put "foo" into fld 1


livecodepanos:
Closing this PR as it can cause performance problems in the IDE. See comment 3 here:

http://quality.runrev.com/show_bug.cgi?id=15552

PR #2548, opened 2015-07-08
[Bug 15253] Setting property to function in message box causes error

livecodepanos:


livecodepanos:
Closing this PR after discussing with @runrevmark. This patch is not correct since it will cause all functions to go to this card of the default stack as soon as they've failed to go through the message path.
The bug is in the message box, not the engine

PR #2633, opened 2015-07-22
Add gyp file for server only builds

livecodestephen:
This adds the ability to just build server.  A server only build is performed by running `config.sh` with the BUILD_EDITION set to server, e.g.

```
    BUILD_EDITION=server ./config.sh
    make -C build-linux-x86_64/livecode 
```

(Running `make` on the root directory doesn't work as this re-runs `config.sh`.)

There is also an install rule that installs LiveCode server in `/opt/livecode/server`.

```
    sudo make -C build-linux-x86_64/livecode install
```

I haven't tested the install rule quite as much as I'd like -- I don't want to overwrite the current install on our server just now -- and I was having some problems where the install target was causing other files to be remade, but I think that this is problem with my computer and not with the build system.

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -347,6 +347,10 @@ basic_args="${format_args} --depth ${DEPTH} --generator-output ${GENERATOR_OUTPU
 
 if [ "${BUILD_EDITION}" == "commercial" ] ; then
   basic_args="${basic_args} ../livecode-commercial.gyp"
+elif [ "${BUILD_EDITION}" == "server" ] ; then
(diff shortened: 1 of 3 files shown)


livecodestephen:
I should add that I've only tested this on linux, although I suspect the build rules will work on other platforms.  The install target will only work on linux - perhaps I should have put this within a 'conditions' section.

livecode-vulcan:
📝 Hi @livecodestephen, I haven't been able to verify that you've signed our [LiveCode Contributor's Agreement](https://livecode.com/account/developer/contribute).

PR #2648, opened 2015-07-23
[[ Clone and Run ]] Ensure iOS built engine path matches gyp-generated one

livecodeali:


livecodeali:
Already implemented here: https://github.com/runrev/livecode/pull/2580

PR #2681, opened 2015-07-31
MCCustomFileHandle::Read(): Correct error check logic.

peter-b:


peter-b:
This may need to be applied to 7.0, so don't merge this quite yet ;-)

peter-b:
Confirmed that this needs to be fixed in develop-7.0.

PR #2702, opened 2015-08-07
[Bug 15692] Fixed malformed xml of 'flip' command and added some more info

livecodepanos:
about synonyms of vertical and horizontal


livecodepanos:
Github seems to be confused and it does not show the diff correctly.

livecodeali:
I think you have changed the line endings to carriage returns when they should be line feeds or perhaps CRLFs

livecodepanos:
Hmm, ok, maybe this happened because I edited the xml using a version of the docs_editor stack. I will close this PR and submit a new one, where I will edit directly the xml using xcode

PR #2727, opened 2015-08-19
Revert "[[ Bug 15743 ]] binzip-copy target cannot copy dSYM files in …

livecodesebastien:
Reverts runrev/livecode#2724

The pull request breaks commercial compilation, and was not fixing the issue in the best way.

### DIFF
--- a/config/debug_syms.gypi
+++ b/config/debug_syms.gypi
@@ -1,122 +1,65 @@
 {
-	'targets':
+	'conditions':
 	[
-		{
-			'target_name': 'debug-symbols',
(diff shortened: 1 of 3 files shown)


peter-b:
@livecodesebastien: can you please update the PR, putting the reason for the revert into the actual commit message?  If you just put it into the PR's comments, then it won't make its way into the git history! Also, should probably run this past @livecodefraser

livecodesebastien:
Superseded by https://github.com/runrev/livecode/pull/2728

PR #2746, opened 2015-08-22
libfoundation: MCListAppend(): Fall back to copy description

peter-b:
When printing error messages from some LiveCode Builder programs it
was possible to hit the default path in `MCListAppend()`'s type switch
statement, causing an immediate (and uninformative) abort.

This patch makes `MCListAppend()` use `MCValueCopyDescription()` to
format values that it can't handle directly.  There were a couple of
alternatives to this approach:
- Use `MCStringFormat()`
- Ignore them silently


peter-b:
Changing this to use `MCStringFormat()`, per conversation with @livecodefraser.

PR #2755, opened 2015-08-24
[[ Emscripten ]] Re-name the emscripten standalone engine to include the version

livecodefraser:


peter-b:
@livecodefraser I think you need to update the `getHtmlTemplateFile()` handler in the standalone builder stack, too. Sorry, I forgot about that before.

PR #2765, opened 2015-08-26
[emscripten] Update top-level build instructions.

peter-b:
- prebuilt libicu is now available on the server
- no longer necessary to manually create a root filesystem image
- built products are now put in the right place
- can now generate HTML5 standalones


peter-b:
@livecodefraser Updated. Thanks for your comments.

PR #2934, opened 2015-09-28
bugfix-16032.md

BerndN:
catch OnMouseCancel in gradientrampeditor

### DIFF
--- a/docs/notes/bugfix-16032.md
+++ b/docs/notes/bugfix-16032.md
@@ -1,0 +1,1 @@
+[[ 16032 ]] catch OnMouseCancel in gradientrampeditor


peter-b:
Hi @BerndN, could you please make this change as part of your other pull request?

BerndN:
Hi @peter-b,  I hope I managed to get it right.

peter-b:
@BerndN: You need to add the bugfix file in your `patch-1` branch, so that it appears as part of #2930.

livecodeali:
@BerndN In theory the instructions here should help!
https://github.com/livecode/livecode/blob/community-docs/docs/CONTRIBUTING.md#adding-a-bug-fix-release-note

BerndN:
@livecodeali, I tried to push from patch-1 branch, don't see if it has worked, sorry, gitHub is not where I feel at home. Thanks for the guidance.

PR #2970, opened 2015-10-02
[[ Bug 15929 ]] Speed up 'read' command

livecodesebastien:
String Fast Appending

Fast appending a string consists of making Unicode prevail over nativeness
(which is the opposite of the usual behaviour).

The fast Append functions are mainly intended for tigh loops where small
chunks of text are appended, and where we don't need nor want to check the
nativeness of the string after each appending.

That is the case in MCFilesExecPerformReadCodeUnit, where the encoding is
fixed for the whole reading, and that we don't want to spend a huge amount
of time checking whether the string is native, as it is not the most likely.

Additionally, creating a string from UTF-8 bytes now checks whether the string
is not a native string (all chars < 128), and if it is, uses the bytes as native
chars.

### DIFF
--- a/docs/notes/bugfix-15929.md
+++ b/docs/notes/bugfix-15929.md
@@ -1,0 +1,1 @@
+# reading from Unicode files is extremely slow in LC 7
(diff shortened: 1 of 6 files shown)


livecodesebastien:
Breaks the reading (the number of chars to read is not respected) on the top of not fixing the issue.
Superseded by https://github.com/livecode/livecode/pull/2974

PR #2980, opened 2015-10-05
[[ Bugfix 15878 ]] Fix a crash caused by waiting during modal session setup

livecodefraser:
Creating a modal window session on Cocoa can trigger event processing
(e.g. because another window lost focus). Waiting during these
events caused a crash because the wait loop attempted to run the
modal session that hadn't actually been created yet.


livecodesebastien:
@livecodefraser, I'm afraid that this is a 6.7-introduced bug... Can you rebase your fix?

livecodefraser:
Hmm... according to @trevordevore, the crash only happens in 7. It might be a combination of changes that brought about these exact circumstances.

livecodefraser:
Moved to 6.7: https://github.com/livecode/livecode/pull/2981

trevordevore:
@livecodefraser FWIW I have been shipping my app with the 6.7 engine for over a year and I've never seen the crash.

PR #3042, opened 2015-10-09
[Bug 157] is a color should return false for integers < 0 and > 255

livecodepanos:


engine/src/uidc.cpp
```diff
@@ -1690,22 +1695,29 @@ Boolean MCUIDC::parsecolor(const MCString &s, MCColor *color, char **cname)
...
 		// MDW-2013-06-12: [[ Bug 10950 ]] non-numeric second argument present
 		if (l != 0)
 			return False;
 		// only a single integer as the color specification,
-		// restrict it to 0-255 and get values from sccolors array
-		i1 = MCU_max(1, MCU_min(i1, 256)) - 1;
-		i3 = sccolors[i1].blue;
-		i2 = sccolors[i1].green;
-		i1 = sccolors[i1].red;
+		// get values from sccolors array
```
livecodepanos, line 1708: This fixes another bug:
answer color with 255 -->
Expected result: Color picker opens and white color is selected
Observed result: Color picker opens and black color is selected

The opposite happens if executing: answer color with 0

livecodeali:
@livecodepanos slightly annoyingly, there are actually 257 predefined colours of this kind (0 is white and 256 is black)

livecodeali:
@livecodepanos My mistake, there are indeed 256, but color 0 = color 1 previously (i.e. they are 1-indexed)

livecodepanos:
Closing this PR as discussed offline with @runrevmark, since we have to maintain backwards compatibility, for example `set the backcolor to 500` should continue to work. I'll mark the bug report as an anomaly, so that we can revisit it in the future.

PR #3072, opened 2015-10-19
[[ Bug 16232 ]] Do not overwrite existing files when copying at stand…

livecodesebastien:
…alone saving

Files such as externals can be copied before the Copy Files content, in which case they should not be overwritten


livecodesebastien:
This regression should be fixed against `release-6.7.8`

PR #3139, opened 2015-10-28
[[ Tests ]] Allow SetupTests handler to be run before tests in order to load libraries

livecodeali:


livecodeali:
Hooray!

tests/_testlib.livecodescript
```diff
@@ -117,3 +117,13 @@ on ErrorDialog executionError, parseError
    write executionError & return to stderr
    quit 1
 end ErrorDialog
+
+function EngineRepositoryPath
+  set the itemdelimiter to "/"
+  return item 1 to -3 of the filename of me
+end EngineRepositoryPath
```
peter-b: Can this be called `TestGetEngineRepositoryPath` or something? Just to avoid any potential conflict with non-test libraries.

peter-b:
Merged as #3150.

PR #3149, opened 2015-10-30
[Tests] Better description of Google Test documentation.

peter-b:
I failed to find the documentation for writing C++ tests using
Google Test on the first attempt, but it turns out that it actually
does exist and is fairly good.


peter-b:
Well, I didn't mean to merge that unreviewed. :fearful:

PR #3197, opened 2015-11-05
MCObject::load() shouldn't be called for MCStack instances

peter-b:
When loading a stack, the type byte needs to be passed to
`MCStack::load()` in order to allow the correct stack loading mode to
be selected.  All other objects can be fully loaded without the type
byte.

Previously, only a 3-argument variant of `load()` was defined in
`MCStack`, which meant that it was possible to (incorrectly) call the
base class (2-argument) implementation of `MCObject::load()` by
upcasting an `MCStack` pointer.  This was flagged up by the emscripten
C++ compiler with a "overloaded-virtual" warning.

This patch explicitly overrides `MCObject::load()` in `MCStack` with
an implementation that indicates a programming error.

### DIFF
--- a/engine/src/stack.h
+++ b/engine/src/stack.h
@@ -743,6 +743,7 @@ public:
 	void getstackfile(MCStringRef p_name, MCStringRef &r_name);
 	void setfilename(MCStringRef f);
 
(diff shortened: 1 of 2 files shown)


peter-b:
Merging this into another pull request.

PR #3209, opened 2015-11-06
Replace references to quality.runrev.com

peter-b:
BugZilla is now at http://quality.livecode.com/.

**Note:** This PR deliberately doesn't modify any release notes in order to avoid confusing the release notes builder.


livecodesebastien:
Would you mind doing it against `develop-6.7`, so that we have something up-to-date there as well?
Also, removing all the empty-line whitespaces in `livecodescript`s is not ideal

PR #3248, opened 2015-11-11
[[ Builder ]] Add docs data to sqlite database on installer build

livecodeali:
There is no easy way of creating a docs api in the IDE using the
JSON data, so we include an sqlite database with each entry's docs
data array arrayEncoded for easy retrieval.

The database includes a table of libraries (currently livecode script
and livecode builder) and a table for each library containing its
docs data.


peter-b:
Just to check: does the database also get built by the IDE when running from the git working directory?

livecodeali:
It will do, I havent added the IDE portion of this yet though.

What should probably happen is that the builder and IDE should use exactly the same code to do these things, but it's slightly awkward at the moment. 

I'll put on my list of things to do to refactor this (and the JSON building).

peter-b:
@livecodeali Is this useful to merge in as it is?  Or is this still WIP?

livecodeali:
I should refactor it a bit first probably

PR #3264, opened 2015-11-13
[Bug 16407] libfoundation: Change MCStringSplit by empty

peter-b:
`MCStringSplit()` misbehaves when it receives an empty delimiter
string, entering an infinite loop.

This patch alters the behaviour of `MCStringSplit()` to treat empty
delimiters as never existing in the target string.  This is consistent
with the behaviour of the `contains` operator:
`"abc" contains empty -> false`

### DIFF
--- a/docs/notes/bugfix-16407.md
+++ b/docs/notes/bugfix-16407.md
@@ -1,0 +1,1 @@
+# Splitting by empty causes a hang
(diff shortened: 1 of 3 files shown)


peter-b:
Argh, wrong target branch.

PR #3268, opened 2015-11-14
Ignore .lci dir

montegoulding:
The linter for LCB in the LiveCode package for Atom outputs *.lci files
into a .lci directory.


peter-b:
Wouldn't it be better to make the linter use a directory in /tmp?

montegoulding:
It could be...happy to change the linter if you think so. Every time tmp is trashed there will be potential dependency errors though.

montegoulding:
I'm guessing this will never make it in so closing.

PR #3351, opened 2015-11-27
[Bug 15970] Change mechanism for saving in specific stack file formats

peter-b:
Only the LiveCode "8.0" stack format supports widgets.  However, it is otherwise identical to the "7.0" format.  By default, the LiveCode 8 `save` command uses the "8.0" format for stacks that contain widgets, and the "7.0" format otherwise.

[Bug 15970](http://quality.livecode.com/show_bug.cgi?id=15970) related to the behaviour of the `save` command when `the stackFileVersion` was set to something other than "8.0".  The IDE's "Save As..." dialog offers the option of which stack file format to use.  When saving a stack that uses widgets, and selecting the "7.0" format, the engine would use the "8.0" format, regardless -- and this caused confusing errors when attempting to load the resulting file into the LiveCode 7 IDE.

This pull request changes the mechanism for choosing the file format version used by the `save` command. It:
- Refactors all of the object save logic to pass the version to use as a parameter, rather than using the `MCstackfileversion` global static variable.
- Adds a new optional `with format _` clause to the `save` command, which can be used to specify the file format version.
- Adds a new optional `with newest format` clause, which causes `save` to always use the latest & greatest stack file format
- Marks `the stackFileVersion` as deprecated, and recommends specifying the file format to use directly when calling `save`.

The behaviour of all existing stacks that make use of the `save` command or `the stackFileVersion` are unaffected by this change.

### DIFF
--- a/docs/dictionary/command/save.lcdoc
+++ b/docs/dictionary/command/save.lcdoc
@@ -2,20 +2,20 @@ Name: save
 
 Type: command
 
-Syntax: save <stack> [as <filePath>]
+Syntax: save <stack> [as <filePath>] [with format <stackFormat> | with newest format]
 
 Summary: Saves a <stack file> on the user's system.
 
 Introduced: 1.0
 
-OS: mac,windows,linux,ios,android
+OS: mac,windows,linux,ios,android,html5
 
 Platforms: desktop,web,mobile
 
 Security: disk
 
 Example:
-save stack "project1"
+save stack "project1" with newest format
 
 Example:
 save this stack as "Backup"
@@ -23,21 +23,35 @@ save this stack as "Backup"
 Example:
 save stack "project1" as "/Disk/Folder/File"
 
+Example:
+save stack "MyOldApp" with format 7.0
+
 Parameters:
 stack: Any open stack.
-filePath: The filePath specifies the name and location of the file you want to save to. If you specify a name but not a location, LiveCode assumes the file is in the defaultFolder. If the file does not exist, LiveCode creates it.
 
-The result: If the stack cannot be saved (for example, if you try to save it to a nonexistent drive or to a CD-ROM), the result <function> is set to "Can't open stack file".
+filePath: The <filePath> specifies the name and location of the file you want to save to. If you specify a name but not a location, LiveCode assumes the file is in the <defaultFolder>. If the file does not exist, LiveCode creates it.
+
+stackFormat: The <stack version|stack file format version> to use when saving.
+
+The result: If the stack cannot be saved (for example, if you try to save it to a nonexistent drive or to a CD-ROM), the <result> will contain an error message.  Otherwise, the <result> will be empty.
 
 Description:
 Use the <save> <command> to save changes to a <stack> or to save a copy of a <stack> into another <file>.
 
 The <save> <command> saves all <stacks> stored in the same <file> as the specified <stack>. That is, if you save a <main stack>, all <substacks> of that stack are also saved in the same <file>, and if you save a <substack>, its <main stack> and any other <substacks> are also saved.
 
-If the stack has not yet been saved and doesn't have a filename, you must use theform.
+If the stack has not yet been saved and doesn't have a filename, you must use the `as` form.
 
 You cannot save to a standalone application's file; <standalone application|standalones> are read-only.
 
-References: stackFileVersion (property), substacks (property), stack (object), as (keyword), file (keyword), substack (glossary), stack file (glossary), command (glossary), standalone application (glossary), main stack (glossary), stacks (function), create stack (command), function (control_st)
+The <save> command can save the <stack> in several different versions of the LiveCode stack file format.
+
+* If you use the `with newest format` form of the <save> command, the <stack> is saved using the newest <stack version>
+* If you use the `with format` form of the <save> command, the <stackFormat> should be the particular <stack version> you want the stack to be saved as.
+* If you do not specify a format to use, the current value of the <stackFileFormat> property is used.
+
+**Warning:** If you save in an older <stack version> that doesn't support all of the features used in the stack, then some of contents of the stack may be lost.  In general, you are recommended to use the `with newest format` form of the <save> command.
+
+References: substacks (property), stack (object), as (keyword), file (keyword), substack (glossary), stack file (glossary), command (glossary), standalone application (glossary), main stack (glossary), stacks (function), create stack (command), function (control_st), stack version (glossary), defaultFolder (property), stackFileFormat (property)
 
 Tags: file system
--- a/docs/dictionary/object/stack.lcdoc
+++ b/docs/dictionary/object/stack.lcdoc
@@ -31,6 +31,6 @@ Stacks contain one or more cards, and may optionally contain backgrounds.
 
 The stack object has a number of properties and messages associated with it. To see a list of messages that can be sent to a stack as a result of user actions or internal LiveCode events, open the "LiveCode Language Dictionary" page of the main Documentation window, ensure that the stack column is visible and use sort and filter to bring the relevant entries to the top
 
-References: templateStack (keyword), mode (property), stackFileVersion (property), defaultStack (property), style (property), topLevel (command), go (command), modal (command), palette (command), modeless (command), revLoadedStacks (function), mouseStack (function), stack (object), object (object), property (glossary), dialog box (glossary), command (glossary), object type (glossary)
+References: templateStack (keyword), mode (property), stack version (glossary), defaultStack (property), style (property), topLevel (command), go (command), modal (command), palette (command), modeless (command), revLoadedStacks (function), mouseStack (function), stack (object), object (object), property (glossary), dialog box (glossary), command (glossary), object type (glossary)
 
 Tags: objects
--- a/docs/dictionary/property/stackFileVersion.lcdoc
+++ b/docs/dictionary/property/stackFileVersion.lcdoc
@@ -6,74 +6,29 @@ Syntax: set the stackFileVersion to <version>
 
 Syntax: the stackFileVersion
 
-Summary: Reports the version of the stack file format to use.
+Summary: The version of the stack file format to use when saving stacks
 
 Introduced: 2.7
 
-OS: mac,windows,linux,ios,android
+OS: mac,windows,linux,ios,android,html5
 
 Platforms: desktop,server,web,mobile
 
-Example:
-command legacySave pStack
-  local
-tOldStackFileVersion
-  put the stackFileVersion into
-tOldStackFileVersion
-  set the stackFileVersion to 2.4
-  save
-stack pStack
-  set the stackFileVersion to
-OldStackFileVersion
-end legacySave
-
 Value:
-The <stackFileVersion> currently has two possible values:
-  - "2.4" - This stack file version is used by LiveCode 2.6.1 and arlier
-  - "2.7" - This stack file version is used by versions of LiveCode after 2.6.1 and before 5.5.0.
-  - "5.5" - This stack file version is used by versions of LiveCode after 5.5.0 and before .0.
-  - "7.0" - This stack file version is used by versions of LiveCode 7.0. and after.
+The name of a <stack version>.
 
 Description:
-Use the <stackFileVersion> to obtain or change which version of the stack file format LiveCode will use when saving stacks.
+**Note**: The <stackFileVersion> is deprecated.  Instead, please specify the <stack version> directly, as an optional parameter of the <save> <command>.
 
-The <stackFileVersion> is useful when stacks need to be saved in legacy formats for backwards compatibility but it should be used with caution.
+Use the <stackFileVersion> to obtain or change the default stack file format version LiveCode uses when saving stacks.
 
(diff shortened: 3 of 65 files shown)


peter-b:
This probably needs tests, but I'm really not sure how to test it.

mwieder:
1. Testing is indeed a problem. We might need something like mocks. Possibly saving into memory or into a docker image.
2. Ignoring the stackfileversion will break backwards compatibility. I set the stackfileversion to 2.4 in order to ensure that stacks that don't rely on newer functionality will still work with older builds. I can add the "save with format" syntax everywhere I use the stackfileversion, but I'd have to a) test the version each time and b) use the "do" command since the new syntax won't compile properly for earlier versions.

Is there a downside to checking the MCstackfileversion global (no matter what its name) before setting the default value to of p_version 7000 in dispatch.cpp? I think that would maintain backward compatibility while allowing the new save_as syntax, which I like.

runrevmark:
It should be easy enough... You can tell what format a stackfile is in by checking the first few bytes. Using that you can tell after a save command that it has been saved in the expected format.

montegoulding:
@mwieder I think it is worth breaking a bit of code here. I wonder how much work has been lost over the years as a result of some code that sets the stackFileVersion and fails to set it back for whatever reason (poor logic, poor error handling etc)... From the IDE's perspective it might be nice if there were a dry run option to give it a chance to throw up an error dialog and a way to allow the save even if data is lost. Another nice feature would be to have an option to save as the minimum version that supports the properties of the stack and if that is feasible it could be a good default behavior.

docs/dictionary/command/save.lcdoc
```diff
@@ -23,21 +23,29 @@ save this stack as "Backup"
...
 
 Description:
 Use the <save> <command> to save changes to a <stack> or to save a copy of a <stack> into another <file>.
 
 The <save> <command> saves all <stacks> stored in the same <file> as the specified <stack>. That is, if you save a <main stack>, all <substacks> of that stack are also saved in the same <file>, and if you save a <substack>, its <main stack> and any other <substacks> are also saved.
 
-If the stack has not yet been saved and doesn't have a filename, you must use theform.
+If the stack has not yet been saved and doesn't have a filename, you must use the <as> form.
+
+If you specify a <stack version> to be used when saving, but the <stack> cannot be saved in that format without losing data, <save> will fail.
```
montegoulding: Am I wrong or is this only the case for widgets? If so perhaps that should be clarified.

docs/dictionary/command/save.lcdoc
```diff
@@ -23,21 +23,29 @@ save this stack as "Backup"
 Example:
 save stack "project1" as "/Disk/Folder/File"
 
+Example:
+save stack "MyOldApp" with format 7.0
```
mwieder, line 27: I think the valid formats should be listed here (7.0, 5.5, 2.7, 2.4)
It's not clear to me from the code what happens if an invalid format (e.g. 6.0) is requested.

mwieder:
runrevmark- yes, that's easy enough to test that way, but placing it into the automated test suite means saving multiple copies of the stack or having to deal with the operating system file overwrite dialog each time.

mwieder:
Monte- has the stackfileversion not being reset actually been a problem? I see this as breaking code in the name of fixing something that isn't really broken. This also has an impact on the feature of the IDE that preserves stack file versions. If the stackfileversion global really isn't going to be checked by the engine any more, then it should be more than deprecated: the compiler should raise an error, because the code that attempts to set/use it won't do anything useful.

montegoulding:
@mwieder You can't see the potential for data loss? It's not the actual bug this is trying to resolve though. I think the actual bug could be resolved by checking the stackFileVersion before saving widget data. It sounds like it is just presuming version 8 if there is widget data. Saving with a stackFileVersion less than the engine should reasonably result in data loss but like I said it would be nice if the ide were able to do a dry run to check if data will be lost and throw up a warning.

montegoulding:
Oh and maybe stackFileVersion could be changed to a local property? That resolves my data loss concerns

mwieder:
Monte- sure - I can see the _potential_. I'm just wondering if it's a problem that's actually come up or if this is just speculation about something that someone might fall into some day.

And regarding checking the stackfileversion, my concern is that the engine now is going to ignore it, even it you check it. The only options are saving in LC7 or LC8 formats unless you use the new "with format x" syntax, and that syntax is only going to be supported in LC8. So it requires version checking before saving if you want to save in pre-LC7 formats, and using the "do" syntax because the code won't compile properly in pre-LC8 versions otherwise.

Does what I wrote make sense? Feels like maybe I'm going around in circles. What I'm suggesting is that before the final "save the stack in 7.0 format" command the stackfileversion be checked to see if it's been changed from the default. That's it - a one-line code addition, I think.

montegoulding:
It's just speculation but I'm imagining any number of unreproducible bug reports involving data loss...

I agree with you with regard to checking the stackFileVersion first. It seems to me that in the absence of with format which should override the stackFileVersion that the default setting should be the stackFileVersion. Still, it would be nice if it were a local property to alleviate my other concerns but that has nothing to do with the reported bug.

mwieder:
OK - I'm with you on making the stackfileversion a local property ala itemdelimiter. I don't see any real drawbacks to that unless someone is making some very wide-ranging assumptions. But the complete loss of stackfileversion in the format algorithm and its deprecation to the point of unusability sets my nerves on edge.

montegoulding:
It's worthwhile pointing out though that stackFileVersion isn't the only global property that I have a problem with. The most dangerous is libURLSetSSLVerification which is not only global but it's not possible to limit the setting to particular domains.

peter-b:
@mwieder Thanks for the feedback, and good point about code that's intended to run on both 7.x and 8.x. Unfortunately, I don't feel like it would be very useful to make `the stackFileVersion` a local property.  There's a fair amount of code out there that sets `the stackFileVersion` and then calls another user-defined command to do the actual saving.  For that code, making it a local property would be indistinguishable from ignoring it, because local properties are reset on entering a handler.

I'm thinking of trying something like:
- `the stackFileVersion` remains deprecated and is empty by default
- `save ... with version _` is still recommended
- When calling `save`, the version is determined in the following order:
  1. if present, the version from the `with version _` clause
  2. if not empty, the value of `the stackFileVersion`
  3. otherwise, the default version (i.e. "8.0")

I've deliberately consolidated all the information about what stack file format versions are valid into the "stack version" dictionary entry, so that there's only one place in the dictionary that needs to be updated when the list of format's changes.  I've linked to it copiously from both `save`'s entry and `the stackFileVersion`'s.  I think that it's clear enough.  Passing an invalid version string should already be (identically) erroneous with both the command and the property, but I'll check exactly what happens.

Don't worry about the tests -- the `save` command will very happily write over an existing file without troubling anyone with such pedestrian niceties as confirmation dialogs.

@montegoulding `libURLSetSSLVerification` is a problem, yeah.  Is there an open bug report about that?  Global variables that you have to remember to save and restore are, fundamentally, a recipe for defects that are annoyingly hard to track down.

I don't really know what to do about the older formats.  Unlike with widgets, there's no mechanism to detect whether a stack can be successfully saved in a particular version _x_ of the stack format.  Also, there will be existing code that relies on saving with an old version succeeding with data loss.  There seems to be no right answer here -- the only practical way to ensure consistency seems to be to make save always silently lose data when saving in an older version, which doesn't seem like the Right Thing.  (And what if I _really want_ to save in an old version, losing as much data as necessary?)

montegoulding:
@peter-b I think we need to come to the same conclusion we did with the scriptOnly property i.e give the scripter the rope and let them hang themselves ;-) If there needs to be a nanny let it be the IDE and if possible have it throw up a dialog if the user is losing data when saving a stack. Perhaps a parameter on saveStackRequest so anyone wanting to override the IDE nanny dialog would need to lock messages which is reasonable.

peter-b:
@montegoulding That doesn't solve the problem of detecting which versions a stack _can_ be saved with (in order to enable the provision of a warning dialog).

montegoulding:
Indeed! That would definitely involve some extra work unless of course the IDE just says "Saving in an older file format may cause data to be lost." with continue/cancel options.

runrevmark:
This seems to be a backwards compatibility issue rather than anything else. The 'save with version' form is definitely the 'correct' way to do things in the sense that:
- save is a rare operation (relatively speaking)
- if you are saving a stackfile then you should know at the point of save what format you need

I suggest the following:
- stackFileVersion is deprecated
- stackFileVersion's effect only occurs when there is no 'with format' clause in the save command
- we leave the stackFileVersion behavior precisely as it is now in 8 (before this patch)

This meets our goals with regards the IDE - we can update all instances of 'save' with 'save with version' and thus get an error if the requested file cannot be saved in pre-8 format. It can then popup a dialog to tell the user it will be saved in 8.0 format (which only happens if there are widgets) to let them know what is happening.

Leaving 'save' as it is and working with stackFileVersion as it does now, means that existing code will run as it did before.

mwieder:
That seems entirely reasonable to me, although I would have liked the local property thing. But I do see problems both ways with that for existing code. Also I think possibly setting the stackFileVersion on loading a stack (to the current version of the stack) would help in preserving the version, i.e. a stack in 5.5 format would remain that way unless updated by either the "with format" command or the addition of widgets.

montegoulding:
@mwieder Do you mean a read only fileVersion property of a stack? It's probably reasonable that a stack continues to be saved as it's current file version unless it is saved explicitly as a different format. I'm not sure how this would interact with the stackFileVersion global though... perhaps the lesser of the two is what is used???

mwieder:
Monte- no, I wasn't suggesting yet another property. You can tell from the first few bytes of the stack file what the original format is/was. I think that should be preserved unless the stackFileVersion overrides it or the "with format" syntax does or the presence of widgets forces it. There's an IDE preference for that, and it sometimes works... over time there have been various reports of it not working, so I just don't trust it... when I need to specify a stack format I explicitly set the stackFileVersion. It would be nice not to have to remember to do this.

montegoulding:
OK, its essentially the same thing with the exception of the exposed read only property. It makes sense that the format of a stack file doesn't change unless someone explicitly wants it to, however, I think this would create a different set of backwards compatibility issues though so running with @runrevmark's (which is basically what you originally suggested) is I think the only solution.

peter-b:
Hi everyone, I want to make some progress on this.

Key criteria:
- There must be a way to save stacks while ignoring `the stackFileVersion`
- `save stack` must continue to obey `the stackFileVersion`

My current plan is as follows:
1. Leave `the stackFileVersion` how it is
2. Make `save stack` always obey `the stackFileVersion`, even if that means losing data
3. Add `save stack with latest format` which always saves in whatever the newest and fanciest file format is. This would currently use 8.0 format if there are widgets present, and 7.0 format otherwise.
4. Add `save stack with format _` which always saves in the specified format even if that means losing data.

If you have any strong feelings about this, please make them known ASAP!

mwieder:
Peter- I'm ok with what you're suggesting as long as file version preservation is maintained by default (as with the IDE preference). So point 2 would then be "Make save stack always obey the stackFileVersion, even if that means losing data *unless the 'preserve file version' preference is true"

I would think one way of doing this would be to check the preference on loading the stack, and if true, set the stackFileVersion to the stack's version.

peter-b:
@mwieder If "file version preservation is maintained by default" then that actually breaks backward compatibility!  Also, the `save` command is a low-level engine command -- it neither knows nor cares about the IDE or its preferences.  I'm afraid that what you suggest is simply impractical.

The changes I have now made to this pull request ensure that all pre-existing stacks continue to work in exactly the same way -- but all new stacks can safely ignore `the stackFileVersion`.  For all new development, the `save STACK with newest format` syntax should be used, unless an old file format is specifically needed.  This restores the "losing data" behaviour of LiveCode 7 -- in many respects, you could view it as "fixing" a "regression".

The IDE already will use the newest format by default, and will present a click-through warning message if saving in an older format, so that much is unchanged.

mwieder:
Peter- I commented before you pulled in your commits. At an engine level I'm fine with what you're suggesting. My concern is with losing the stack file format preservation in the IDE, and I don't think this in any way messes that up. No worries.

runrevmark:
@livecode-vulcan review ok c840fdf

livecode-vulcan:
💙 review by @runrevmark ok c840fdf68b35ae7598370161307de31f2f757451

livecode-vulcan:
😎 test success c840fdf68b35ae7598370161307de31f2f757451
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/1547)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/1566)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/1543)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/1554)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/1566)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/1567)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/1569)

PR #3398, opened 2015-12-10
Add support for compiling for Linux ARMv6-HF (RaspberryPi!)

livecodefraser:
A few tweaks are needed to revbrowser/libbrowser to ensure that they don't try to build any CEF components for non x86(_64) platforms. Libbrowser can't be completely disabled as it lives in-engine and is referenced elsewhere so it instead lacks any browser instance factories.

The prebuilt libraries for the Linux ARMv6HF target have been added to the downloads server.

There are some changes in the thirdparty repo that are needed for building, too.

I've not actually tried running the engine that this builds - this is just support for being able to build it at all!


peter-b:
@livecodefraser Could you please update `INSTALL-linux.md` accordingly?

PR #3430, opened 2015-12-21
[Bug 16686] libfoundation: Better __MCNativeChar_CheckedFold() accuracy

peter-b:
This patch modifies `__MCNativeChar_CheckedFold()` to return a more
accurate result.  It now looks up the input char in a constant bit
array in order to determine whether it needs to be folded.  The
constant bit array is only 32 bytes, which means it fits into half a
cache line.  Because `__MCNativeChar_CheckedFold()` now only calls
back to `__MCNativeChar_Fold()` if necessary, searching strings for
this change should provide a speedup for case-sensitive string
operations on inputs that contain a large amount of digits or
punctuation.

### DIFF
--- a/docs/notes/bugfix-16686.md
+++ b/docs/notes/bugfix-16686.md
@@ -1,0 +1,1 @@
+# Improve native string performance by using non-folding operations whenever possible
--- a/libfoundation/src/foundation-string-native.cpp.h
+++ b/libfoundation/src/foundation-string-native.cpp.h
@@ -96,11 +96,74 @@ static inline bool
 __MCNativeChar_CheckedFold(char_t p_char,
                            char_t& r_folded_char)
 {
+	/* These bit-tables can be automatically generated from the tables
+	 in __MCNativeChar__Fold() and __MCNativeChar_Uppercase() using a
+	 script like:
+
+	 function MakeCheckMap pLowerMap, pUpperMap
+	    local tResult, tRow, tBit, tChar, tBitMap
+	    put empty into tResult
+	    repeat with tRow = 1 to 8
+	       put 0 into tBitMap
+	       repeat with tBit = 1 to 32
+	          put (tRow - 1) * 32 + tBit into tChar
+
+	          if (item tChar of pLowerMap) is not \
+	                (item tChar of pUpperMap) then
+	             add 1 * (2 ^ (tBit - 1)) to tBitMap
+	          end if
+	       end repeat
+	       put baseConvert(tBitMap, 10, 16) into tBitMap
+	       repeat while the number of chars in tBitMap < 8
+	          put "0" before tBitMap
+	       end repeat
+	       put "0x" & tBitMap & "U" after tResult
+	       put comma & return after tResult
+	    end repeat
+	    return tResult
+	 end MakeCheckMap
+	*/
+	uint8_t t_hi = p_char >> 5;
+	uint8_t t_lo = p_char & 0x1F;
+#if defined(__WINDOWS_1252__) || defined(__ISO_8859_1__)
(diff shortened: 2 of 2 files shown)


peter-b:
Running the benchmarks in release mode for this changeset make for seriously weird results.  This patch _only_ touches `__MCNativeChar_CheckedFold()`, which is provably not **ever** called for "Exact" native string operations -- but the speed of some "Exact" ops is much slower with this change applied.  What's going on? :confounded: 

[perf-native-checked-fold.txt](https://github.com/livecode/livecode/files/68215/perf-native-checked-fold.txt)

runrevmark:
The operations that are affected are Exact operations which use ForwardScan to find the needle in the haystack (as opposed to just a single character). Both Offset and Contains dispatch directly to ForwardScan, which does a dynamic choice based on p_options. The only thing I can think of is the slightly larger code size of the new CheckedFold operation has changed the compilers choice of when to inline, or increased the amount of code between the entry to ForwardScan and dispatch to the Unfolded path at the very end. The logic probably should be:

```
if (options is unfolded)
  return unfolded op

... do general path which optimizes folded ops if it can ...
```

runrevmark:
Note: When doing the optimization I found how __MCNativeOp_IsFolded was implemented had a huge effect on performance. The comparison was the quickest, bit operations slowed things down significantly.

runrevmark:
Actually, I wonder if the code path should look something like this:

```
if (is_unfolded)
{
unfolded:
  return unfolded_op
}
else
{
  if (length < prefold_limit)
  {
    if (!checkfolded(needle))
      goto unfolded

   return prefolded_op
  }
  else
    return folded_op
}
```

The idea here is that 'exact' is the fastest so should be dealt with as likely and first. Then, the next fastest thing is a caseless small needle, so do that next - reusing the unfolded path. The slowest thing is general folding case, so make sure that is least likely and last.

Also, marking certain things as force_inline might help...

livecodefraser:
holiday (n): a period of one or more days taken off work by an employee for leisure.

runrevmark:
That definition hinges on what one defines as 'leisure' ;)

peter-b:
@runrevmark The changes we went through yesterday seem to have fixed the performance regression.

[perf-native-checked-fold-2.txt](https://github.com/livecode/livecode/files/79685/perf-native-checked-fold-2.txt)

peter-b:
Updated the patch slightly to fix a stupid error that broke the testsuite.

[perf-native-checked-fold-3.txt](https://github.com/livecode/livecode/files/79794/perf-native-checked-fold-3.txt)

runrevmark:
@livecode-vulcan review ok da424e0

livecode-vulcan:
💙 review by @runrevmark ok da424e0502d00f5bcfe0ace9f13fb85e5713825b

livecode-vulcan:
😎 test success da424e0502d00f5bcfe0ace9f13fb85e5713825b
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/1539)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/1558)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/1535)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/1546)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/1558)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/1559)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/1561)

PR #3438, opened 2016-01-06
Get the "System" and "User" UI fonts on OSX using NSFont

livecodefraser:
The font that you get by requesting "Helvetica Neue" on OSX is
_not_ the same as the HN display font used by the system. To get
that one, [NSFont systemFontOfSize:] is required. And similarly
for the user font.

This patch changes the CoreText font code to handle those names
specially and return the appropriate magic font.

### DIFF
--- a/engine/engine-sources.gypi
+++ b/engine/engine-sources.gypi
@@ -541,7 +541,7 @@
 			'<(SHARED_INTERMEDIATE_DIR)/src/quicktimestubs.mac.cpp',
 			
 			# Group "Desktop - Mac"
-			'src/coretextfonts.cpp',
+			'src/coretextfonts.mm',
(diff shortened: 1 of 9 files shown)


livecodefraser:
Things that still need to be done:
- [x] choose less problematic names (e.g. "AquaSystemFont")
- [x] ensure that fields use the user font (because Helvetica Neue is for UI only)
- [ ] test on all of 10.9, 10.10 and 10.11

This should also be done before RC1 as it will change the appearance of users' apps. In particular, the display Helvetica Neue font has narrower strokes than the normal HN font and boldness is very subtle.

livecodefraser:
Superseded by https://github.com/livecode/livecode/pull/3447

PR #3495, opened 2016-01-21
🍄 docs: Add specification for LCB widget signals and events.

peter-b:
Initial draft of specification for revised LiveCode Builder signal and event subsystem.

[Rendered version](https://github.com/peter-b/livecode/blob/feature/lcb-signals-and-events/docs/specs/lcb-signals-and-events.md)

**Unresolved issues:**
- [x] What happens when an event handler blocks/waits in the "dispatch" phase (possibly indefinitely)?
- [x] Relationship with LCS `before` and `after` handlers
- [ ] More / better examples

### DIFF
--- a/docs/specs/lcb-signals-and-events.md
+++ b/docs/specs/lcb-signals-and-events.md
@@ -55,6 +55,8 @@ The `emit` statement may return before all the signal handlers for the signal ha
 
 The signal handlers for a signal get run exactly once for each emission.  All of the handlers for an emission must complete before any of the handlers for a subsequent emission begin [4].
 
+The signal handlers for a signal may be run in any order.
+
 The `emit` statement does not return a value [5].
 
 **Notes**:
@@ -129,6 +131,8 @@ Evaluating `the event` outwith an event handler is an error.
 
 `the event` is immutable.
 
+The `the event target` expression evaluates to the event's target control [3].
+
 **Notes:**
 
 1. For example, when creating a mouse event, it is important to capture the position of the mouse at the time the event began to be dispatched into the event data.  If computationally-intensive event handlers are run, the user could move the mouse between event creation and event handling.
@@ -140,6 +144,10 @@ Evaluating `the event` outwith an event handler is an error.
    can easily be implemented as:
 
        the event["mouse position"]
+
+
+3. It might be useful to provide `the event target is me` (or equivalent syntax) to test whether an event is targeted at the handling control or one of its children.
+
 
 ### Event handlers
 
@@ -163,13 +171,15 @@ A control can use the pre-event handler to **capture** an event:
 
     capture event
 
-The `capture event` statement makes the capturing control becomes the "target" for the event; all its children are removed from the list of interested controls [2].
+The `capture event` statement makes the capturing control becomes the "target" for the event; all its children are removed from the list of interested controls [2,3].
 
 **Notes:**
 
 1. As per the `mouseEnter`/`mouseLeave`, implicit passing reflects the expectation that pre- and post-event handlers are usually be used to implement behaviour that complements rather than conflicts with that of other interested controls.
 
 2. This allows composed widgets to "hide" events from their children.
+
+3. Since `capture event` causes the capturing control to become the target, the capturing control can't do anything that requires knowledge of the "original" target in its on-event handler unless it saves `the event target` in the pre-event handler before `capture event`.  On the other hand, there's not much point in having separate pre-event and on-event handlers when capturing because they're guaranteed to be run immediately one after another.
 
 ### Dispatch phase
 
--- a/docs/specs/lcb-signals-and-events.md
+++ b/docs/specs/lcb-signals-and-events.md
@@ -17,7 +17,7 @@ This document describes a new, more rigorous scheme for ensuring that these beha
 
 There are several issues with the way signals and events currently work for controls implemented in LCB.  The problems affect both the way that widgets can be used to construct other widgets (i.e. composed widgets), and the easy of handling widgets alongside legacy controls from a LiveCode Script (LCS) point of view.
 
-1. Signal emission and handling is not very well checked.  At a language level, there is no way for widgets to describe what type of signals they can emit.  Similarly, there is no syntax for distinguishing between widget handlers and "normal" handlers, meaning that (in principal) a handler that is not intended to be a signal handler could end up getting called in a signal context if a child widget happens to start emitting a signal with a matching name.  Finally, there is no facility for checking that signal prototypes match signal handler prototypes.
+1. Signal emission and handling is not very well checked.  At a language level, there is no way for widgets to describe what type of signals they can emit.  Similarly, there is no syntax for distinguishing between widget handlers and "normal" handlers, meaning that (in principle) a handler that is not intended to be a signal handler could end up getting called in a signal context if a child widget happens to start emitting a signal with a matching name.  Finally, there is no facility for checking that signal prototypes match signal handler prototypes.
 
 2. Similar issues to 1. above apply for event handling.
 
@@ -37,47 +37,53 @@ For the purposes of this document we assume that:
 
 ## Signals
 
+### Signal definitions
+
+A LiveCode module may include signal type definitions in its top-level scope [1-2]:
+
+    public signal type <name:id>(<params:formals>)
+
+If the module is imported by another module via the `use` syntax, the signal definitions are made available in the same way as a variable, handler or type definition.
+
+**Notes**:
+
+1. These definitions provide a convenient anchor point for documentation comments about the purpose and usage of a signal.
+
+2. It would be helpful to provide a `com.livecode.signals` module in the standard library that provides definitions of "standard" signals and descriptions of the recommended way to use them.  Examples of standard signals might include `Clicked`, `SelectionChanged`, etc.
+
 ### Signal emission
-
-A widget's LCB source code must explicitly name, and provide prototypes for, the signals that it can emit [1]:
-
-    public signal <name:id>(<params:formals>)
-
-Signal declarations must be placed in module scope.  The signal **name** must be unique in the module.
 
 Signals are explicitly emitted by a widget using an `emit` statement:
 
     emit <name:id>(<params:expressions>)
 
-It is a run-time error to emit a signal with the wrong number or types of emission parameters [2].  The compiler may statically verify that signals are emitted with compatible parameter lists.
+It is a run-time error to emit a signal with the wrong number or types of emission parameters [1].  The compiler may statically verify that signals are emitted with compatible parameter lists.
 
-The `emit` statement may return before all the signal handlers for the signal have finished running [3].
+The `emit` statement may return before all the signal handlers for the signal have finished running [2].
 
-The signal handlers for a signal get run exactly once for each emission.  All of the handlers for an emission must complete before any of the handlers for a subsequent emission begin [4].
+The signal handlers for a signal get run exactly once for each emission.  All of the handlers for an emission must complete before any of the handlers for a subsequent emission begin [3].
 
 The signal handlers for a signal may be run in any order.
 
-The `emit` statement does not return a value [5].
+The `emit` statement does not return a value [4].
 
 **Notes**:
 
-1. These declarations provide a convenient anchor point for documentation comments about the purpose and usage of a signal.
+1. In principle a widget could set properties on itself before emitting a signal.  However, when adding features to LCS previously it's been found that users prefer relevant information to be passed as parameters to messages rather than out-of-band.  Passing parameters to signal emission also helps to ensure that all of the correct setup has been performed for signal handling - for example, if a parameter is added or removed from the signal, it's easy to verify that all relevant code has been updated.
 
-2. In principle a widget could set properties on itself before emitting a signal.  However, when adding features to LCS previously it's been found that users prefer relevant information to be passed as parameters to messages rather than out-of-band.  Passing parameters to signal emission also helps to ensure that all of the correct setup has been performed for signal handling - for example, if a parameter is added or removed from the signal, it's easy to verify that all relevant code has been updated.
+2. Although, initially, signal emission _could_ be implemented by iterating over registered signal handlers and calling them within the execution of the `emit` statement, this could easily lead to arbitrarily-deep signal emission chains going backwards and forwards between LCS and LCB contexts.  By allowing signal emission to be delayed, this is avoided (the current stack can be allowed to unwind before signal dispatch is performed).  Delaying signal handling also helps to reason about the safety of deleting a widget in response to receiving a signal from it.
 
-3. Although, initially, signal emission _could_ be implemented by iterating over registered signal handlers and calling them within the execution of the `emit` statement, this could easily lead to arbitrarily-deep signal emission chains going backwards and forwards between LCS and LCB contexts.  By allowing signal emission to be delayed, this is avoided (the current stack can be allowed to unwind before signal dispatch is performed).  Delaying signal handling also helps to reason about the safety of deleting a widget in response to receiving a signal from it.
+3. Emitting a signal twice from the same handler - even with the same parameters - will cause all of the handlers for that signal to be run twice.
 
-4. Emitting a signal twice from the same handler - even with the same parameters - will cause all of the handlers for that signal to be run twice.
-
-5. This is a consequence of asynchronous signal handling.
+4. This is a consequence of asynchronous signal handling.
 
 ### Signal handling
 
 Signal handlers are explicitly declared in an LCB program using a `signal handler` declaration [1]:
 
-    signal handler for <name:id>(<params:formals>) is <handler:constexpr>
+    signal handler for <name:id> is <handler:constexpr>
 
-At most one signal handler for each distinct signal prototype may be declared.  The **handler** must conform to the signal prototype.
+At most one signal handler for each distinct signal may be declared.  The **handler** must conform to the signal's prototype.
 
 Signal handlers have no return value.
 
@@ -87,19 +93,32 @@ It is an error to evaluate `the signal source` outside a signal handling context
 
 **Notes:**
 
-1. This allows a single LCB handler to be used to handle multiple signals.  It allows multiple signals with different prototypes to be handled by a single widget's script without overloading.
+1. Explicitly connecting signals to handlers is helpful because it allows:
+
+   * Using a single handler for multiple signals with the same formals
+   * Handling multiple signals with the same name but different formals (it would be necessary to use a fully-qualified signal **name** when doing this)
 
 ### Signal connection
 
-When a child widget is added to a composed widget, each declared signal of the child widget is examined.  If the parent widget declares a signal handler with a matching prototype, then it is connected to the signal [1].
+When a child widget is added to a composed widget, the composed widget is connected to receive signals from the child.  If a signal is emitted by the child widget, the corresponding handler (if any) is run in the composed widget.
 
-When a child widget is removed from a composed widget, all signal handlers in the parent widget that are connected to a signal of the child widget are disconnected.
+When a child widget is removed from a composed widget, the composed widget is disconnected from the child widget's signals.
 
 **Notes:**
 
-1. When these rules are applied to "host widgets", each public script `on` handler in the host widget's script is automatically considered to have been declared as a polymorphic signal handler.
+1. When these rules are applied to "host widgets", each public script `on` handler in the host widget's script is automatically considered to have been declared as a polymorphic signal handler for all signals with a matching name.  For example, an `on Clicked` handler in a host widget's script would handler both `com.livecode.signals.Clicked` and `org.example.signals.Clicked` signals.
 
 ## Events
+
+### Event declarations
+
+A LiveCode module may include event declarations in its top-level scope [1]:
+
+    public event type <name:id>
+
+**Notes:**
+
+1. It would be necessary to provide a `com.livecode.events` module that includes event type definitions for all of the engine events.  With the events API described in this document, it wouldn't really be useful for other modules to declare new event types because only the engine can enqueue events.
 
 ### Event processing
 
@@ -111,9 +130,9 @@ Events are processed in three phases [1]:
 
 3. **Leave**: perform any tasks that should be carried out after event handling (for example, tearing down any state set up during event enter)
 
-Before beginning to process an event, the event manager first determines which control should be the **target** of the event (taking into account visibilities, and bounding rectangles) [2].
+Events are processed synchronously.  If an event is submitted to the event manager while another event is being processed, that event will be queued.
 
-A list of "interested controls" is built.  This recursively contains every parent of the target.  This is used to determine which objects should have their event handlers run.
+Before beginning to process an event, the event manager first determines which control should be the **target** of the event (taking into account visibilities, and bounding rectangles) [2].  A list of "interested controls" is built.  This recursively contains every parent of the target.  This is used to determine which objects should have their event handlers run.
 
 **Notes:**
 
@@ -123,30 +142,19 @@ A list of "interested controls" is built.  This recursively contains every paren
 
 ### Event data
 
-When an event is submitted to the event manager, it has a **event name** and a set of **event data**.  The event data is an array containing all of the captured data relating to the event itself [1].
+When an event is submitted to the event manager, it has a **event type** and a set of **event data**.  The event data is a datastructure containing all of the captured data relating to the event itself [1-2].
 
-In an event handler, `the event` evaluates to the event data array [2].
-
-Evaluating `the event` outwith an event handler is an error.
-
-`the event` is immutable.
-
-The `the event target` expression evaluates to the event's target control [3].
+The `the event target` expression evaluates to the event's target control [3-4].
 
 **Notes:**
 
 1. For example, when creating a mouse event, it is important to capture the position of the mouse at the time the event began to be dispatched into the event data.  If computationally-intensive event handlers are run, the user could move the mouse between event creation and event handling.
 
-2. This permits simple implementation of syntactic sugar for frequently-needed event data.  For example:
-
-       the mouse position of the event
-
-   can easily be implemented as:
-
-       the event["mouse position"]
-
+2. Various syntax related to event parameters (such as `the mouse position`) should probably refer to the event data of the latest event processed.
 
 3. It might be useful to provide `the event target is me` (or equivalent syntax) to test whether an event is targeted at the handling control or one of its children.
+
+4. If `the event target` is a widget, then `the target` LCS expression evaluates to the corresponding host widget.
 
 
 ### Event handlers
@@ -156,6 +164,16 @@ All event handlers (for enter, dispatch and leave phases) must conform to:
     public type EventHandler() returns any
 
 The return value of all event handlers is discarded.
+
+If an event handler includes a statement that "blocks", i.e. pauses script execution while preventing events from being handled, it prevents all further event handling until it completes (although new events may be enqueued) [1].
+
+If an event handler includes a statement that "yields", i.e. pauses script execution while permitting events to be handled, it "detaches" from event processing (i.e. event processing will move on to subsequent handlers for the current event while the script is paused) [2].
+
+**Notes:**
+
+1. This approximates the effect of `wait` statements and `url` expressions in LCS.  In these cases, the user expects nothing to happen until the expected operation completes, which precludes continuing to handle events.
+
+2. This approximates the effect of `wait with messages` statements in LCS.  `wait with messages` implies that events should continue to be handled while the script that calls it is paused, which includes event handling.  However, only one event can be being processed at a time.  A good way to reconcile these two requirements is to make the event manager treat an event handler that yields as if it had returned.
 
 ### Enter phase
 
@@ -195,11 +213,15 @@ An on-event handler can use the `pass event` statement to prevent the default-bl
 
 The `pass event` statement does not return.
 
+If an on-event handler yields, it is treated as having blocked further on-event handler execution.  It is an error to `pass event` after yielding [3].
+
 **Notes:**
 
 1. Since most widgets will only need to handle events during the dispatch phase, on-event handlers may be declared without a qualifier.
 
 2. This is for conceptual compatibility with LCS.
+
+3. Since yielding detaches an event handler from event processing, there is subsequently no "next handler" for the on-event handler to pass the event on to.
 
 ### Leave phase
 
@@ -216,7 +238,3 @@ It is not possible to capture or block post event handler execution.
 **Notes:**
 
 1. All controls that run a pre-event handler are guaranteed to get their post-event handler run.
-
-## Event handler connection
-
-The handling of LCS events is unchanged.
--- a/docs/specs/lcb-signals-and-events.md
+++ b/docs/specs/lcb-signals-and-events.md
@@ -65,7 +65,7 @@ The `emit` statement does not return a value [5].
 
 3. Although, initially, signal emission _could_ be implemented by iterating over registered signal handlers and calling them within the execution of the `emit` statement, this could easily lead to arbitrarily-deep signal emission chains going backwards and forwards between LCS and LCB contexts.  By allowing signal emission to be delayed, this is avoided (the current stack can be allowed to unwind before signal dispatch is performed).  Delaying signal handling also helps to reason about the safety of deleting a widget in response to receiving a signal from it.
 
-4. Emitting a signal twice from the same handler - even with the same parameters - all of the handlers for that signal with
+4. Emitting a signal twice from the same handler - even with the same parameters - will cause all of the handlers for that signal to be run twice.
 
 5. This is a consequence of asynchronous signal handling.
 
--- a/docs/specs/lcb-signals-and-events.md
+++ b/docs/specs/lcb-signals-and-events.md
@@ -1,0 +1,212 @@
+# Signals and Events for LiveCode Builder Widgets
+Copyright © LiveCode Ltd.
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


peter-b:
@trevordevore @montegoulding @livecodeali @livecodefraser @runrevmark

I'd appreciate some prompt feedback, please, since getting this sorted is blocking LiveCode 8 RC 1.

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,222 @@
...
+4. Emitting a signal twice from the same handler - even with the same parameters - will cause all of the handlers for that signal to be run twice.
+
+5. This is a consequence of asynchronous signal handling.
+
+### Signal handling
+
+Signal handlers are explicitly declared in an LCB program using a `signal handler` declaration [1]:
+
+    signal handler for <name:id>(<params:formals>) is <handler:constexpr>
+
```
montegoulding, line 85: @peter-b As with event handlers it seems to me that it is unlikely that you would declare multiple handlers for a signal or event so would it be simpler to just replace public with `signal` or `on event` in the handler?

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,222 @@
...
+All event handlers (for enter, dispatch and leave phases) must conform to:
+
+    public type EventHandler() returns any
+
+The return value of all event handlers is discarded.
+
+### Enter phase
+
+During the enter phase, **pre-event** handlers are run.  They are run from "outside-in"; a parent widget's pre-event handler runs before the pre-event handler of its child widgets.
```
montegoulding, line 180: @peter-b I understand the desire to override child controls but I'm a bit worried about the way this turns the object model on its head. Should `host widgets` also have the same rights? I know it's a huge ask at this stage but I'd rather see subclassing of child widgets in cases where you want to alter their behavior or have them ignore signals/events etc...

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,222 @@
...
+    emit <name:id>(<params:expressions>)
+
+It is a run-time error to emit a signal with the wrong number or types of emission parameters [2].  The compiler may statically verify that signals are emitted with compatible parameter lists.
+
+The `emit` statement may return before all the signal handlers for the signal have finished running [3].
+
+The signal handlers for a signal get run exactly once for each emission.  All of the handlers for an emission must complete before any of the handlers for a subsequent emission begin [4].
+
+The signal handlers for a signal may be run in any order.
+
```
montegoulding, line 67: One of the things I've been contemplating is whether the host widget script could act as a delegate for the widget. I think using this model we would need to call a public handler on the widget from the scripted signal handler as emit has no return value. It's probably not as neat as being able to return data but it would work...

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,222 @@
...
+When an event is submitted to the event manager, it has a **event name** and a set of **event data**.  The event data is an array containing all of the captured data relating to the event itself [1].
+
+In an event handler, `the event` evaluates to the event data array [2].
+
+Evaluating `the event` outwith an event handler is an error.
+
+`the event` is immutable.
+
+The `the event target` expression evaluates to the event's target control [3].
+
```
montegoulding, line 148: As I can't come up with any functional difference between `the event target` and `the target` might I suggest just using the existing syntax?

docs/specs/lcb-signals-and-events.md
peter-b, line 85: @montegoulding There are three reasons that a separate declaration is necessary:
- Handling multiple signals with distinct **name**s but the same **formals** with a single handler
- Handling multiple signals with the same **name** but different **formals** (LCB handlers will never be able to be overloaded)
- Avoiding ambiguity when a widget can both emit and handle a signal with the same name, without enforcing specific handler name mangling.

I will add more notes tomorrow!

docs/specs/lcb-signals-and-events.md
peter-b, line 180: @montegoulding Host widgets _should_ have the same rights, yes.  It's not just about overriding child controls; it's the only way to be able to properly handle `mouseEnter` / `mouseLeave`.  Subclassing every child widgets is not feasible when implementing generic container widgets (which is something that's required in the future; consider a "widget datagrid").

Note that the model proposed here matches the HTML5 event model quite closely.

docs/specs/lcb-signals-and-events.md
peter-b, line 67: @montegoulding Signals are asynchronous and are very specifically _not_ intended to be used like method calls. I guess if you wanted to be really abusive, though, the script could set a property on `the signal source`, since it's currently possible to assume that there will be only a single handler connected to a given signal.

docs/specs/lcb-signals-and-events.md
peter-b, line 148: @montegoulding The main objective here is to make sure that all syntax we add to LCB is unambiguous as to what it does.  There are already quite a lot of differences between LCB and LCS in order to achieve that... I guess it could go either way.

In our design discussion I guess we were thinking that it would be best to be as explicit as possible to begin with, because we can always add simpler synonyms later if necessary.

Heh. There's always `the target of the event` I suppose, to choose a 3rd way...

docs/specs/lcb-signals-and-events.md
montegoulding, line 85: @peter-b Regarding point 2 is this to deal with multiple child widgets that may emit the same signal name? One thing is it could be feasible for different child widgets to emit the same name with the same formals but for logically different reasons or at least the parent widget author may want to handle them separately. Perhaps the `<name:id>` should/could include the widget kind? Maybe a nice thing to do would be something like:

```
signal handler for [<widget kind>|*]:<name:id>(...
```

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,222 @@
...
+An on-event handler is declared using an `event handler` declaration [1]:
+
+    [on] event handler for <name:id> is <handler:constexpr>
+
+By default, if an on-event handler is found for an event, then it **blocks** further on-event handler execution for the event [2].
+
+An on-event handler can use the `pass event` statement to prevent the default-blocking behaviour.
+
+The `pass event` statement does not return.
+
```
montegoulding, line 215: @peter-b can you clarify if uncaptured events are considered captured by the host widget? What I mean by that is if a widget and its children don't mess with an event can a handler in the host widget script capture it? A related question is if a child widget captures and passes an event what is the value of `the target` in LCS? If `the target` and `the event target` are different in that case then ignore what I said before about it...

docs/specs/lcb-signals-and-events.md
trevordevore, line 67: @peter-b @montegoulding Looking over how `emit` works I'm not sure how one would go about designing a widget data grid that displays data from an external source. The host widget itself shouldn't know anything about the data it is displaying. It should request the data and then feed it to a child widget (the one in charge of displaying a row). The trick is that the widget only wants data that needs to be displayed so it will request data as the user scrolls.

Historically what I've done is dispatch a message, `DataForRow`, whenever a row comes into view. This message is handled further on down the message path (e.g. the card). The handler uses a variable (`@pData`) that is passed by reference. The card provides data from a database, XML file, etc. for the a specific row by placing it in the `pData`. Since signals are going to be asynchronous I don't see how this type of design would work with widgets. The signal would be emitted, LCS would gather data for the requested row(s), and then notify the widget that the data is ready to be displayed.

I may be overlooking something obvious. But how would one go about feeding data from an external source to a widget for display on an as-needed basis?

montegoulding, line 67: Yep, that's one of the scenarios I was considering in my comment. As we have discussed recently I tend to do stuff asynchronously where possible so I'm not going to argue against the plan for `emit`. Actually it's possible calls to the data source would also be asynchronous (imagine a JSON api with paging as a data source) so even if we could return something from the signal it won't cover all use cases. My suggestion is splitting `DataForRow` into `LoadDataForRow` and `SetDataForRow`. `SetDataForRow` could be called at any time but it's likely to be from within the running `LoadDataForRow` signal. You might even have a `CancelDataForRow` signal if the user is scrolling too fast to bother loading some data...

docs/specs/lcb-signals-and-events.md
peter-b, line 215: @montegoulding LiveCode Script doesn't know about child widgets at the moment, so `the target` for any event targeted at a widget would be the widget host control.  I'll add a note.

I'm unclear what you mean by "uncaptured" events in this context... could you elaborate?

Taking a guess, I think what you mean is "during the dispatch phase, if no widget code handles the event, does the widget host control's script receive the event"?  In that case, the answer is yes.

peter-b:
@montegoulding @trevordevore I've been thinking about the possibility (and was discussing with @livecodeali earlier today) the possibility of simplifying the events & signals stuff a bit more.

At the moment, we’re assuming that signal and event handling is based on a textual match between handler name and signal name.  However, there are going to be a whole host of “standard” signals that widgets are going to want to reuse, such as `clicked` and `selectionChanged`

If we make signals and events first-class types, then it could simplify everything quite a lot.

We can create a library that provides definitions for “standard” signals:

```
module com.livecode.signals
public signal type Clicked()
…
end module
```

Widgets that want to emit the signal can then emit it by name in almost exactly the same way:

```
widget org.example.SignallingWidget
use com.livecode.signals
private handler synthesizeClick()
   emit signal Clicked()
end handler
end widget
```

And widgets that need to handle the signal can do so in an unambiguous and type-checked fashion, knowing that they’re handling the specific signal that they want to:

```
widget org.example.ClickReporter
use com.livecode.signals
signal handler for Clicked is ConsumeClick
private handler ConsumeClick()
   log "%@ was clicked!" with [the signal source]
end handler
end widget
```

Finally, having a standard library of well-known signals provides a useful place to document best practice, and lets composed widget authors more easily know how child widgets are expected to behave.

And so on for events.

docs/specs/lcb-signals-and-events.md
trevordevore, line 67: The issue I see with an asynchronous call is that you can ensure that the data will be provided before the widget needs to display the data on the screen. If the widget is being fed from a local data source then that sort of visual delay should never happen.

But perhaps the old way of feeding data into an LCS control isn't the way to handle it with widgets.  Maybe using signals for getting data into a widget is the wrong approach. 

I've been toying with the idea of a DataProvider module that would provide a common interface that could feed data from different sources into a widget. The widget would call defined methods in the DataProvider when it wants data and the DataProvider returns it. I don't know if this helps the situation at all but I was thinking about something like this:

```
local theData

# Create a data provider from text. Assumed to be line delimited.
put "Line 1" & cr & "Line 2" & cr & "Line 3" into theData
create dataprovider "MyLineProvider" with theData

# Create a data provider from a numerically indexed array.
put "Line 1" into theData[1]["label"]
put "Line 2" into theData[2]["label"]
put "Line 3" into theData[3]["label"]
create dataprovider "MyArrayProvider" with theData

# Create a data provider that uses a callback to get data.
create dataprovider "MyDatabaseProvider" with callback "GetDataForDatabaseRow" of stack "MyDatabaseStack"
set the number of rows of dataprovider "MyDatabaseProvider" to 300

set the dataprovider of widget "MyRowView" to "MyLineProvider" | "MyArrayProvider" | "MyDatabaseProvider"
```

If a DataProvider was using text or array data then it would just return the data assigned to it. The idea of the callback is that the DataProvider would ask LCS for the data in a synchronous way. But there could also be an asynchronous way of a DataProvider getting data.

I don't know if this approach would make any more sense, but we do need to figure out a way of solving the problem if we can't get external data into a widget while the screen is locked during a scrolling event.

trevordevore:
@peter-b I like this improvement.  The lesson work that has to be done for each individual widget the better.

docs/specs/lcb-signals-and-events.md
runrevmark, line 215: @montegoulding: Widgets are always entirely opaque to LiveCode Script - they are black-boxes. So 'the target' will always be the widget and never its children. As an analog, think of the engine combo-box, that contains a field and the 'popup' button. You can't access the field directly, instead the control provides a means to access the field in the ways that make sense for that control.

docs/specs/lcb-signals-and-events.md
runrevmark, line 180: @montegoulding: I think it is important to remember here that we are talking about aggregation and not subclassing. The parent widget adapts the children as it sees fit so they can work together to provide the overall widget's functionality - this idea requires that the parent can completely control the environment of the widget both how it is embedded, its properties, how it talks to the outside world and what events it receives. 

There is already an element of this in the engine controls - for example, blocking keyDown in a field will prevent the engine action of inserting a key press from happening. However, allowing message sends anywhere is a really bad idea as it leads to inconsistency and potentially problems ensuring mutations to the control tree don't cause unwanted side-effects (I couldn't enumerate the number of bugs in the engine I've fixed over the years related to this).

In widget-parlance, you wouldn't not pass keyDown to let the engine handle it... Instead you'd capture the keyDown in pre.

docs/specs/lcb-signals-and-events.md
runrevmark, line 67: @trevordevore: Whilst in the specific case of _having_ the data you can do that - but what if you don't? In all cases where there is a _large_ amount of data to fetch, a responsive app will want to do the fetching in the background (for example if it was doing a complex query range on a db). There is simply no way to do that with the 'callback returns the data' model as if the callback blocks, you have blocked in a lock screen and thus your app freezes until the data is provided. Thus, really, the 'immediate return' form is a very specific case. So, the issue comes down to caching of the data in advance, and then having a means to indicate in the widget that data is still being loaded - for a truly general data display which allows incremental data supply you have to have both. As long as the caching logic is reasonable, then as you scroll the data will hopefully be cached and available to the widget _before_ that part of it is scrolled to - and in the cases it is not the scrolling would be going so fast that you probably wouldn't notice the fact it had a 'loading marker' anyway.

docs/specs/lcb-signals-and-events.md
runrevmark, line 180: Okay, having written that there is some pondering to do here because it means that the widget event handling mental model does not quite subsume the current engine event handling model (in this case at least - the behavior of engine messages in this regard is very inconsistent).

Obviously for leaf widgets (those which have no children) pre/post are trivial/pointless - as you can do the pre/post actions in the on handler. Thus _perhaps_ an alternative formulation would be something like this:

```
on mouseUp
  do pre event stuff
  [ propagate ] -- this would continue the event down further towards the target
  do post event stuff
  [ pass ] -- this would allow the event to continue to propagate back up to the parent
end mouseUp
```

This is perhaps 'closer' to how the engine works for certain controls - but would mean that anything which 'contains' controls in LiveCode Script which wants to handle a contained control event _after_ the contained control has had a go at it would have to do:

```
on mouseUp
  propagate
  do post event stuff
  [ pass ]
end mouseUp
```

In some ways this eliminates the redundency of pre/post on leaf controls _and_ means that state can be preserved across the descent into the child control within the handler. On the other hand it would be a (not inconsiderable!) change to LiveCode Script.

docs/specs/lcb-signals-and-events.md
trevordevore, line 67: @runrevmark I've coded a fair number of views that display data from web data, local databases, and arrays, all of which have used the 'callback returns the data' model. There are obviously other ways of approaching the problem but there definitely is a way of using this approach and it works quite well and is easy to understand.

The first round of caching always occurs prior to displaying anything in the view. You are merely showing a loading UI element of some sort until you get something to display. The view isn't being asked to display anything yet.

Once you have queried the database, fetched data from the web, etc. you have data in your cache ready to display. The data access will be very quick. 

If you are paging data then when the user scrolls to the bottom of the current data in the cache then you send off a request to the database/web server/whatever and get more data, tell the view that there are more rows to display, and then the view requests the next rows it needs to display. There shouldn't be any lag in the actual callback that returns the data.

trevordevore, line 67: @runrevmark I should clarify something. I'm not using the view to request all of the data. It is only requesting the data to display for specified rows. It is assumed that the data is already in the cache. The view isn't asking the app to request, cache, and then display data. That is handled in the controller who then asks the view to display what is already in the cache.

runrevmark, line 67: @trevordevore: Okay - so your use-case is one where all the data is in memory, but needs processing to turn it into a form for the view to display; thus you use a callback to format data rather than to fetch.

I do wonder whether the callback approach does just 'shift' the 'smooth scrolling' problem down the chain (to the scripter) as it is then entirely a function of the speed of the format request for each row. So, in a case where there is a significant amount of processing required to generate the formatted data from the cached data the application holds (which has previously been fetched from the database) the application will have to hold cached formatted data in addition to the original data - the cost to format this data has to be paid at some point, most likely as the pagination occurs. (Although I appreciate that is handled in the current model as it currently is prepared for a pause at page breaks).

So, to my mind, one way to potentially go about this would be to have the view cache the formatted data (and precache, as necessary) and just display holding rows if it hasn't yet been populated. This means that the formatted data requests can be asynchronous and so it doesn't matter what the requests have to do to get the data to the view. This then means you can then have a view either visibly paginates (pauses at visible points and then extends), or one which scrolls completely smoothly but with the usual 'holding' image turning into real data a small delay after (which is quite a common thing you see in modern apps).

Now, of course, you have a great deal more experience writing dataview type controls than me - so I'm more than willing to be 'shot down in flames' :)

I'm sure it is possible to allow 'synchronous' signals safely - however, they will always come with limitations in what can be done within them. Basically, if a synchronous signal does anything which causes the event loop to be touched then any code which is to execute afterwards becomes 'dangling', and would only occur _after_ things you might not be expecting (redraws, for example). So either, in this case, there would have to be a programmatic error; or the view would have to cope with the situation (which would be equivalent to the ability to be able to display 'pending' rows).

docs/specs/lcb-signals-and-events.md
runrevmark, line 85: Technically you cannot have two signals with the same name and different formals. Signals are declared in a module, so their name is actually 'module_name.short_signal_name'. The idea is that if an identifier for a definition are imported from two different modules then the identifier becomes ambiguous, and you need to use the full name for it in all places. (Note that lc-compile doesn't currently implement this...).

In regards to the automatic attachment then there is more we can do there. At the moment attachment is hard-coded based on name, but you create the child widgets in code. Ideally, there'd either be a means to define the child widgets as definitions in the module (if your widget always uses a fixed set) which would mean you wouldn't need to write code to do it; or a means to dynamically connect signals of just created widgets to handlers in the current module.

peter-b, line 85: See [my top-level comment](https://github.com/livecode/livecode/pull/3495#issuecomment-173874594) about connecting to and emitting uniquely specific signals.  If we implement it that way, then (a) it's easy to reason about (just use the existing definition lookup code and (b) there's no problem with child widgets defining differently incompatible versions of the same named signal.

runrevmark, line 85: Heh - I think (having skimmed the comment earlier and then forgotten about it!) had your further suggestion implicitly in mind when I wrote the above... Yes - the improvement @peter-b suggested combined with the proposed name conflict semantics solve the intent and signature problems.

docs/specs/lcb-signals-and-events.md
montegoulding, line 67: @trevordevore I'm a little confused here. If the data is already cached is the only reason you aren't setting the data in one hit because you don't want to duplicate it in memory? If so then perhaps that's an angle of attack and perhaps that's where you are going with your data provider module which might enable both widget data views and LCS to manipulate a common data model. Like I said before though I think something like this if done right needs abstract modules/libraries or protocols so you can have many data providers that use a common api. There's a few things that need to be in place for that to be a good option though (LCB database access etc).

I think at the moment it just comes down to the view emitting the request for data early enough that the asynchronous callback is not something the user notices unless scrolling really fast or there is a significant time required between request and response.

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,222 @@
...
+The signal handlers for a signal get run exactly once for each emission.  All of the handlers for an emission must complete before any of the handlers for a subsequent emission begin [4].
+
+The signal handlers for a signal may be run in any order.
+
+The `emit` statement does not return a value [5].
+
+**Notes**:
+
+1. These declarations provide a convenient anchor point for documentation comments about the purpose and usage of a signal.
```
montegoulding: @peter-b Have you considered how the use of a common module to declare signals will impact on documentation. It would be handy if things need only be documented once but if the module is one big blob of signals we perhaps need a way to tell the documentation generator which ones we are emitting? Could it be intelligent enough to parse our emit statements? Along similar lines of thought people need to know what events can be handled in the host widget script. Could the documentation generator determine what events are aren't captured and unpassed by the widget and its children.

docs/specs/lcb-signals-and-events.md
montegoulding, line 180: Is there any reason we can't extend the use of before & after handlers to end up with basically the same thing you are proposing for LCB:

```
# stack script
before mouseUp pButtonNumber
   if pButtonNumber is not 1 then
     capture event
   end if
end mouseUp

on mouseUp pButtonNumber
   if pButtonNumber is 3 then
      -- show contextual menu
   end if
end mouseUp

# button script

on mouseUp
   -- we can now ignore the button number
end mouseUp
```

docs/specs/lcb-signals-and-events.md
montegoulding, line 215: @peter-b @runrevmark That's what I was expecting. As it is a functional difference between `the event target` and `the target` I wanted to discuss it so we could establish that reusing `the target` was a bad idea of mine

docs/specs/lcb-signals-and-events.md
montegoulding, line 180: I did a quick test on a behavior hierarchy and found before handlers are evaluated in child -> parent order. In order for my suggestion to be a good one I think we would need to be prepared to change that to child <- parent order and execute all before handlers in the hierarchy even if the before handler is unpassed. I actually think this would be more useful for behavior scripts if it were feasible to make the change. I suspect it is unlikely that huge numbers of people would be impacted by the change as it would only impact those that have overridden a parent behavior before handler with a child one.

docs/specs/lcb-signals-and-events.md
trevordevore, line 67: @runrevmark said:

> Okay - so your use-case is one where all the data is in memory, but needs processing to turn it into a form for the view to display; thus you use a callback to format data rather than to fetch.

@montegoulding said:

> I'm a little confused here. If the data is already cached is the only reason you aren't setting the data in one hit because you don't want to duplicate it in memory?

Both of these statements get to the heart of the issue. The data is already in memory and portions of it need to be displayed to the user. While I currently use dispatch to get data to display, I realize that probably isn't the ideal way to handle things going forward with widgets. I don't think we have everything in place to solve this problem the right way though.

As an extreme (real world) example, think of a SQLite database cursor with 50,000 records. The view only needs to extract data from the records in the cursor that are currently visible (or just out of view). Trying to read everything into the current DataGrid in LiveCode is horrendously slow. Just displaying what is visible makes the DataView really snappy.

Another example is a document object that has a visual representation. As changes are made to the document you don't want to have to update the data in the dataview and update the data in the actual model.

> If so then perhaps that's an angle of attack and perhaps that's where you are going with your data provider module which might enable both widget data views and LCS to manipulate a common data model. Like I said before though I think something like this if done right needs abstract modules/libraries or protocols so you can have many data providers that use a common api. There's a few things that need to be in place for that to be a good option though (LCB database access etc).

I think you are right. It may just be that until we can get to a common data provider api that widgets only get data by setting a property. I've been trying to figure out how to get widgets to play nicely with data in LCS but there may not be a good approach for that just yet. Coming back to the original issue, making synchronous signals so that a widget can pull in row data in order to render it probably isn't the answer.

runrevmark, line 67: @trevordevore said:

> Another example is a document object that has a visual representation. As changes are made to the document you don't want to have to update the data in the dataview and update the data in the actual model.
> Realistically any document view will require cached data to make things render efficiently - there is no difference between that and a dataview. I don't see the problem of a dataview control caching formatted data for it to display - as long as the cache can be trimmed as necessary if more memory is needed for the main app.

So you aren't updating the data in two places - you are storing an 'abstract' set of data in the app which is then used by the view controls to display the data... Caching of formatted data in such view controls is perfectly reasonable.

For example, the field's abstract data is an attributed string but it requires a whole lot of concrete (layout related) data to actually render. Although not quite structured like that at the moment, notionally the field control turns the abstract attributed string into layout structures as needed, dumping appropriate parts of the cached layout data when the field content changes.

You could imagine it structured as this... First you have an inner 'DataViewDisplay' which requires data to be provided to it, ready for display via a callback; then wrapping this you could have the actual 'DataView' wrapping it which provides the caching of formatted data requested async from the host. I'd probably suggest that (morally) the DataViewDisplay would be a semi-concrete baseclass with a overridable method for fetching the data. (By the way, I'm talking conceptually here - at the moment you'd have to write the DataView as a combined DataView/DataViewDisplay).

peter-b:
@trevordevore @montegoulding @runrevmark Here's a new revision, which addresses many of the issues mentioned above.  In particular, behaviour w.r.t. blocking operations is now well-defined, and signals and events are now type definitions.

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,240 @@
...
+### Leave phase
+
+During the leave phase, **post-event** handlers are run.  Like on-event handlers, they are run from "inside-out", starting from the target control.
+
+All interested controls have the opportunity to run a post-event handler, even if an on-event handler blocked the event during the dispatch phase [1].
+
+A post-event handler is declared using a `post event handler` declaration:
+
+    post event handler for <name:id> is <handler:constexpr>
+
```
montegoulding, line 235: My comment about continuing the same model through to LCS and using before and after handlers have been lost. Should I write it out again or was it a silly idea. I might also suggest changing pre and post here to before and after mainly because `post` has multiple meanings.

docs/specs/lcb-signals-and-events.md
```diff
@@ -0,0 +1,240 @@
...
+2. Although, initially, signal emission _could_ be implemented by iterating over registered signal handlers and calling them within the execution of the `emit` statement, this could easily lead to arbitrarily-deep signal emission chains going backwards and forwards between LCS and LCB contexts.  By allowing signal emission to be delayed, this is avoided (the current stack can be allowed to unwind before signal dispatch is performed).  Delaying signal handling also helps to reason about the safety of deleting a widget in response to receiving a signal from it.
+
+3. Emitting a signal twice from the same handler - even with the same parameters - will cause all of the handlers for that signal to be run twice.
+
+4. This is a consequence of asynchronous signal handling.
```
montegoulding, line 78: What's a consequence? Should this sentence be in 3?

peter-b, line 78: The fact that `emit signal` does not return a value is a consequence of asynchronous signal handling.

montegoulding, line 78: Ah, sorry

docs/specs/lcb-signals-and-events.md
peter-b, line 235: It wasn't lost!

Unfortunately, and as you quite rightly pointed out in a previous comment, the `before` and `after` handlers in LCS work have quite a different semantic that relates to behaviours and not really specifically to events.  @runrevmark and I had a long discussion about whether it was possible to rejig the way they work so that they can be used to do "pre" and "post" event handling.  Unfortunately, we couldn't find a way to make that sane, so for the time being I think the plan is to make "pre" and "post" event handling available only in LCB.

This also gives the opportunity to check that the event model makes sense, since we are reserving the right to make backwards-incompatible changes to LCB in the future (but we don't have that luxury for LCS).

trevordevore, line 235: That makes sense. I'm looking forward to using the new pre/post approach. I like the idea of the outside-in/inside-out path for the events.

peter-b, line 235: It's tricky.  In theory there are ways to do things in LCS that result in getting messages in a weird order.  Imagine that you have a button with the following script:

```
on mouseEnter
   wait 10
   pass mouseEnter
end mouseEnter

on mouseLeave
   pass mouseLeave
end mouseLeave
```

On a card with the following script:

```
on mouseEnter
   put "mouseEnter" && the target & return after msg
end mouseEnter

on mouseLeave
   put "mouseLeave" && the target & return after msg
end mouseLeave
```
1. What would you expect the message box to contain when you move the mouse pointer over and away from the button?
2. What is the actual output that you get?

Most of this spec is designed to make sure that, for LCB widgets, expectation and reality match!

montegoulding, line 235: @peter-b I see what you mean about the event order.

peter-b, line 235: @trevordevore @montegoulding Even if LCS behaviour is left unchanged, it's not clear how things should be ordered...

If there's an LCS `mouseUp` handler in a widget's script, and that `mouseUp` handler waits, when should the _widget's_ MouseUp post-event handler get run?

trevordevore, line 235: @peter-b I will say this - unbalanced messages have caused me much pain over the years (lock messages can cause this). In the example you mention the following scenario immediately came to mind.

Suppose a widget changes the visual state in mouseDown and then restores the visual state on mouseUp. Shouldn't the visual state remain in the mouseDown state until mouseUp has completed? If the widget gets the post-event handler before the wait finishes, then the visual state would return to normal even though the widget is still going through a mouseUp operation due to the LCS wait. Visually this would deceive the end user into thinking that the operation associated with the click on the widget had actually completed when it hadn't.

peter-b, line 235: @trevordevore Well, I suppose there's two things here.
1. Modern UX design practice would recommend that you disable the button in the mouseUp handler, show some sort of "I am busy" UI, and then `send BeginOperation to me in 0`.  Then, when the operation ends, you'd hide the "I am busy" UI, and enable the button again. That would of course work perfectly under the scheme proposed in this spec.
2. In practice, people like to do very costly things in mouseUp handlers, like performing 23 database queries over a dial-up connection to a database running on a Z80 in Antarctica.

I feel like the latter is probably something that should be discouraged, though.

trevordevore, line 235: @peter-b Regarding (1) - I know that and you know that, but you as you state in (2) lots of LiveCode users don't do that. (2) should definitely be discouraged but it also involves a deeper understanding of why code should be written that way. 

Coding becomes more complicated when you start thinking about async operations, disabling controls while an operation is performed, or executing a handler AFTER the mouseUp event has completed using <send>. While these types of approaches are what I use in my apps, I don't think LiveCode has a LiveCode-like approach to dealing with these problems. For example, `send MESSAGE in 0 seconds` isn't self-documenting at all.

But the original question is about when the post-event handler in the widget should be triggered. Without taking any of the implementation details into account, I would say the messages between LCS and LCB should be balanced. If that proves too difficult given the types of things that can go on in LCS during any given event then the world won't end.

peter-b, line 235: @trevordevore What do you mean when you say, "the messages between LCS and LCB should be balanced"?  I'm not clear -- could you please give an example? I just want to make sure I understand what you're proposing here.

trevordevore, line 235: In your original question you asked when the post-event should be sent if the LCS mouseUp event has a wait. My understanding of the question is that the post-event in LCB can be triggered after the LCS mouseUp event finishes (which would mean waiting until `wait` is done) or the post-event in LCB can be triggered in the same engine cycle that produced the mouseUp event. In the later case, the post-event in LCB would trigger before the LCS mouseUp event finished. In my mind the messages are then unbalanced because LCB is finishing before LCS does.

peter-b, line 235: Right, but here's the problem -- the LCB widget implementation then has to be able to deal with, for two mouse clicks, the corresponding MouseDown, MouseUp, MouseDown, MouseUp post-event handlers being run _in any order_ or possibly _not at all_ (because nothing says that a LCS handler actually has to complete). Not to mention the fact that it is then _impossible_ to ensure that only one event is in-flight at any given time.  I'm not convinced that that behaviour is either sane or feasible to implement.

This is a discussion that's been going round-and-round-and-round-and-round between me and @runrevmark for some time now... Both options are basically terrible.

trevordevore, line 235: @peter-b If it is impossible to ensure that only one event is in-flight at any given time then you know it isn't a viable solution. Problem solved :-) So the sane solution is to have LCB event always sent within the same cycle the message was generated, right? That is straight forward.

So what types of problems does this lead to? LCS and LCB can now become out of sync with each other when it comes to event message order. The widget developer has no control over this as it is the developer using the widget that can cause this. The developer creates out of sync state by doing anything the blocks (wait, db operations, etc.) during one of these events.

Does that sum up the issue?

peter-b, line 235: @trevordevore The current version of the spec would ensure that, if you block in an LCS handler, the LCB post-event handlers would be dispatched immediately and then normal LCS message handling would resume.  This would mean that the widget's "leave"-phase handlers would be run while LCS is still technically in the "dispatch" phase.

I think it's a reasonable compromise. The LCS developer who blocks in an event handler would have to take responsibility for any ensuing out-of-sync state -- but that doesn't introduce any problems that aren't already present in LCS.

trevordevore, line 235: I agree that it is a reasonable compromise. Carry on.

PR #3497, opened 2016-01-22
[Bug 16738] Ensure widgets in substacks are saved.

peter-b:
Although there was an `MCStack::substackhaswidgets()` method defined,
`MCObject::haswidgets()` wasn't calling it.

When saving stack files with widgets only in substacks (not in the
main stack), the save logic thought that no widgets were present and
downgraded the file format to 7.0, causing widgets in substacks to be
discarded.

This patch makes `MCObject::haswidgets()` virtual and overrides it in
`MCStack` with an implementation that calls `substackhaswidgets()`.

### DIFF
--- a/docs/notes/bugfix-16739.md
+++ b/docs/notes/bugfix-16739.md
@@ -1,0 +1,1 @@
+# Save widgets in substacks when main stack doesn't have widgets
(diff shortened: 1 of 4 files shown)


peter-b:
Argh, wrong bug number!

PR #3531, opened 2016-01-28
[Bug 16768] Added bugfix note

livecodepanos:
Bug 16768 was fixed in livecode-thirdparty, where there is no `notes` folder, thus I am adding the bugfix note here


livecodepanos:
Just a bugfix note added, thus no need for review

PR #3541, opened 2016-01-29
[[ Community Docs ]] Specified that time is in epoch seconds

angerangel:
for the pendingMessages function. Moreover I switched from unordered list to numbered list, regarding the items of each line of the pendingMessages list.

### DIFF
--- a/docs/dictionary/function/pendingMessages.lcdoc
+++ b/docs/dictionary/function/pendingMessages.lcdoc
@@ -22,7 +22,7 @@ repeat with x = 1 to the number of lines of the pendingMessages
 
 The result: The message ID is the same as the one placed in the result function when it was scheduled with the <send> command.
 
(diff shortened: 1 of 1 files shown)


livecodeali:
@angerangel Thanks for this. There is one minor change needed - the return value type is not an enum, so this can just be removed. Otherwise the entry will not be formatted correctly. (It won't be formatted correctly anyway at the moment, until https://github.com/livecode/livecode-ide/pull/842 is merged)

peter-b:
@livecode-vulcan review ok 003d4ad

peter-b:
@angerangel Thanks for making that tweak. Looks good to me. :+1:

livecode-vulcan:
💙 review by @peter-b ok 003d4adf38a371af845586af796e9f5b3b7052ac

PR #3567, opened 2016-02-05
[[ Bug 13807 ]] Always notify control's parent of resize when calling layer_setrect

livecodeali:


peter-b:
Am I right in thinking that it's safe to remove the calls to `resizeparent()` because the call you added to `layer_setrect()` covers all of them?

livecodeali:
I must admit I didn't go through the logic that way - my reasoning was that if the control is either invisible or unopened then it never needs to call resizeparent. Thinking about it now, I'm not sure that's true for unopened - I will have another look.
(@livecodeian may know off the top of his head)

runrevmark:
If a control is invisible it certainly then I think resizeparent() is still applicable... Group two buttons, hide the group, set the topLeft of the first button and check the rect of the group. I think the same is true of unopened, as you can set properties on a control which mean resizeparent() needs to be called and they aren't guarded by being opened.

runrevmark:
Looking into the original bug report - the relevant place where 'resizeparent()' is not being done is in MCUIDC::handlemoves(). Now, this is for interpolating the position of an object over time - which essentially amounts to automatically setting its rect as time ticks by. Therefore, it might be better to change ::handlemoves() to use SetRectProp() of the target rather than setrect().

However the problem is then the 'without messages' form... At the moment as controls are moved by the move command I don't believe they update mouse focus appropriately. If we use the SetRectProp() method than mouse focus will be synchronized if necessary, sending a message, which shouldn't occur if 'without messages' is specified on the command. Indeed, this can be verified by using a button with a hover icon, and moving it to the mouse loc, you need to move the mouse again at the end before the hover state appears even though the mouse is over it.

Certainly if 'without messages' isn't specified then the mouse focus should synchronize at each change; if it is specified should it synchronize at the end?

peter-b:
@livecodeali, @runrevmark: Is this fix blocking any freeze items? If not, let's bump it.

livecodeali:
@peter-b No, it should be bumped.

PR #3620, opened 2016-02-23
[[ Per-Edition Docs ]] Build per-edition dictionary and guides

livecodeali:
Required: https://github.com/livecode/livecode-private/pull/69

### DIFF
--- a/Installer/package.txt
+++ b/Installer/package.txt
@@ -357,9 +357,21 @@ component Documentation
 		file "ide:Documentation/rev/videos.rev"
 
 	into [[ToolsFolder]]/Documentation place
-		rfolder "ide:Documentation/guides"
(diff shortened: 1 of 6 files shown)


peter-b:
See [bug 16556](http://quality.livecode.com/show_bug.cgi?id=16556).

peter-b:
@livecode-vulcan review ok 913307d

peter-b:
@livecodeali Are there any changes needed to the installer manifest to make sure that the correct docs go into each edition's installers?

livecodeali:
@peter-b I don't think so - as far as I can see each edition's installers just use the current 'output' ide. So I've just made it overwrite the relevant data each time.

livecodeali:
Superseded by https://github.com/livecode/livecode/pull/3631

PR #3623, opened 2016-02-23
[[ Bug 16983 ]] GradientRampEditor put upper handle higher

BerndN:
added accidentally deleted y value for drawing upper handle of selectedStopp


peter-b:
@BerndN The bugfix note needs to be in a quite specific format, so that the tool that constructs the release notes can correctly put the bug information into the correct place.

Please make the contents of the file something like:

``` markdown
# [16983] Put GradientRampEditor upper control handle higher
```

I'll explain:
1. The release notes builder already knows that the file relates to the GradientRampEditor, and will put the information in the correct place
2. The line beginning with `# [NUMBER] SUMMARY` says, "Record that bug NUMBER was fixed, and the change summary is SUMMARY"; it must all be on one line

peter-b:
Unfortunately if the format is wrong then the release notes builder will generate an error, and it will break the build :disappointed:

PR #3654, opened 2016-02-27
Fix externalv1 compiler warnings

mwieder:
Now that the gyp thing is fixed up I can compile properly again.
The changes here get rid of the compiler warnings about unused enums in a switch statement and the need for parentheses in the octal parsing test. End result is that the externalv1.cpp file no longer generates compiler warnings.

### DIFF
--- a/engine/src/externalv1.cpp
+++ b/engine/src/externalv1.cpp
@@ -280,8 +280,10 @@ static MCExternalError string_to_real(MCStringRef p_string, MCExternalValueOptio
 		return kMCExternalErrorNotANumber;
 	
 	// Now see if it has to be interpreted as an integer (0x or 0 prefix).
-	if (*s == '0' &&
-		(l >= 2 && (s[1] == 'X' || s[1] == 'x')) ||
+	// MDW-2016-02-24 : [[ fix_compiler_warnings]] t_error was not getting checked
+	//								also added many parentheses to placate compiler warnings
+	if ((*s == '0') &&
+		(l >= 2 && ((s[1] == 'X') || (s[1] == 'x'))) ||
 		options_get_convert_octals(p_options))
 	{
 		MCExternalError t_error;
@@ -514,6 +516,9 @@ MCExternalError MCExternalVariable::Append(MCExternalValueOptions p_options, MCE
 	MCExternalError t_error;
 	MCAutoStringRef t_string;
 	t_error = p_value -> GetString(p_options, &t_string);
+	// MDW-2016-02-24 : [[ fix_compiler_warnings]] t_error was not getting checked
+	if (t_error != kMCExternalErrorNone)
+		return t_error;
 	return AppendString(p_options, *t_string);
 }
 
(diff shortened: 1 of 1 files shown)


peter-b:
@mwieder Thank you very much for going through and fixing these warnings!

However, please don't add explanatory comments to the code for changes as trivial as adding disambiguating parentheses! They don't add any value. Could you please revise this PR so that it doesn't add the `// MDW-2016-02-24 : [[ fix_compiler_warnings]]` comments?

Thanks!

mwieder:
Will do, although the comment about the parentheses was there so that someone could verify that I got the grouping right instead of inadvertently changing the result by changing the precendence. Do you want me to remove the other comments as well (e.g. the switch statement)?

peter-b:
It's possible to verify the precedence by reviewing the diff, which quite clearly shows that only `(...)` were added!

Interestingly, the first set of parentheses you've added changes the meaning, because [`||` has lower precedence than `&&`](http://en.cppreference.com/w/cpp/language/operator_precedence).  However, I think that in doing so, you've fixed a bug!  Can you spot what the bug is? :wink: 

In the `switch` statement, the "ignored" cases shouldn't be ignored -- because `MCObject::Dispatch()` should never return them! To make sure that they generate an abort in a debug build, please use:

``` C++
MCUnreachable();
break;
```

instead of just `break`.

mwieder:
I added the MCUnreachable trap. As far as the parentheses go, I went with what I perceived to be the intent of the statement. I realize that examining the diff will show the changes, but the comment was supposed to help ease that process, i.e., not a "this adds 1 to the variable" sort of comment, but more of a "I grouped these, please pay attention to this to make sure I did it right" comment. And no, I'm a bit caffeine-deprived at the moment, so I'm not about to dig into the free-range operators.

peter-b:
@mwieder Thanks :)

@livecode-vulcan review ok 5c84f7c

livecode-vulcan:
💙 review by @peter-b ok 5c84f7cc341eb921875caa99a5e183b3f1a35827

livecode-vulcan:
😎 test success 5c84f7cc341eb921875caa99a5e183b3f1a35827
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/1952)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/1965)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/1946)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/1958)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/1968)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/1973)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/1978)

PR #3696, opened 2016-03-07
Application resume notification during startup causes crash on mac

montegoulding:
MCdefaultstackptr was not being checked before attempting
to send a message to the current card. El Capitan appears to
be sending resume notifications when applications are
launched.

### DIFF
--- a/docs/notes/bugfix-16872.md
+++ b/docs/notes/bugfix-16872.md
@@ -1,0 +1,1 @@
+# Application resume notification during startup causes crash on mac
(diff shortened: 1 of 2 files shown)


montegoulding:
It appears [this](https://github.com/livecodefraser/livecode/commit/97319db08138697d28c9ffcf8590106d63763c3e) might have been handy if it were back ported. I can't use 6.7 on my machine at all without this fix.

It's apparent that my fix isn't for 16872 but for 13480 which produce _very_ similar crash reports.

montegoulding:
Closing in favour of https://github.com/livecode/livecode/pull/3697

PR #3742, opened 2016-03-15
Update our use of LCMS to v2

livecodefraser:
There are some updates required in the thirdparty submodule to update the LCMS header we ship.


livecodeali:
Superseded by https://github.com/livecode/livecode/pull/5477

PR #3753, opened 2016-03-16
[[ Bug 17154 ]] Update mfocused when parent rect is changed

livecodeali:
This fixes the issue seen in the tutorial where clicking on the
inspector menubar icon failed sometimes.

The reason is that before the mouse down message is sent, the
menubar performs a layout which temporarily moves controls in
a group underneath the mouse pointer (which updates the mouse
focus). The group itself is subsequently moved away, which does
_not_ update the mouse focus, as MCObject::SetRectProp only
checks if the card focus is the object being moved.

This fixes the issue by recalculating the mouse focus if the moved
control is an ancestor of the focused control, rather than equal
to it.

### DIFF
--- a/engine/src/object.cpp
+++ b/engine/src/object.cpp
@@ -5327,19 +5327,7 @@ void MCObject::sync_mfocus(void)
             // MW-2012-02-22: [[ Bug 10018 ]] If the target is a group then check
             //   to see if it is the ancestor of the mfocused control; otherwise
             //   just compare directly.
-            if (mfocused != nil && gettype() == CT_GROUP)
-            {
-                while(mfocused -> gettype() != CT_CARD)
-                {
-                    if (mfocused == this)
-                    {
-                        needmfocus = True;
-                        break;
-                    }
-                    mfocused = mfocused -> getparent();
-                }
-            }
-            else if (this == mfocused)
+            if (isancestorof(mfocused))
                 needmfocus = true;
         }
         else if (MCU_point_in_rect(rect, MCmousex, MCmousey))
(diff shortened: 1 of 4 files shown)


livecodeali:
WIP while I work out (or someone else points out) if this is correct and if and how this can be done better, and while I cook up a simple test case (which should be possible now I know what is actually happening).

livecodeali:
I just got a crash from this branch so it is probably incorrect.

engine/src/object.cpp
```diff
@@ -5361,6 +5349,20 @@ void MCObject::sync_mfocus(void)
         MCmousestackptr->getcard()->mfocus(MCmousex, MCmousey);
 }
 
+bool MCObject::isancestorof(MCObject *p_object)
+{
+    if (p_object == nil)
+        return false;
```
peter-b, line 5355: Should this arguably be:

``` c++
MCAssert(p_object != nil);
```

Or is it expected that there will be occasions where it is useful and desirable to call this method for a null argument?

livecodeali, line 5355: I'm not sure - it would be possible to call it on something like `MCtemplatebutton` whose `parent` is `nil`, although I think in all the existing cases it is protected from that by checking the `opened` variable, which presumably is always false for `MCtemplate...`, but I don't know off the top of my head if it's generally true that you can't be `opened` without a `parent`. @runrevmark?

peter-b:
To summarize: it shouldn't be an assertion failure if  `p_object == nil`, but we want to check that this is an ancestor of the parent of the object rather than checking whether the parent of this is an ancestor of the object. Yay!

livecodeali:
@peter-b Updated.

peter-b:
@livecode-vulcan review ok 31057db

livecode-vulcan:
💙 review by @peter-b ok 31057dbe9886e26a2daa9ed1206d08d8974f9e71

livecode-vulcan:
😎 test success 31057dbe9886e26a2daa9ed1206d08d8974f9e71
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/2248)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/2257)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/2241)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/2258)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/2264)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/2269)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/2280)

PR #3825, opened 2016-03-31
Fix an incorrect check in generate-dmg.sh

livecodefraser:
The return codes from hdiutil cannot be trusted. Zero doesn't indicate success and non-zero doesn't indicate failure.


livecodefraser:
Only affects installer builds.

PR #3844, opened 2016-04-04
Ensure the OSX IDE app bundle has universal read/execute permissions

livecodefraser:
This makes the app bundle in the DMG independent of the umask
setting of the build host. Write is disabled for all users,
including the owner, to prevent accidental IDE modifications from
changing the app bundle contents and causing signature verification
to fail.


livecodeali:
@livecode-vulcan review ok 2c0d3ae

livecodefraser:
Merging without waiting for a build as it only affects the installers.

PR #3861, opened 2016-04-06
[[Community Docs]] Add note about waiting for idle

macMikey:
This just clarifies that there is no guarantee that at <time> the message will be sent - if the developer was hoping that some event would be interrupted, they will be disappointed.

### DIFF
--- a/docs/dictionary/command/send.lcdoc
+++ b/docs/dictionary/command/send.lcdoc
@@ -55,4 +55,6 @@ When the send command is used the stack containing the target handler temporaril
 
 >*Note:*  Using the <send> <command> is slower than directly <execute|executing> the <command|commands> using the normal <message path>. For best efficiency, use the <send> <command> only when you want to delay the <message> or when the <handler> you want to execute is not in the <message path>.
 
+>*Note:*  LiveCode will interrupt other handlers when <time> arrives.  It will only check the pendingMessages queue after <time> when it is idle.
+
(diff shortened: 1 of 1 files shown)


livecodeali:
@macMikey I think this information is already covered in the `Important` section above it that states

> If you use the send in time form of the &lt;send&gt; &lt;command&gt; --even if you specify a &lt;time&gt; of zero seconds--the current handler finishes &lt;execute|executing&gt; before the &lt;message&gt; is sent.

livecodeali:
@macMikey Although it could be made slightly more accurate, perhaps by saying

> any handler &lt;execute|executing&gt; when &lt;time&gt; elapses will complete &lt;execute|execution&gt; before the &lt;message&gt; is sent.

macMikey:
You are correct, the note is included.  I missed that, somehow.  I don't
like the suggestion - I would like the language to be plainer.  I'll play
with it and submit.

macMikey:
Hey, Ali, I wonder if maybe we should change this just a little more radically, to make it more clear.  Send really adds the message to the queue, right?  Then LC distributes it when it gets an idle message, right?  Spelling that out in the beginning would make this whole thing much more clear, conceptually.  Let me know what you think.  If you agree, then I'll work on both the send and pendingMessages entries, again.

livecodeali:
Superseded by https://github.com/livecode/livecode/pull/3882/files

PR #3862, opened 2016-04-06
[[ Community Docs ]] Add note about not interrupting

macMikey:
Make it clear that while there is a scheduled time for an event, the event will not actually fire until after any currently executing handlers achieve.


livecodeali:
@macMikey I agree it makes sense to add a similar warning here to that of the send command. Could you phrase it similarly, make something like

> any handler &lt;execute|executing&gt; at the time a pending message is scheduled for will complete &lt;execute|execution&gt; before the &lt;message&gt; is sent.

macMikey:
Yes, something.  I want the words to be plain and obvious so I'll mess with
it.

peter-b:
@macMikey You said on the mailing list that you were not going to persevere with getting these changes merged, so I'm going to close this pull request for now.

If you'd like to get involved again at some point in the future, please reopen it or create a new one!

Thanks 😃

PR #3882, opened 2016-04-07
[[ Community Docs ]]

macMikey:
Ali,
This is what I was thinking to make it more clear what send is doing,
namely, it has two distinct modes and behaviors.


livecodeali:
@macMikey The modified summary is too long here, and you should revert your removal of 'a' and 'an'

```
Either sends a <message> to an <object> immediately, or adds a <message> to the message queue.
```

The links here should be to the message and object glossary entries, the summary should be more high-level. I understand why you have made the change you have, but the correct thing here is to change the names of the parameters in the syntax (perhaps to 

```
send <messageName> [ to <targetObject> [in <time> [seconds | ticks | milliseconds]] ]
```

)

That way the links in the summary will be to the glossary entries rather than internal parameter references.

I think `The result:` description should still say that `the result` is the ID of the message in the <pendingMessages> queue if the `in <time>` form is used.

I'm not keen on the phrasing here:

```
, pausing and interrupting any currently executing handler
```

and I don't think it's necessary, as the meaning is clarified in the following clause. There's nothing complicated or unclear about 

```
Any handler it <trigger|triggers> is completed before the rest of the current handler is <execute|executed>
```

and pause/interrupt is not technically correct in  terms of what is actually happening.

What was the purpose of removing the `<handler>` link?

macMikey:
Ali,
In much of the message, I'm not sure what you are trying to tell me.  Why not tweak what I did, instead, so I can see what you're thinking.

docs/dictionary/command/send.lcdoc
```diff
@@ -4,7 +4,7 @@ Type: command
 
 Syntax: send <message> [ to <object> [in <time> [seconds | ticks | milliseconds]] ]
```
livecodeali: Currently the message and object parameters here are not distinct from the names of existing dictionary entries, which means that the links will not work correctly. I assume that's why you removed the indefinite articles from the summary below. 

The best way to fix this would be to change the syntax to 

```
send <messageName> [ to <targetObject> [in <time> [seconds | ticks | milliseconds]] ]
```

and then make sure all the instances of `<message>` and `<object>` later on in the entry that are supposed to refer to the actual parameter rather than link to the general message and object dictionary entries are changed to `<messageName>` and `<targetObject>`.

docs/dictionary/command/send.lcdoc
```diff
@@ -4,7 +4,7 @@ Type: command
...
 
-Summary: Sends a <message> to an <object>.
+Summary: Either sends <message> to <object> immediately (if "in" is not specified), or adds <message> to the <pendingMessages><function> queue to be sent to <object> at a later time (if "in" is specified).
```
livecodeali: The modified summary is too long here, and you should revert your removal of 'a' and 'an' - perhaps change it to

```
Either sends a <message> to an <object> immediately, or adds a <message> to the message queue.
```

docs/dictionary/command/send.lcdoc
```diff
@@ -38,21 +38,21 @@ message: An expression that evaluates to a message name, possibly including para
 object: Any object reference. If you don't specify an object, the message is set to the object whose handler contains the send command. If you specify a time, you must also specify an object.
 time (integer): If you don't specify a unit of time, the default of ticks is used.
 
-The result: If you specify a <time>, the <message> is sent to the <object> after the specified time has elapsed. The <message> is added to the <pendingMessages> <function>. The ID of the <message> is <return|returned> by the <result> <function>. To cancel the <message>, use the <cancel> <command> with this ID.
+The result: The ID of the <message> in the <pendingMessages> <function> queue.
```
livecodeali: I think `The result:` description should still say that `the result` is the ID of the message in the <pendingMessages> queue if the `in <time>` form is used - how about 

```
The result: If you specify a <time>, <the result> is the ID of the <message> in the <pendingMessages> <function> queue.
```

docs/dictionary/command/send.lcdoc
```diff
@@ -38,21 +38,21 @@ message: An expression that evaluates to a message name, possibly including para
...
 
 Description:
-Use the <send> <command> to override the normal <message path>, or to delay a <command> until a specified time. 
+Use the <send> <command> to override the normal <message path>, or to delay a <command> until after a specified time.
 
-The <message> can be either the name of a <handler> or a <LiveCode> <statement>. If the <message> is a <literal string> containing more than one <word> (for example, if it includes <parameter|parameters>, or if it is a multi-word command), it must be enclosed in <double quote|double quotes>. 
+The <message> can be either the name of a handler or a <LiveCode> <statement>. If the <message> is a <literal string> containing more than one <word> (for example, if it includes <parameter|parameters>, or if it is a multi-word command), it must be enclosed in <double quote|double quotes>.
```
livecodeali: The angle brackets around `<handler>` here are to provide a link to the handler glossary entry- they shouldn't be removed.

docs/dictionary/command/send.lcdoc
```diff
@@ -38,21 +38,21 @@ message: An expression that evaluates to a message name, possibly including para
...
 
->*Important:*  Specifying a <time> can affect the order in which <statement|statements> are <execute|executed>. If you don't specify a <time>, the <message> is sent immediately, and any <handler> it <trigger|triggers> is completed before the rest of the current <handler> is <execute|executed>. If you use the send in time form of the <send> <command> --even if you specify a <time> of zero seconds--the current handler finishes <execute|executing> before the <message> is sent.
+>*Important:*  If you don't specify a <time>, the <message> is sent immediately, pausing and interrupting any currently executing handler.  Any handler it <trigger|triggers> is completed before the rest of the current handler is <execute|executed>. However, if you specify a <time>, even one of 0, the current handler finishes <execute|executing> before the <message> is sent.
```
livecodeali: I don't think

```
, pausing and interrupting any currently executing handler
```

is necessary, as the meaning is clarified in the following clause:

```
Any handler it <trigger|triggers> is completed before the rest of the current handler is <execute|executed>
```

livecodeali:
@macMikey I have commented on sections inline in order to clarify what I'm talking about. Sorry that it wasn't very clear! 

Git pull requests are not very convenient for collaborative tweaking - I would just have to submit a whole new pull request anyway. 

Perhaps it would be better to start with a more minor documentation change first, before tackling such a significant one as you are here.

docs/dictionary/command/send.lcdoc
macMikey: good idea

docs/dictionary/command/send.lcdoc
macMikey: Also a good idea - but I disagree on the article after we change <object> to <objectName> - see latest update.

docs/dictionary/command/send.lcdoc
macMikey: I can live with it.

docs/dictionary/command/send.lcdoc
macMikey: It didn't parse correctly in the preview stack so I assumed it was a mistake.  I probably should have checked the send dictionary entry.

macMikey:
Ali,
I thought this WAS a minor change, which is why I went with it, first.  After we get send done, I'll go back and work on pendingMessages.  FYI, if pulling/pushing/merging is such a problem, then should we use another tool for documentation, like a wiki editor, perhaps?  Anyway, latest changes are pushed/merged.

docs/dictionary/command/send.lcdoc
```diff
@@ -35,24 +35,24 @@ end timerIncrement
...
-Use the <send> <command> to override the normal <message path>, or to delay a <command> until a specified time. 
+Use the <send> <command> to override the normal <message path>, or to delay a <command> until after a specified time.
 
-The <message> can be either the name of a <handler> or a <LiveCode> <statement>. If the <message> is a <literal string> containing more than one <word> (for example, if it includes <parameter|parameters>, or if it is a multi-word command), it must be enclosed in <double quote|double quotes>. 
+The <messageName> can be either the name of a <handler> or a <LiveCode> <statement>. If the <messageName> is a <literal string> containing more than one <word> (for example, if it includes <parameter|parameters>, or if it is a multi-word command), it must be enclosed in <double quote|double quotes>.
 
-Any parameters are evaluated before they are passed to the <send> <command>. For example, the following statement sends the <mouseUp> <message> with 2 as the first parameter:
+Any parameters are evaluated before they are passed to the <send> <command>. For example, the following statement sends the <mouseUp> <messageName> with 2 as the first parameter:
```
livecodeali, line 48: The `<message>` here refers to the general concept of message rather than the `<messageName>` parameter, so shouldn't be changed.

livecodeali:
Heh, well making any changes to the description of functionality is much more complicated than fixing a typo, adding/modifying an example or fixing a syntax error, for example. This one has spiralled slightly from the initial change you suggested.

docs/dictionary/command/send.lcdoc
```diff
@@ -35,24 +35,24 @@ end timerIncrement
...
 
   send "mouseUp 1+1" to button "Example"
 
-When the send command is used the stack containing the target handler temporarily becomes the defaultStack. All object references in the <message> are <evaluate|evaluated> in the current context i. e. the defaultStack. Therefore references within the message that refer to "this card" or "this stack" will be referring to the card or stack where the target handler is located. 
+>*Important:* When the send command is used, the stack containing the target <handler> temporarily becomes the defaultStack. All object references in the <messageName> are <evaluate|evaluated> in the current context i. e. the defaultStack. Therefore references within the message that refer to "this card" or "this stack" will be referring to the card or stack where the target <handler> is located.
```
livecodeali, line 52: I think `the <messageName>` is not quite right here as it's just a string - it should be `the target <handler>`.

docs/dictionary/command/send.lcdoc
```diff
@@ -35,24 +35,24 @@ end timerIncrement
...
 
->*Important:*  Specifying a <time> can affect the order in which <statement|statements> are <execute|executed>. If you don't specify a <time>, the <message> is sent immediately, and any <handler> it <trigger|triggers> is completed before the rest of the current <handler> is <execute|executed>. If you use the send in time form of the <send> <command> --even if you specify a <time> of zero seconds--the current handler finishes <execute|executing> before the <message> is sent.
+>*Important:*  If you don't specify a <time>, the <messageName> is sent immediately, and any <handler> it <trigger|triggers> is completed before the rest of the current <handler> is <execute|executed>.  However, if you specify a <time>, even one of 0, the current <handler> finishes <execute|executing> before the <messageName> is sent.
```
livecodeali, line 54: I think these occurrences should not have a definite article 

```
If you don't specify a <time>, <messageName> is sent immediately
```

and 

```
the current <handler> finishes <execute|executing> before <messageName> is sent
```

make more sense to me. What do you think?

docs/dictionary/command/send.lcdoc
```diff
@@ -35,24 +35,24 @@ end timerIncrement
...
 
->*Note:*  Using the <send> <command> is slower than directly <execute|executing> the <command|commands> using the normal <message path>. For best efficiency, use the <send> <command> only when you want to delay the <message> or when the <handler> you want to execute is not in the <message path>.
+>*Note:*  Using the <send> <command> is slower than directly <execute|executing> the <command|commands> using the normal <message path>. For best efficiency, use the <send> <command> only when you want to delay the <messageName> or when the <handler> you want to execute is not in the <message path>.
```
livecodeali, line 56: I think this should also either be `the <message>` or `<messageName>` without the definite article.

livecodeali:
@macMikey There are a few issues with the pull request at the moment:
1) You seem to have added the IDE submodule to it
2) It should be based on and targeted at `community-docs` rather than `develop`
3) Some of your commit messages address me, rather than describing the change you have made.

I suggest copying your modified file somewhere, creating a new branch based on community-docs and copying it back, and submitting a new pull request.

peter-b:
@macMikey You said on the mailing list that you were not going to persevere with getting these changes merged, so I'm going to close this pull request for now.

If you'd like to get involved again at some point in the future, please reopen it or create a new one!

Thanks 😃

PR #3883, opened 2016-04-07
HTMLText - fix missing links in desc

asayd:
- In two instances the property name was listed inside links as htmltext rather than HTMLText. This caused the text and the link to be omitted from the Description text. Markdown seems to be case sensitive, at least for keywords inside links.


livecodeali:
HI @asayd, could you re-submit this against the community-docs branch please!

Thanks.

asayd:
Closed. This PR was incorporated into PR #3896.

PR #3897, opened 2016-04-11
Added "text processing" Tag to this item

marksmithhfx:


marksmithhfx:
Let me know if I did this right, or screwed up. Thanks

peter-b:
@marksmithhfx It looks like you've got almost everything right here!
- you've submitted the change to the correct branch of the correct repository 
- your patch fixes exactly one thing without bringing any extra unnecessary changes along
- you've explained what you've changed in the commit message

Unfortunately, there are two issues that mean we probably can't accept the pull request:
- I don't think that this change is quite correct; the `textSize` property isn't actually related to text processing (unlike `put ... after`, `matchText`, `filter`, etc.), but it's more closely related to text _rendering_ -- which there isn't a specific tag for; I think we need to have a larger discussion about how to use tags and make it clearer which things particular tags apply to
- Please [make sure your commit messages summarize the changes you've made more precisely](http://chris.beams.io/posts/git-commit/), because it's very helpful to be able to tell just by reading your commit message that you've modified the documentation for the `textSize` property and added a tag; this helps ensure that the pull request only changes the things that you meant to change

Another easy fix that you might be able to make to practice going through the process of contributing documentation changes would be to fix the broken reference at the end of the [stop session](https://github.com/livecode/livecode/blob/community-docs/docs/dictionary/command/stop-session.lcdoc) command's dictionary entry.

``` patch
-The location where the $_SESSION array is stored is configured using the <sessionSavePath property>.
+The location where the $_SESSION array is stored is configured using the <sessionSavePath(property)>.
```

Thank you so much for being interested in helping to improve the LiveCode open source edition. It's really encouraging to see so many people increasingly pitching in with documentation (and other) improvements over the last few weeks!

marksmithhfx:
Thanks Peter. Will give it another go.

mwieder:
@peter-b <sigh> Pedantically you're right. Pragmatically I disagree. If I were looking to do things with text and wanted to see what keywords were available to me, I'd look at "text processing". It's hard enough for new users to find their way around the documentation as is, I think _not_ having this association is doing a disservice where more hints would be desirable. Text "rendering" is more of an advanced concept, and while it's technically the correct tag, I think adding both tags would be a good idea. I realize you already closed this PR, but there's my two cents.

PR #3960, opened 2016-04-20
Add an example to the encrypt command

marksmithhfx:
Attempted to add 2 examples using the documentation guidelines.

### DIFF
--- a/docs/dictionary/command/encrypt.lcdoc
+++ b/docs/dictionary/command/encrypt.lcdoc
@@ -34,4 +34,22 @@ The password and salt value are combined and scrambled to form the key and IV wh
 
 Some modes of block ciphers will pad data to be a multiple of block size. The padding method is that used by the openSSL utility and is a minimum of one byte.
 
+Examples:
+
+local myData, myPassword
+encrypt myData using ciphername with password myPassword 
+
+To obtain a list of cipher names type the following into the message box: put the cipherNames
+
+/* the following example copies a file from disk, stores it in a variable (myData), encrypts the result, 
+then copies the file back to disk. */
+
+local path2input, myData, path2output
+/* put the file paths to your input and output files into path2input and path2output */
+put URL ("binfile:" & path2input) into myData 
+encrypt myData using "aes-256-cbc" with password “@&^2fy” 
+put it into URL ("binfile:" & path2output) 
+
+Or you can write the result back to the input file if you prefer. 
+
 References: decrypt (command), cipherNames (function), sslcertificates (property)
--- a/docs/dictionary/command/encrypt.lcdoc
+++ b/docs/dictionary/command/encrypt.lcdoc
@@ -2,7 +2,7 @@ Name: encrypt
 
 Type: command
 
-Syntax: encrypt <source> using <cipher> with [password|key] <passorkey> [and salt saltvalue] [and IV <IVvalue>] [at <bitvalue> bit]
+Syntax: encrypt <source> using <cipher> with [password|key] <passorkey> [and salt <saltvalue>] [and IV <IVvalue>] [at <bitvalue> bit]
 
 Summary: Encrypt data using a cipher.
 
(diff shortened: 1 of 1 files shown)


livecodeali:
Hi Mark, thanks for the submission! There are a few issues here:

1) I'd like to keep roughly the same order of elements, so could you put the `Example:`s before the `Description:` please? 

2) A dictionary entry can have multiple examples, but each one must be after its own `Example:` - using `Examples:` will not work. In fact though, I'm not sure the first example here adds anything that is not given by the syntax itself and the second example, so I would be in favour of just having 

```
 /* the following example copies a file from disk, stores it in a variable (myData), encrypts the result, 
 then copies the file back to disk. */

 local path2input, myData, path2output
 /* put the file paths to your input and output files into path2input and path2output */
 put URL ("binfile:" & path2input) into myData 
 encrypt myData using "aes-256-cbc" with password “@&^2fy” 
 put it into URL ("binfile:" & path2output) 
```

3) The `Example:` element can only have code in it, so other text either needs to be in comment blocks or moved into the `Description:` section.

I would probably remove them as it states a few times in the description that you can use the cipherNames function to get the valid cipher names.

Actually looking at this file, there are a lot of pre-existing issues with the doc
- none of the parameters have descriptions
- none of the occurrences of the parameters in the description are marked as such with angle brackets
- none of the occurrences of encrypt, decrypt or cipherNames in the description have angle brackets, so none of them link to the relevant entries

Obviously it's fine if you just want to add an example, just thought I'd note these in case you want to make additional improvements!

marksmithhfx:
Thanks Ali. I’ll take another look and try and make some of the other improvement. I appreciate your patience: as Devin said, we are on a learning curve with this so I am sure it will take a few attempts before I get the necessary experience to do this appropriately the first time. Cheers, and thanks

Mark

On Apr 21, 2016, at 6:42 AM, Ali Lloyd <notifications@github.com<mailto:notifications@github.com>> wrote:

Hi Mark, thanks for the submission! There are a few issues here:

1) I'd like to keep roughly the same order of elements, so could you put the Example:s before the Description: please?

2) A dictionary entry can have multiple examples, but each one must be after its own Example: - using Examples: will not work. In fact though, I'm not sure the first example here adds anything that is not given by the syntax itself and the second example, so I would be in favour of just having

 /\* the following example copies a file from disk, stores it in a variable (myData), encrypts the result,
 then copies the file back to disk. */

 local path2input, myData, path2output
 /\* put the file paths to your input and output files into path2input and path2output */
 put URL ("binfile:" & path2input) into myData
 encrypt myData using "aes-256-cbc" with password “@&^2fy”
 put it into URL ("binfile:" & path2output)

3) The Example: element can only have code in it, so other text either needs to be in comment blocks or moved into the Description: section.

I would probably remove them as it states a few times in the description that you can use the cipherNames function to get the valid cipher names.

Actually looking at this file, there are a lot of pre-existing issues with the doc
-   none of the parameters have descriptions
-   none of the occurrences of the parameters in the description are marked as such with angle brackets
-   none of the occurrences of encrypt, decrypt or cipherNames in the description have angle brackets, so none of them link to the relevant entries

Obviously it's fine if you just want to add an example, just thought I'd note these in case you want to make additional improvements!

livecodeali:
@marksmithhfx This looks great in terms of the content of the changes, thanks! There are just a couple of remaining things:

1) (non-essential) It would be really helpful if you would hard wrap the lines of the description at 72 characters just to make the file more readable.

2) The commit message should be free of external context, i.e. you shouldn't reference my suggestions in it. Unfortunately this is impossible to change using the web interface, so could you submit a new pull request with all these changes in please? And then the commit message could be something like

```
[[ Community Docs ]] Encrypt command enhancements
- add encryption example 
- provide descriptions for parameters
- add < > brackets to most parameters in the description and to encrypt, decrypt and cipherNames... so they link
- fix miscellaneous typos
```

livecodeali:
If you need a copy of the doc with the changes you've made here, it's https://raw.githubusercontent.com/marksmithhfx/livecode/patch-5/docs/dictionary/command/encrypt.lcdoc

livecodeali:
Superseded by https://github.com/livecode/livecode/pull/3986

marksmithhfx:
Hi Ali, I added a pull request to update DECRYPT.

Question for you: should all documentation entries with long para’s (like description and result in some cases) have their lines wrapped at 72 chars?

If so, that is something I could do while reading through documentation entries on GIT.

I didn’t add any TAGS. Don’t know where to start with that. Any suggestions?

Thanks

M

On May 3, 2016, at 3:48 AM, Ali Lloyd <notifications@github.com<mailto:notifications@github.com>> wrote:

Closed #3960https://github.com/livecode/livecode/pull/3960.

livecodeali:
@marksmithhfx Yes, in theory all entries with long paragraphs should be wrapped, except the
Summary (I believe this needs to be a single line, although that restriction might have been lifted), 
and the Syntax and Example elements (in which new lines are preserved).

I'm not sure about tags, I would maybe go for `SSL & Encryption`, but we probably need to have a think before adding new tags about whether they encompass a reasonable number of entries etc. Plus when a new tag is added, probably everything that should have that tag should be updated in one pull request, so as to avoid the situation where a tag exists but is incomplete.

PR #3968, opened 2016-04-21
[[Community Docs ]] Cancel command fixes

asayd:
- Corrected formatting of example embedded in Description element. ( @livecodeali - are indented lines interpreted as code blocks by markdown?) Should this be moved to the Examples element?
- Linked words that refer to control structures (repeat, function) do not show up in final dictionary text. Is this related to bug #17408?
- Reviewed and updated links and references.
- Added a more complete example.
- Is Tag: objects appropriate here?

### DIFF
--- a/docs/dictionary/command/cancel.lcdoc
+++ b/docs/dictionary/command/cancel.lcdoc
@@ -4,7 +4,8 @@ Type: command
 
 Syntax: cancel <messageQueueID> 
 
-Summary: Removes a <message> that was queued with the <send> <command> and is waiting to be sent.
+Summary: Removes a <message> that was queued with the <send> <command> 
+and is waiting to be sent.
 
 Introduced: 1.0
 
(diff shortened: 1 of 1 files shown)


docs/dictionary/command/cancel.lcdoc
```diff
@@ -18,24 +19,56 @@ cancel 2298
...
+
+All <pendingMessages|pending messages> are automatically canceled when 
+the application quits.
+
+It is common to need to <cancel> a number of <message|messages> when leaving a <card>
+or <stack>, if the <message|messages> only pertain to that card or stack. For
+example, you might have queued a number of <message|messages> that create an
+animated display on the current <card>, and need to <cancel> them when the
+user goes to another <card>. The best solution in a case like this is to
+place each messageQueueID in a <script local variable>, a <global> <variable>,
```
livecodeali, line 57: This messageQueueID should have angle brackets around it so that it gets highlighted as a parameter

livecodeali:
To answer your questions:
- Yes, indented lines are interpreted as code blocks, and I think it's fine to keep that code in the description. 
- Yes, I think the non-linking of repeat and function is related to 17408 - it will be fixed in the next release
- No, I don't think objects is an appropriate tag, it should probably be something like messages I guess

livecodeali:
Would you mind re-doing this and removing the questions from the commit message? It should just state what changes have been made.

asayd:
Closing this PR. Will redo with new commit.

PR #3985, opened 2016-04-29
[[ Bug 17516 ]] Ensure popup code checks MCmousestackptr

livecodeali:


peter-b:
Please add a regression test!

livecodeali:
Bumping this from 8.0.1-rc-1 as it needs further investigation.

PR #3996, opened 2016-05-09
Brush, spray, and eraser brushes were not initialized properly

mwieder:
The tools palette in LC8 attempts to assign some very large numbers (no doubt converted from negatives) to the spray, brush, and eraser properties, and these numbers do not correspond to valid images. That causes an error which is not normally displayed to the users, since it happens in a system stack. This patch fixes some suspicious-looking math in the MCInterfaceSetBrush function and also vectors setting the brush values in MCGraphicsExecResetPaint through MCInterfaceSetBrush so that the values get validated on the way in.

For extra credit, I fixed the default value of the spray brush, which was getting set to 31 instead of the documented value of 34. Something else must be fixing the value later on, because retrieving the value shows 34 anyway.

This fixes http://quality.livecode.com/show_bug.cgi?id=17553. It's working locally for me, so I'm opening this up for discussion.

### DIFF
--- a/engine/src/exec-interface2.cpp
+++ b/engine/src/exec-interface2.cpp
@@ -1991,7 +1991,8 @@ void MCInterfaceSetRelayerGroupedControls(MCExecContext& ctxt, bool p_value)
 void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_value)
 {
 	uint4 t_newbrush = p_value;
-	if (t_newbrush < PI_PATTERNS)
+	// MDW-2016-05-06 [[ bugfix_17553 ]] safer to compare against PI_BRUSHES than PI_PATTERNS
+	if (t_newbrush < PI_BRUSHES)
 		t_newbrush += PI_BRUSHES;
 
 	// MW-2009-02-02: [[ Improved image search ]]
--- a/engine/src/exec-interface2.cpp
+++ b/engine/src/exec-interface2.cpp
@@ -2031,10 +2031,7 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_v
 void MCInterfaceGetBrush(MCExecContext& ctxt, uinteger_t& r_value)
 {
 	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
-	if (MCbrush < PI_PATTERNS)
-		r_value = MCbrush > PI_BRUSHES ? MCbrush - PI_BRUSHES : MCbrush;
-	else
-		r_value = MCbrush;
+	r_value = MCbrush > PI_BRUSHES && MCbrush < PI_PATTERNS ? MCbrush - PI_BRUSHES : MCbrush;
 }
 
 void MCInterfaceSetBrush(MCExecContext& ctxt, uinteger_t p_value)
@@ -2045,10 +2042,7 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, uinteger_t p_value)
 void MCInterfaceGetEraser(MCExecContext& ctxt, uinteger_t& r_value)
 {
 	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
-	if (MCeraser < PI_PATTERNS)
-		r_value = MCeraser > PI_BRUSHES ? MCeraser - PI_BRUSHES : MCeraser;
-	else
-		r_value = MCeraser;
+	r_value = MCeraser > PI_BRUSHES && MCeraser < PI_PATTERNS ? MCeraser - PI_BRUSHES : MCeraser;
 }
 
 void MCInterfaceSetEraser(MCExecContext& ctxt, uinteger_t p_value)
@@ -2059,10 +2053,7 @@ void MCInterfaceSetEraser(MCExecContext& ctxt, uinteger_t p_value)
 void MCInterfaceGetSpray(MCExecContext& ctxt, uinteger_t& r_value)
 {
 	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
-	if (MCspray < PI_PATTERNS)
-		r_value = MCspray > PI_BRUSHES ? MCspray - PI_BRUSHES : MCspray;
-	else
-		r_value = MCspray;
+	r_value = MCspray > PI_BRUSHES && MCspray < PI_PATTERNS ? MCspray - PI_BRUSHES : MCspray;
 }
 
 void MCInterfaceSetSpray(MCExecContext& ctxt, uinteger_t p_value)
--- a/engine/src/exec-interface2.cpp
+++ b/engine/src/exec-interface2.cpp
@@ -1993,7 +1993,7 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_v
 	uint4 t_newbrush = p_value;
 	// MDW-2016-05-21 [[ bugfix_17553 ]] map low brush values up to 101..135 range
 	// if less than 36, map to 101..135
-	if (t_newbrush < (PI_PATTERNS-PI_BRUSHES))
+	if (t_newbrush <= (PI_PATTERNS-PI_BRUSHES))
 		t_newbrush += PI_BRUSHES;
 
 	// MW-2009-02-02: [[ Improved image search ]]
@@ -2031,7 +2031,7 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_v
 void MCInterfaceGetBrush(MCExecContext& ctxt, uinteger_t& r_value)
 {
 	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
-	r_value = MCbrush > PI_BRUSHES && MCbrush < PI_PATTERNS ? MCbrush - PI_BRUSHES : MCbrush;
+	r_value = MCbrush > PI_BRUSHES && MCbrush <= PI_PATTERNS ? MCbrush - PI_BRUSHES : MCbrush;
 }
 
 void MCInterfaceSetBrush(MCExecContext& ctxt, uinteger_t p_value)
@@ -2042,7 +2042,7 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, uinteger_t p_value)
 void MCInterfaceGetEraser(MCExecContext& ctxt, uinteger_t& r_value)
 {
 	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
-	r_value = MCeraser > PI_BRUSHES && MCeraser < PI_PATTERNS ? MCeraser - PI_BRUSHES : MCeraser;
+	r_value = MCeraser > PI_BRUSHES && MCeraser <= PI_PATTERNS ? MCeraser - PI_BRUSHES : MCeraser;
 }
 
 void MCInterfaceSetEraser(MCExecContext& ctxt, uinteger_t p_value)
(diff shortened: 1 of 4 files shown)


montegoulding:
@mwieder I'm wondering if we need to note why it's important the condition is different on setting and getting. Also the PR will need a note and probably tests on reset paint.

peter-b:
@mwieder Thank you very much for fixing this!

Could you please add a test? The documentation for how to add test is here: https://github.com/livecode/livecode/blob/develop/docs/testing.md

I think this will also need a release note, please.

montegoulding:
@peter-b We might need to provide guidance on the tests because they would require loading the brush patterns

peter-b:
@montegoulding I believe some of the existing LCS level tests already use resource files.  For example, the [clipboard tests](https://github.com/livecode/livecode/blob/develop/tests/lcs/core/engine/clipboard.livecodescript) use some .gif resource files.

montegoulding:
@peter-b I think it would be a case of loading revicons.rev from the ide. Are tests being run with submodules initialised & updated?

peter-b:
@montegoulding Yes, the tests should always be run with the IDE submodule present. But maybe a test for this would be better located in the IDE tests directory?

montegoulding:
A bit of a grey area perhaps because it is testing the engine...

peter-b:
@montegoulding Maybe if revicons.rev is required for the engine to work properly it should be located in the engine repository rather than the IDE repository.

mwieder:
I am at any rate out of town until next week, so I won't be adding anything until then.

mwieder:
Right. I'm back in town again. I do think proper tests for this should be at a scripting level rather than at the engine level. I'm agnostic as to whether revicons.rev ends up in the engine module or the ide module, aside from my dislike of submodules in general. So I'll await further guidance as to where to insert the tests.

montegoulding:
@mwieder The issue with revicons.rev is because of the change in functionality where `reset paint` now validates the default patterns when previously it didn't and the validation requires them to be available. I can't really see any reason why revisions.rev needs to be moved as its path relative to the tests should be constant. It would be good I think to test both with revicons unavailable (which should throw an error) and with it available (which should work fine). Perhaps `reset paint` tests should be run after getter and setter tests for the individual paint tools considering that the way to test `reset paint` has worked correctly would be to get `the spray` etc... 

Something like:
- test individual paint tools sans revicons
- test reset paint sans revicons
- load revicons
- test indvidual paint tools round trip correctly to something other than the default
- test individual paint tools round trip correctly set to < 100
- test individual paint tools round trip correctly set to 100-136
- test individual paint tools round trip correctly > 136
- test reset paint resets to the documented values

Thanks again!

mwieder:
Having now spent several hours writing tests for these, it's apparent that things are Not As They Seem. The revicons stack has images with ids 101 through 132. I don't know where 133, 134, and 135 live. Writing tests for brush ids outside these limits will fail not because of id mapping, but because the stacks that hold the images haven't been loaded, and that's a whole different test. In addition, I think I'd like to look at the test framework as a whole as a separate project... what I'd like is to print the message attached to the failing test, and I haven't found how to do that.

Also, tests that run properly when I run them from the message box fail from the test framework:

on TestRemappedBrush
    local tBrush

```
open stack "../ide/Toolset/palettes/revicons.rev"
    -- this test fails
repeat with tBrush = 101 to 102
    set the brush to tBrush
    TestAssert "100 < brush < 136 subtracts 100", the brush is (tBrush-100)
end repeat
```

--  set the brush to 101 -- this test works by itself
--  TestAssert "brush = 101 unchanged", the brush is 1
end TestRemappedBrush

So far I'm able to provide trivial tests for setting brushes and reset paint, but this mixed engine/ide test environment is getting to be a pain.

montegoulding:
@mwieder thanks for your efforts! Using the resolve image command we should be able to determine where those images reside. I'll look at that today for you. I'm not sure why they wouldn't be together.

montegoulding:
@mwieder resolve image id indicates that those ids 133-135 are in revicons so I guess we need to work out why you are having issues with them.

engine/src/exec-interface2.cpp
```diff
@@ -2028,7 +2030,11 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_v
 
 void MCInterfaceGetBrush(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	r_value = MCbrush < PI_PATTERNS ? MCbrush - PI_BRUSHES : MCbrush;
+	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
+	if (MCbrush < PI_PATTERNS)
```
montegoulding: We can get rid of the ternary operators here @mwieder by making the condition && MCbrush > PI_BRUSHES

tests/lcs/core/graphics/graphics.livecodescript
```diff
@@ -90,4 +90,120 @@ TestAssert "test", not ("4,9" is not within "-5,-10,5,10")
...
+	catch e
+		TestDiagnostic e
+	end try
+	return the brush
+end SetBrushZero
+
+on TestBrush
+	open stack "../ide/Toolset/palettes/revicons.rev"
+	set the brush to 30
+	TestAssert "brush > 35 unchanged", the brush is 30
```
montegoulding: @mwieder I think you mean < here

tests/lcs/core/graphics/graphics.livecodescript
```diff
@@ -90,4 +90,120 @@ TestAssert "test", not ("4,9" is not within "-5,-10,5,10")
...
+end TestBrush
+
+on TestRemappedBrush
+	local tBrush
+	
+	open stack "../ide/Toolset/palettes/revicons.rev"
+#	repeat with tBrush = 101 to 102
```
montegoulding: @mwieder we should write the tests based on how we know the engine should perform and work out why they are failing if they fail. Perhaps the first test to run once we load revicons is on the existence of image id 101-136?

tests/lcs/core/graphics/graphics.livecodescript
```diff
@@ -90,4 +90,120 @@ TestAssert "test", not ("4,9" is not within "-5,-10,5,10")
...
+-- brush < 101 = brush
+-- 101 < brush < 135 = brush - 100
+-- brush > 136 = brush
+
+on TestBrushWithNoIcons
+	TestAssert "set brush with revicons stack not loaded", SetBrushNoIcons() is not 2
+end TestBrushWithNoIcons
+
+on TestBadBrush100
+	open stack "../ide/Toolset/palettes/revicons.rev"
```
montegoulding: @mwieder we should use `TestGetIDERepositoryPath` here

montegoulding: We should also delete the stack at the end of the test handler as some tests depend on it either being loaded or not being loaded

montegoulding:
@mwieder I really appreciate your efforts. I know this has turned into more than you expected! I'm going to do something horrible and suggest a couple more things though. I think it would be good if `TestResetPaint` tested all the changes made by `reset paint` rather than just the paint tool patterns. That means that these also need to be tested:
- the centered is false
- the filled is false
- the grid is false
- the gridSize is 8
- the lineSize is 1
- the multiple is false (unused by LC anyway)
- the multiSpace is 1 (unused by LC anyway)
- the penPattern is empty
- the brushPattern is empty
- the polySides is 4
- the roundEnds is false
- the slices is 16
- the penColor is 0,0,0
- the brushColor is 255,255,255

There is also something set called MCpattern which appears to be unused by the engine and MCmagnification looks like something which could be a global property (but isn't) for the amount an image is magnified by when its magnify property is set to true. There's no way to test either of those at this stage.

montegoulding:
Oh and I'm also going to suggest we should be testing the initial defaults at the start of the test.

engine/src/exec-interface2.cpp
mwieder: Will do. I separated things here so I could fiddle with things bit by bit.

tests/lcs/core/graphics/graphics.livecodescript
mwieder: Yep - I haven't kept the comments up to date.

tests/lcs/core/graphics/graphics.livecodescript
mwieder: Well, technically I'm not trying to test the revicons stack here. On the first card of the revicons stack I can see brushes with ids 101->132. I'm not sure where the others are hiding. And for the purpose of setting the brush I don't really care. Mapping image ids has always been broken in the sense of mapping ids that don't need to be mapped, so I'm restricting the mapped ids to the 101..135 range, which map to 1..35. This is complicated now because the existence of those ids is checked at the script level, not at the engine level.

tests/lcs/core/graphics/graphics.livecodescript
mwieder: OK - will check that out. And good point about unloading the stack - probably the TestTeardown handler is the place for that.

mwieder:
The reset paint tests should at least be easy to implement. I'm not sure what the initial defaults should be before setting things. I could probably do a bunch of TestDiagnostic statements to discover them empirically, but I'm not sure that's really the way to test them.

I'm really more concerned about why the tests fail in a repeat loop (or why else they would be failing). Possibly something stupid in my code, but the same code runs fine in the live IDE, just not in an lcs test. I'd blame the testrunner framework, but I haven't been over the code in enough detail to know.

tests/lcs/core/graphics/graphics.livecodescript
montegoulding: They appear to be on card ids 1027, 1029, 1031.

Try running this:

```
on mouseUp
   put empty
   repeat with X = 101 to 135
      resolve image id X relative to me
      put X && ">" && the long name of it & return after msg
   end repeat
end mouseUp
```

tests/lcs/core/graphics/graphics.livecodescript
montegoulding: @mwieder I think you could only use TestSetup & TestTeardown for the revicons stack if you separate this into two test scripts. One that requires revicons and one that doesn't.

montegoulding:
Logically `reset paint` should reset things to the engine defaults. If it doesn't it should probably be called something else ;-)

Yes it is odd that your repeat loops are failing. I'm new to the test library myself so I'll investigate to see if I can work out the issue.

montegoulding:
I think the descriptions in your repeat loops are bumping into this:

```
if "0123456789" contains codepoint 1 of pDescription then
   throw "Bad test description '" & pDescription & "': starts with digit"
end if
```

mwieder:
Jesus H Christ! Yes, that fixed it. I saw that code earlier this morning and ignored it because I didn't need to worry about it. Why does that trap even exist? Who cares if a description starts with a number?

Anyway, some interesting discoveries about 'reset paint':
The penPattern is actually 136 initially and after reset paint.
So is the brushPattern.
The grid is initially true, reset paint sets it to false.
The gridsize is initially 4, reset paint sets it to 8.
The lineSize is initially 0, reset paint sets it to 1.
Not sure what to make of that.

montegoulding:
@mwieder Thanks it might be good for higher powers to comment here @peter-b or @livecodemark

From the reset paint docs which would indicate that any difference between the engine defaults and the state reset paint command would be a bug:

> Use the reset paint command as a shortcut to clean up paint properties by resetting them to their default settings.

I think there's a docs bugs too because it mentions `the pattern` which is not a property as far as I can tell.. just a variable floating around doing nothing.

The penPattern and brushPattern being freed is sort of documented by the penColor and brushColor being set. You can test the behavior by testing the penColor and brushColor because they are one of these color/pattern either or properties.

However I did find something important out! Setting the penPattern and brushPattern was broken in the refactor. If you are keen to fix that too that would be wonderful otherwise we should open a report on it. Actually it is very close to this http://quality.livecode.com/show_bug.cgi?id=10468 but I believe it was introduced in the refactor and impacts both pen and brush.

mwieder:
Well, by the time the IDE is up and ready to use it appears the reset paint command has already been issued, because the defaults from the messagebox are different from the defaults from the raw engine, so I'm not very concerned about the differences.

Personally, I've never liked the mapping of brushes/patterns to the actual ids. If we want to use an image id as a brush, why not use the id instead of sometimes an index into a small set, and other times the actual id? I'd prefer to have reset paint set the brush to 108, the eraser to 102, the spray to 134 and not have to deal with the weird indexing. I'm just afraid of what might break in the IDE.

I didn't want to fiddle with the docs too much except for fixing the erroneous spray entry, but yes, the 'pattern' instead of 'brush' bothers me in all three locations.

montegoulding:
@mwieder I've just submitted a [PR](https://github.com/livecode/livecode/pull/4060) to cover the pattern number behavior issues I found. I've also removed the MCpattern var which was doing nothing.

mwieder:
The range is now 1..36 and all tests are passing.

montegoulding:
Hmm... I thought this was just about good to go @mwieder

mwieder:
I see nobody has fixed this yet, so I'll resubmit it if you want.

montegoulding:
That would be great @mwieder as it would have been a shame to duplicate what you had already done

PR #4005, opened 2016-05-10
[[Bug 16730]] Added the ability to access aspect-specific bounding

hwbehrens:
Adds the ability to access the height and width of an SVG path with
subpixel precision when maintainAspectRatio is engaged.

---

This change is [Reviewable](https://reviewable.io/reviews/livecode/livecode/4005)

### DIFF
--- a/extensions/widgets/svgpath/tests/properties.livecodescript
+++ b/extensions/widgets/svgpath/tests/properties.livecodescript
@@ -22,6 +22,7 @@ end getPathData
 
 on TestSetup
 	TestLoadExtension "com.livecode.library.widgetutils"
+	TestLoadExtension "com.livecode.library.iconSVG"
 	TestLoadExtension "com.livecode.widget.svgPath"
 	create widget "testSvgPath" as "com.livecode.widget.svgPath"
 end TestSetup
--- a/extensions/widgets/svgpath/tests/properties.livecodescript
+++ b/extensions/widgets/svgpath/tests/properties.livecodescript
@@ -1,0 +1,67 @@
+﻿script "SvgPathWidgetProperties"
+/*
+Copyright (C) 2016 LiveCode Ltd.
+
+This file is part of LiveCode.
+
+LiveCode is free software; you can redistribute it and/or modify it under
+the terms of the GNU General Public License v3 as published by the Free
+Software Foundation.
+
+LiveCode is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or
+FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+for more details.
+
+You should have received a copy of the GNU General Public License
+along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
+
+private function getPathData
+     return "M1664 698Q1664 720 1638 746L1275 1100 1361 1600Q1362 1607 1362 1620 1362 1641 1351.5 1655.5 1341 1670 1321 1670 1302 1670 1281 1658L832 1422 383 1658Q361 1670 343 1670 322 1670 311.5 1655.5 301 1641 301 1620 301 1614 303 1600L389 1100 25 746Q0 719 0 698 0 661 56 652L558 579 783 124Q802 83 832 83 862 83 881 124L1106 579 1608 652Q1664 661 1664 698Z"
+end getPathData
+
+on TestSetup
+     TestLoadExtension "com.livecode.library.widgetutils"
+     TestLoadExtension "com.livecode.widget.svgPath"
+     create widget "testSvgPath" as "com.livecode.widget.svgPath"
+end TestSetup
+
+on TestTeardown
+     delete widget "testSvgPath"
+end TestTeardown
+
+on TestCreateHeader
+     TestAssert "created svgPath exists", there is a widget "testSvgPath"
+end TestCreateHeader
+
+on TestSetPath
+     set the iconPath of widget "testSvgPath" to getPathData()
+     TestAssert "iconPath set to default 'star' value", the iconPath of widget "testSvgPath" is getPathData()
+end TestSetPath
+
+on TestMaintainAspectRatioBasic
+     set the maintainAspectRatio of widget "testSvgPath" to true
+     TestAssert "maintainAspectRatio set to true", the maintainAspectRatio of widget "testSvgPath" is true
+     
+     set the maintainAspectRatio of widget "testSvgPath" to false
+     TestAssert "maintainAspectRatio set to false", the maintainAspectRatio of widget "testSvgPath" is false
+end TestMaintainAspectRatioBasic
+
+on TestSetRect
+     set the rect of widget "testSvgPath" to "0,0,100,100"
+     TestAssert "the rect was set to 0,0,100,100", the rect of widget "testSvgPath" is "0,0,100,100"
+end TestSetRect
+
+on TestScaledSizing
+     set the iconPath of widget "testSvgPath" to getPathData()
+     set the maintainAspectRatio of widget "testSvgPath" to true
+     set the rect of widget "testSvgPath" to "0,0,100,100"
+     TestAssert "scaledHeight with maintainAspectRatio = true", the scaledHeight of widget "testSvgPath" is 95.372589
+
+     set the rect of widget "testSvgPath" to "0,0,150,100"
+     TestAssert "scaledWidth with maintainAspectRatio = true", the scaledWidth of widget "testSvgPath" is 104.851929
+
+     set the maintainAspectRatio of widget "testSvgPath" to false
+     TestAssert "scaledHeight with maintainAspectRatio = false", the scaledHeight of widget "testSvgPath" is 100
+     TestAssert "scaledWidth with maintainAspectRatio = false", the scaledWidth of widget "testSvgPath" is 150
+end TestScaledSizing
--- a/extensions/widgets/svgpath/tests/properties.livecodescript
+++ b/extensions/widgets/svgpath/tests/properties.livecodescript
@@ -58,11 +58,8 @@ on TestScaledSizing
 	set the maintainAspectRatio of widget "testSvgPath" to true
 
 	set the rect of widget "testSvgPath" to "0,0,90,100"
-	TestAssert "scaledWidth with maintainAspectRatio = true", (the scaledWidth of widget "testSvgPath" is 90 and the scaledHeight of widget "testSvgPath" is 90)
-
-	set the rect of widget "testSvgPath" to "0,0,100,90"
-	TestAssert "scaledWidth with maintainAspectRatio = true", (the scaledWidth of widget "testSvgPath" is 90 and the scaledHeight of widget "testSvgPath" is 90)
+	TestAssert "scaled sizing with maintainAspectRatio = true", (round(the scaledWidth of widget "testSvgPath") is 90 and round(the scaledHeight of widget "testSvgPath") is 90)
 
 	set the maintainAspectRatio of widget "testSvgPath" to false
-	TestAssert "scaledHeight with maintainAspectRatio = false", (the scaledHeight of widget "testSvgPath" is not the scaledWidth of widget "testSvgPath")
+	TestAssert "scaled sizing with maintainAspectRatio = false", (round(the scaledHeight of widget "testSvgPath") is not round(the scaledWidth of widget "testSvgPath"))
 end TestScaledSizing
--- a/extensions/widgets/svgpath/tests/properties.livecodescript
+++ b/extensions/widgets/svgpath/tests/properties.livecodescript
@@ -17,14 +17,13 @@ You should have received a copy of the GNU General Public License
 along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
 
 private function getPathData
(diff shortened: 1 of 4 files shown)


livecode-vulcan:
📝 Hi @hwbehrens, I haven't been able to verify that you've signed our [LiveCode Contributor's Agreement](https://livecode.com/account/developer/contribute).

Please see the [information for contributors](https://github.com/runrev/livecode/blob/develop/CONTRIBUTING.md) for more information.

hwbehrens:
I linked my Github with my Contributor Agreement. Hopefully Vulcan will check again?

peter-b:
@hwbehrens Thank you so much for your patch! I have just a few comments:
- Because the change adds functionality in a backwards-compatible way, I recommend updating the version to "2.1.0" (see https://semver.org/)
- Could you please add a release note to the `extensions/widgets/svgpath/notes` directory?

You might also want to add `Related: aspectHeight(property), maintainAspectRatio(property)` to the documentation of `aspectWidth` (etcetera for `aspectHeight`).

I'm not sure whether the naming and semantics of this change are ideal. I wonder if just the straightforward `formattedWidth`/`formattedHeight` might actually be clearer, because it would act in the same way as for existing LiveCode controls such as the field control and the image control.

@livecodeali What do you think?

livecodeali:
@peter-b It's a tricky one. I think it's different from the image control's `formattedWidth`, which returns the  intrinsic width of the image. I seem to remember @runrevmark having a strong opinion about this before ;-)

livecodeali:
OTOH if I wanted to get one of these properties and had to guess without looking at the docs, I would for sure guess `formattedWidth` and `formattedHeight`.

hwbehrens:
What do you think about `pathHeight` and `pathWidth` with a synonym of `formattedHeight` and `formattedWidth`? The former would give a more accurate description, while the latter would provide a more intuitive interface.

I'll change it to `formattedWidth` and `formattedHeight` at minimum, as it seems to be the current consensus unless I heard otherwise.

peter-b:
@hwbehrens I would suggest that we avoid adding synonyms unless it is very necessary!

Please can you explain a bit more why you introduced the relationship with the `maintainAspectRatio` property?  Maybe I don't properly understand the problem that it solves.

hwbehrens:
I only wanted to clarify that when maintainAspectRatio is off, then the formattedWidth and the formattedHeight is equal to the width and height of the widget itself. I can remove that if you don't think it's required.

peter-b:
@hwbehrens Do you think it makes sense to have access to the "true" `formattedWidth` and `formattedHeight` even if the `maintainAspectRatio` is disabled?

hwbehrens:
That would be great, but since I was using `constrainPathToRect()` to calculate the bounding box of the path, I didn't necessarily want to calculate it every time (vs storing the value during the paint). If you think the performance implications are minimal, I can adjust that; I just really had no clue as to the performance characteristics of the constrain function, so I figured I'd be conservative with it.

peter-b:
@hwbehrens What about something like this...
- Add `mFormattedWidth` and `mFormattedHeight` member variables
- Make sure that `mPath` is only ever modified by the `setPath()` handler
- Change `setPath()` so that it updates `mFormattedWidth` and `mFormattedHeight`
- Link the `formattedHeight` property to the `mFormattedHeight` variable (etc.)

That way you only ever have to calculated the formatted sizes the minimum number of times (i.e. when the path data is modified).

What do you think?

Edit: fixed a typo

hwbehrens:
@peter-b Is `setPath()` called when the widget is resized? Since the `formattedWidth` and `formattedHeight` change depending on the size of the widget (unlike a regular image), we'll need to recalculate them in that case. Is there an LCB hook for catching resizes, like `public handler OnResize()` or something similar?

peter-b:
@hwbehrens Oh! I think I have completely misunderstood the point of this feature. I'm very sorry!

I thought that you wanted a feature like the image control has -- you can set the image to any size, but the `formattedHeight` is always the "true" height of the image in pixels.

Is the point of this feature to make it possible to get the subpixel height of the SVG even when the widget is set to a whole number of pixels high (and similarly for the width)?  Would `scaledWidth` and `scaledHeight` or something describe that better?

hwbehrens:
@peter-b Yes, that's exactly right, since (as far as I'm aware) an SVG path doesn't really have a "native" size. I'm fine with `scaledWidth` and `scaledHeight` as naming. Should I preserve the link to `maintainAspectRatio` in the documentation?

peter-b:
@hwbehrens I think it makes sense to preserve the documentation link to `maintainAspectRatio` because the value does depend on it!

peter-b:
@hwbehrens Based on your description I think the current implementation seems quite right. Probably all that is needed is to update the names of the properties and then to add some tests and a release note.

hwbehrens:
@peter-b I incorporated the changes and I added a release note, but I'll need a bit of guidance with how to add a test case to a widget. I'll also start poking around the documentation to see if I can figure out what I need to do there as well.

peter-b:
@hwbehrens Since this change introduces a new feature, could you please add a more detailed description of the feature to the release note?

A good example might be [the release note for adding the fill rule property](https://github.com/livecode/livecode/blob/develop/extensions/widgets/svgpath/notes/feature-svgWidgetFillRuleProperty.md)

In terms of adding a test case, [here's an example from the header bar widget](https://github.com/livecode/livecode/blob/develop/extensions/widgets/header/tests/properties.livecodescript) and [here's our documentation about testing](https://github.com/livecode/livecode/blob/develop/docs/testing.md).

Let me know if you run into any problems or anything's unclear...

runrevmark:
If I understand the thread here correctly then this all sounds reasonable...

The width and height of the bounding box of the SVG path which it has when it has been just constructed from the path string should be considered the 'intrinsic size' of the path - 'intrinsic size' in LiveCode is generally fetched using formattedWidth / formattedHeight so I'd suggest leaving those property names for that precise meaning.

The scaledWidth / scaledHeight seem like perfectly reasonable names for the properties which describe the actual effective width / height of the path - which obviously are only different from width/height if maintainAspectRatio is true.

Thanks for the contribution @hwbehrens!

hwbehrens:
I added some tests and I expanded the release notes into a new feature entry. Shamefully, I haven't had a chance to run the tests on my machine yet, but I'll do that shortly and write a followup if there are any issues.

hwbehrens:
Sorry guys, I guess I need some help with Travis. Can you take a look at my tests and see what I did? They all seem to pass locally. I can try collapsing everything into the test handlers themselves instead of using `TestSetup` and `TestTeardown`, but I can't imagine that that's the problem.

extensions/widgets/svgpath/tests/properties.livecodescript
```diff
@@ -0,0 +1,68 @@
...
+along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
+
+private function getPathData
+	return "M1664 698Q1664 720 1638 746L1275 1100 1361 1600Q1362 1607 1362 1620 1362 1641 1351.5 1655.5 1341 1670 1321 1670 1302 1670 1281 1658L832 1422 383 1658Q361 1670 343 1670 322 1670 311.5 1655.5 301 1641 301 1620 301 1614 303 1600L389 1100 25 746Q0 719 0 698 0 661 56 652L558 579 783 124Q802 83 832 83 862 83 881 124L1106 579 1608 652Q1664 661 1664 698Z"
+end getPathData
+
+on TestSetup
+	TestLoadExtension "com.livecode.library.widgetutils"
+	TestLoadExtension "com.livecode.library.iconSVG"
+	TestLoadExtension "com.livecode.widget.svgPath"
```
peter-b: @hwbehrens The `TestLoadExtension` command currently is case-sensitive on Linux -- the argument has to match the name specified in the widget's LCB `widget` declaration exactly.  You need to change `svgPath` to `svgpath`.

extensions/widgets/svgpath/tests/properties.livecodescript
```diff
@@ -0,0 +1,68 @@
...
+LiveCode is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or
+FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+for more details.
+
+You should have received a copy of the GNU General Public License
+along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
+
+private function getPathData
+	return "M1664 698Q1664 720 1638 746L1275 1100 1361 1600Q1362 1607 1362 1620 1362 1641 1351.5 1655.5 1341 1670 1321 1670 1302 1670 1281 1658L832 1422 383 1658Q361 1670 343 1670 322 1670 311.5 1655.5 301 1641 301 1620 301 1614 303 1600L389 1100 25 746Q0 719 0 698 0 661 56 652L558 579 783 124Q802 83 832 83 862 83 881 124L1106 579 1608 652Q1664 661 1664 698Z"
```
peter-b: I think it would be a good idea to use a really really simple path here -- such as a unit square -- which cannot possibly have any differences between operating systems. At the moment the `TestScaledSizing` test is failing on Travis because there seem to be some subtle differences in how the bounds of the path is being calculated. Don't forget you can change the size of the widget too.

peter-b:
@hwbehrens I've had a look on my Linux system and I think I've found the problem.  Can you verify?

hwbehrens:
I think you're probably right; unfortunately, I am out of the office today,
so I will be unable to submit the correct patch today. You can expect one
tomorrow.

hwbehrens:
Okay, I think I'm stuck again. I've gotten it down to just the last test failing. I thought it might be related to the fact that I was comparing a float to an integer, but adding a round didn't help matters. `rect` and `maintainAspectRatio` were already tested, so I assume those are working correctly.

Any ideas?

peter-b:
@hwbehrens I've run your code on Linux and it looks like fetching `scaledHeight` and `scaledWidth` properties are returning the empty string, which is really weird! I have no idea what's going on there.

livecodeali:
I suspect the problem here is that the variables are being set in the OnPaint handler, which is never actually called when running headless.

Also, the `OnLoad` handler needs to have

```
    if "scaledWidth" is among the keys of pProperties then
        put pProperties["scaledWidth"] into mWidth
    end if
    if "scaledHeight" is among the keys of pProperties then
        put pProperties["scaledHeight"] into mHeight
    end if
```

to prevent errors when loading old svg widgets.
In fact, these properties do not need to be saved, they can just be recalculated  when the widget loads I think.

peter-b:
@livecodeali I agree -- it doesn't make sense for these properties to be persisted with `OnSave()` and `OnLoad()`, because they can be automatically regenerated from other data that is saved.

hwbehrens:
@peter-b @livecodeali I'll go ahead and remove the stored values. Rather than `OnPaint`, maybe I could use `onCreate`, `OnOpen`, and `OnGeometryChanged`?

hwbehrens:
Okay, I added some changes to where the calculations take place, as well as some tweaks to tests. Everything is passing now, hurray! This whole process has really underscored in my mind the importance of unit tests.

peter-b:
Nearly there! The only problem I can see now is that there's a minor typo in the release note! Thank  you for persevering with this, and I'm sorry it's taken so long!

<details><summary>

<i>Previously, hwbehrens (Hans Behrens) wrote…</i></summary>



> Okay, I added some changes to where the calculations take place, as well as some tweaks to tests. Everything is passing now, hurray! This whole process has really underscored in my mind the importance of unit tests.
> </details>

---

Reviewed 2 of 2 files at r6.
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

_[extensions/widgets/svgpath/notes/feature-svgWidgetScaledSizeProperty.md, line 5 [r6]](https://reviewable.io:443/reviews/livecode/livecode/4005#-KI1thbEAoJpfDPUGsNs:-KI1thbEAoJpfDPUGsNt:177880774) ([raw file](https://github.com/livecode/livecode/blob/e09d7a705c421f27df3ea0f628e0dd1b230505bb/extensions/widgets/svgpath/notes/feature-svgWidgetScaledSizeProperty.md#L5)):_

> ``` Markdown
> * New **scaledWidth** and **scaledHeight** properties has been added.  These are read-only values
>   that expose the effective size of the rendered SVG path independent of the widget size. When
>   maintainAspectRatio is false, then these values are equal to the width and height of the widget.
> ```

I think "has been added" should be "have been added".

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/livecode/livecode/4005#-:-KI1txTBjgRzoqkcI7OF:153095730)_

hwbehrens:
@peter-b Whoops! Thanks for pointing that out; I've pushed a new typo fix for that.

peter-b:
Yay, looking good!

@livecode-vulcan review ok 5386fc8

peter-b:
@hwbehrens It looks like @livecode-vulcan has got confused by the very large number of comments on this PR. There will be a brief delay while I fix it.

livecode-vulcan:
💙 review by @peter-b ok 5386fc891dadd1c3fa0999d5c5a0e1de1616bee9

livecode-vulcan:
😎 test success 5386fc891dadd1c3fa0999d5c5a0e1de1616bee9
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/2522)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/2532)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/2517)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/2533)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/2539)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/2543)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/2555)

peter-b:
@hwbehrens Thanks for this change! Sorry it took so long to get it in!

PR #4014, opened 2016-05-12
[Bug 17573] Separate externals to specific platforms

peter-b:
Ensure that lists of enabled externals for standalone deployment are
maintained on a per-platform basis.  This makes it possible to include
specific externals only on platforms that actually need them or are
supported by them.

### DIFF
--- a/docs/notes/bugfix-17573.md
+++ b/docs/notes/bugfix-17573.md
@@ -1,0 +1,1 @@
+# Separate externals to specific platforms when saving standalone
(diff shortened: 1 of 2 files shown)


peter-b:
I'm not sure if this is quite correct -- I wonder if there should be some mechanism for bringing forward older standalone settings which don't have the externals broken down per-platform.

/cc @livecodeali

livecodeali:
I'm also not sure this is correct - I need to have a think about it.

livecodeali:
Ok, this PR https://github.com/livecode/livecode/pull/4020 fixes this issue in a easier way

PR #4021, opened 2016-05-17
[[ Bug 17515 ]] Add support for custom entitlements for iOS

montegoulding:
Custom entitlements can now be added to an iOS app by including one or
more `.xcent` files in the copy files section of the standalone builder
containing an XML snippet of key/value pairs. For example, if you wanted
to add the entitlement for HomeKit to your app you might create a file
named `HomeKit.xcent` with the following content:

```
<key>com.apple.developer.homekit</key>
<true/>
```


ide-support/revsaveasiosstandalone.livecodescript
```diff
@@ -998,6 +995,12 @@ private command revCopyMobileFiles pFiles, pBaseFolder, pAppBundle, pAppTarget,
          next repeat
       end if
       
+      -- If the file is an entitlement then handle appropriately
+      if tSource ends with ".xcent" then
+         put url ("binfile:" & tSource) after rCustomEntitlements
```
peter-b, line 1000: Aren't entitlement files text files? Are you sure it makes sense to be loading them as binary files? If so, do you think it might be a good idea to add a comment to the source code explaining it?

montegoulding, line 1000: Hmm... all the plists being loaded in this script are treated as binary and have been for a long time. They are utf-8 so would need to be read and decoded and then re-encoded to ensure they weren't treated as MacRoman and have their line endings changed. I would suggest leaving as is and adding to the backlog if it's necessary which I'm not convinced it is.

peter-b, line 1000: Makes sense! If `textDecode` could validate then I'd suggest doing that for a sanity check but since it doesn't validate I think the "just treat as binary data" approach makes good sense.

peter-b:
@livecode-vulcan review ok 4c15b80

livecode-vulcan:
💙 review by @peter-b ok 4c15b808cd741d10b74857d7dc927bda38a546b6

PR #4034, opened 2016-05-19
[[ Community Docs ]] Dictionary Editing Checklist

asayd:
I have compiled a checklist style guide to editing dictionary entries, aimed at folks who are new to the editing process.

---

[Rendered version](https://github.com/asayd/livecode/blob/asayd-docs_editing_checklist/docs/dictionary_editing_checklist.md)


asayd:
@livecodeali @peter-b 
I've put this document together to capture some of the lessons learned as I've been engaged in making dictionary updates. My hope is to make it a practical checklist for newbies, which we could publicize as another resource for folks who want to help out with docs. I may also try to develop a validator stack based on these points.

I would appreciate feedback. Specifically,
- Is it just too derivative and repetitive of @livecodeali 's earlier articles?
- Are there inaccuracies that I should fix?
- Are there important things I've missed?
  Once I'm happy with it, and if you choose to commit it to the branch I would like to follow up with a blog post to hopefully restoke the fire for folks who want to help in the dictionary improvement effort.

Thanks,

Devin

docs/dictionary_editing_checklist.md
```diff
@@ -0,0 +1,156 @@
...
+	- Use of parameter names in other elements, such as Description, should be enclosed in &lt;angle brackets&gt;.
+  
+	- Look for conflicts between parameter names and referenced/linked vocabulary; e.g. if a Syntax element uses *property* as a parameter name, it may be best to change it to something like *propertyName*.
+	
+	
+* Value or Returns
+  - These two elements are mutually exclusive; only one should appear in any document.
+  - Use a Returns element for the return value of a function; e.g.:
+  
+      >Return: The abs function returns a positive number.
```
livecodeali: Should be `Returns:` here

montegoulding:
@asayd It would be good to link to this from [Contributing to LiveCode Documentation](https://github.com/livecode/livecode/blob/community-docs/docs/contributing_to_docs.md)

asayd:
@montegoulding @livecodeali That would be great. I'm supposed to do a workshop at the dev conference about documentation contributions. I would base it on this checklist. It might also give you some ideas for the validation process. How's that going btw?

montegoulding:
@asayd The tests are in the develop branch. They cover some basic docs requirements and example compilation. There's quite a bit more we could test for but unfortunately if we enable them we will need to spend a few days sorting out what they find. Probably what would be best is if the test only ran on modified and new files so that if someone touches a dictionary entry it is then fully validated requiring that person to sort out all the problems in the document. I need to discuss with the team whether this might be a useful general feature of the test library or is only really useful for this specific use case. I suspect that something like a note or more general markdown file linter would want to use a similar mechanism. It may also be useful for testing script parsing errors in any script only stacks although in that case it might be better to test the lot given the chance of an untouched script being broken by an engine change.

PR #4040, opened 2016-05-20
Fix formatting for bugfix 17624 note

montegoulding:


livecodefraser:
@livecode-vulcan reviewed by @montegoulding okay 3326eb6

Merging without waiting as it only affects the installer build.

PR #4051, opened 2016-05-23
Update mergExt pointer

montegoulding:


peter-b:
@montegoulding should this be targeted for `develop-8.0` so that the changes go into 8.0.2?

montegoulding:
Yes it probably should, I'll rebase

PR #4062, opened 2016-05-24
Add OSX to our Travis build file (experimental!)

livecodefraser:
Lets see if this works!

The build for OSX is currently set as allowed-to-fail because the lack of a 10.6 SDK will cause revvideograbber to fail to build.


livecodefraser:
Okay, progress: things seem to be working, with the exception of the TestCutVar test, which fails.

livecodeali:
@livecodefraser This should be good to go now - I'm not sure what is needed to make it use the updated `.travis.yml` from develop - it might need pushing to a new branch and submitting a PR. I thought just making Travis try again would work but apparently not.

PR #4100, opened 2016-06-03
[[ Bug 17731 ]] Work around issues with revert command

livecodeali:
- ensure the stack to be reverted is the `topStack`
- ensure the defaultstackptr is changed


livecodeali:
This is my best guess at a fix for 17731.

PR #4111, opened 2016-06-06
Use a different timestamp service

livecodepanos:
Attempt to fix the timestamp-related build failures


livecodefraser:
@livecode-vulcan review okay 67bf5c4

If this works, we should add an outer loop that tries multiple URLs in case one of them is down.

livecodepanos:
Merging this now, since it only affects the installers build

livecode-vulcan:
💙 review by @livecodefraser ok 67bf5c4b71db1dea1f6032b982ec52e82d87bd7d

PR #4060, opened 2016-05-24
[[ Bug ]] Fix pattern number behaviour

montegoulding:
Pattern number behaviour did not match the documented behaviour of
mapping pattern number 1 to 164 to image id 136 to 300.

PR #4128, opened 2016-06-09
Bugfix 17553

mwieder:
Rebased, pushed, and reopened. Couldn't reopen the previous PR (3996) because of the rebase, so I'm resubmitting this.

The tools palette in LC8 attempts to assign some very large numbers (no doubt converted from negatives) to the spray, brush, and eraser properties, and these numbers do not correspond to valid images. That causes an error which is not normally displayed to the users, since it happens in a system stack. This patch fixes some suspicious-looking math in the MCInterfaceSetBrush function and also vectors setting the brush values in MCGraphicsExecResetPaint through MCInterfaceSetBrush so that the values get validated on the way in.

For extra credit, I fixed the default value of the spray brush, which was getting set to 31 instead of the documented value of 34. Something else must be fixing the value later on, because retrieving the value shows 34 anyway.

This fixes http://quality.livecode.com/show_bug.cgi?id=17553.

### DIFF
--- a/docs/dictionary/property/spray.lcdoc
+++ b/docs/dictionary/property/spray.lcdoc
@@ -17,7 +17,7 @@ set the spray to 30
 
 Value:
 The <spray> is a brush specifier.
-A <brushID> is a built-in brush number between 1 and 100. (These brushes correspond to LiveCode's built-in patterns 100 to 135.)
+A <brushID> is a built-in brush number between 1 and 35. (These brushes correspond to LiveCode's built-in patterns 101 to 135.)
 
 An <imageID> is the ID of an <image> to use for painting with the spray can. LiveCode looks for the specified <image> first in the <current stack>, then in other open <stacks>.
 
--- a/docs/dictionary/property/spray.lcdoc
+++ b/docs/dictionary/property/spray.lcdoc
@@ -17,7 +17,7 @@ set the spray to 30
 
 Value:
 The <spray> is a brush specifier.
-A <brushID> is a built-in brush number between 1 and 35. (These brushes correspond to LiveCode's built-in patterns 101 to 135.)
+A <brushID> is a built-in brush number between 1 and 36. (These brushes correspond to LiveCode's built-in patterns 101 to 136.)
 
 An <imageID> is the ID of an <image> to use for painting with the spray can. LiveCode looks for the specified <image> first in the <current stack>, then in other open <stacks>.
 
--- a/engine/src/exec-interface2.cpp
+++ b/engine/src/exec-interface2.cpp
@@ -2026,7 +2026,8 @@ void MCInterfaceSetRelayerGroupedControls(MCExecContext& ctxt, bool p_value)
 void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_value)
 {
 	uint4 t_newbrush = p_value;
-	if (t_newbrush < PI_PATTERNS)
+	// MDW-2016-05-06 [[ bugfix_17553 ]] safer to compare against PI_BRUSHES than PI_PATTERNS
+	if (t_newbrush < PI_BRUSHES)
 		t_newbrush += PI_BRUSHES;
 
 	// MW-2009-02-02: [[ Improved image search ]]
--- a/engine/src/exec-interface2.cpp
+++ b/engine/src/exec-interface2.cpp
@@ -2026,8 +2026,6 @@ void MCInterfaceSetRelayerGroupedControls(MCExecContext& ctxt, bool p_value)
 void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_value)
 {
 	uint4 t_newbrush = p_value;
-	// MDW-2016-05-21 [[ bugfix_17553 ]] map low brush values up to 101..135 range
-	// if less than 36, map to 101..135
 	if (t_newbrush <= (PI_PATTERNS-PI_BRUSHES))
 		t_newbrush += PI_BRUSHES;
 
@@ -2065,7 +2063,6 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_v
 
 void MCInterfaceGetBrush(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
 	r_value = MCbrush > PI_BRUSHES && MCbrush <= PI_PATTERNS ? MCbrush - PI_BRUSHES : MCbrush;
 }
 
@@ -2076,7 +2073,6 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, uinteger_t p_value)
 
 void MCInterfaceGetEraser(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
 	r_value = MCeraser > PI_BRUSHES && MCeraser <= PI_PATTERNS ? MCeraser - PI_BRUSHES : MCeraser;
 }
 
@@ -2087,7 +2083,6 @@ void MCInterfaceSetEraser(MCExecContext& ctxt, uinteger_t p_value)
 
 void MCInterfaceGetSpray(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
 	r_value = MCspray > PI_BRUSHES && MCspray <= PI_PATTERNS ? MCspray - PI_BRUSHES : MCspray;
 }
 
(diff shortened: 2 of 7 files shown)


montegoulding:
Thanks @mwieder could you add a bug fix note please

docs/dictionary/property/spray.lcdoc
```diff
@@ -17,7 +17,7 @@ set the spray to 30
 
 Value:
 The <spray> is a brush specifier.
-A <brushID> is a built-in brush number between 1 and 100. (These brushes correspond to LiveCode's built-in patterns 100 to 135.)
+A <brushID> is a built-in brush number between 1 and 35. (These brushes correspond to LiveCode's built-in patterns 101 to 135.)
```
montegoulding: this should be between 1 and 35

montegoulding: er... I mean 36 sorry

engine/src/exec-interface2.cpp
```diff
@@ -1988,7 +1988,9 @@ void MCInterfaceSetRelayerGroupedControls(MCExecContext& ctxt, bool p_value)
 void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_value)
 {
 	uint4 t_newbrush = p_value;
-	if (t_newbrush < PI_PATTERNS)
+	// MDW-2016-05-21 [[ bugfix_17553 ]] map low brush values up to 101..135 range
+	// if less than 36, map to 101..135
```
montegoulding: @mwieder we don't require the author note anymore and rely on the git log instead. These comments could be removed but we would need to fix the range if they aren't removed.

engine/src/exec-interface2.cpp
```diff
@@ -2025,7 +2027,8 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, Properties p_which, uinteger_t p_v
 
 void MCInterfaceGetBrush(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	r_value = MCbrush < PI_PATTERNS ? MCbrush - PI_BRUSHES : MCbrush;
+	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
```
montegoulding: as above

engine/src/exec-interface2.cpp
```diff
@@ -2035,7 +2038,8 @@ void MCInterfaceSetBrush(MCExecContext& ctxt, uinteger_t p_value)
 
 void MCInterfaceGetEraser(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	r_value = MCeraser < PI_PATTERNS ? MCeraser - PI_BRUSHES : MCeraser;
+	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
```
montegoulding: as above

engine/src/exec-interface2.cpp
```diff
@@ -2045,7 +2049,8 @@ void MCInterfaceSetEraser(MCExecContext& ctxt, uinteger_t p_value)
 
 void MCInterfaceGetSpray(MCExecContext& ctxt, uinteger_t& r_value)
 {
-	r_value = MCspray < PI_PATTERNS ? MCspray - PI_BRUSHES : MCspray;
+	// MDW-2016-05-21 [[ bugfix_17553 ]] map 101..135 down to 1..35
```
montegoulding: and here ;-)

tests/lcs/core/graphics/graphics.livecodescript
```diff
@@ -90,4 +90,163 @@ TestAssert "test", not ("4,9" is not within "-5,-10,5,10")
 TestAssert "test", "-6,-11" is not within "-5,-10,5,10"
 TestAssert "test", "5,10" is not within "-5,-10,5,10"
 
-end TestGraphicsIsNotWithin
\ No newline at end of file
+end TestGraphicsIsNotWithin
+
+-- new brush mapping for 101..135
```
montegoulding, line 95: 136 here again also 101 < brush <= 136 = brush - 100

mwieder, line 95: I don't think that's correct. PI_PATTERNS as defined in parsedef.h is 136, so the brushes stop at 135.

mwieder:
I'd really like to get rid of this whole mapping thing and not have to think about magic numbers any more.

tests/lcs/core/graphics/graphics.livecodescript
```diff
@@ -90,4 +90,163 @@ TestAssert "test", not ("4,9" is not within "-5,-10,5,10")
...
+	TestAssert "the grid is initially true", the grid is true
+
+	TestAssert "the gridSize is initially 4", the gridSize is 4
+	TestAssert "the lineSize is initially 0", the lineSize is 0
+
+	TestAssert "the multiple is initially false", the multiple is false
+	TestAssert "the multiSpace is initially 1", the multiSpace is 1
+
+# not sure if this should work. it doesn't in previous builds either
+	TestAssert "the penPattern is initially 136", the penPattern is 136
```
montegoulding: I have done a separate PR for patterns which is now merged in. It has the behavior of still subtracting `PI_PATTERNS` if the brush == `PI_PATTERNS` that way the default 136 returns 0 because 136 is actually a brush!. So if you test this against a build of develop the initial penPattern should be 0.

montegoulding: Ah.. my patternNumber PR is merged into develop-8.0 and hasn't made its way to develop yet. We will probably merge it up tonight and when it does this test should start failing because it will be initially 0 and reset to 0.

montegoulding: Argh.. actually sorry it will be initially empty because it is documented you can set it to empty and it is initially empty so when it == PI_PATTERNS it will return empty. Here is my [PR](https://github.com/livecode/livecode/pull/4060) if you want to see the tests or implementation considering it impacts yours

tests/lcs/core/graphics/graphics.livecodescript
montegoulding, line 95: The brushes stack has an image ID 136. In this instance the docs are wrong which of course means we need to ensure all the docs say the correct range.  The following from your PR actually has the range 101-136.

```
MCspray > PI_BRUSHES && MCspray <= PI_PATTERNS
```

tests/lcs/core/graphics/graphics.livecodescript
mwieder: Nonetheless, the documentation says, and has always said, that the brushes are between 1 and 35. All I did was bring the remaining errant spray doc in line, since it was quite wrong to begin with. We can start making 136 a brush if you want (and I reiterate that if we remove the mapping then it doesn't matter what we call these things, we can just set the id value and use it as a brush or as a pattern, the way we can do now with other image ids not in the special set.

montegoulding: As long as the docs match the engine functionality I don't mind and I doubt anyone else will. At the moment your PR allows setting and getting 36 <-> 136 though so we need to change it one way or another.

mwieder: Hmmm... you're right. I even have a test for that. I don't really care whether we call image id 136 a brush or a pattern or both. If you've mapped patterns starting at 137 then let me know and I'll adjust the docs. Or if you're using 136 as a pattern I'll change the mapping and test here.

mwieder: ...and I could use a clue as to how to add a release note.

montegoulding: Yes patterns are `137->300` in line with the original MC stack and the LC one. Brushes in the MC stack and LC are `101->136`. Add a release note to `docs/notes/bugfix-{bug number}.md`

mwieder:
Had to adjust the tests. After Monte's latest merge of the pattern PR the initial and reset brush and pattern are set to empty.

montegoulding:
@livecode-vulcan review ok 9d9e035

👍 Thanks @mwieder

livecode-vulcan:
💙 review by @montegoulding ok 9d9e0354a63e98fcc5b9a3ffa2f769b42aac95f4

livecode-vulcan:
😎 test success 9d9e0354a63e98fcc5b9a3ffa2f769b42aac95f4
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/2635)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/2645)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/2630)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/2646)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/2652)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/2658)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/2670)

PR #4135, opened 2016-06-09
make local timezone default, fix setting local tz

BerndN:
clock widget

### DIFF
--- a/bugfix-17373.md
+++ b/bugfix-17373.md
@@ -1,0 +1,1 @@
+#enable setting local timezone, make local the default timezone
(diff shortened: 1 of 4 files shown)


peter-b:
@BerndN Thank you very much for patching this!

Can you please add a test to `extensions/widgets/clock/tests/` to help make sure this bug doesn't happen again?  You can find some examples of tests for widgets in the other widgets' directories.

I think you've created the bugfix note not quite in the right place -- I think it should probably be in `extensions/widgets/clock/notes/`.  The bugfix notes for widgets use a slightly different format to the one used by the engine.  Once again, take a look at the release notes for some of the other widgets for an example.

BerndN:
@peter-b  I hope I got the note right this time, I tried to add a test. Not sure if it is right.

If I only knew how to delete the first note that is to the wrong directory.

BerndN:
The test file is in the clock folder, but should probably be in a clock/tests folder. But I don't know how to create a subfolder. Could I do it by changing the path in the file name?
At least the test seems to have passed Travis CI build.

peter-b:
@BerndN I've taken the liberty of taking your changes and cleaning up the version control information.  I've made a new pull request, #4170.

PR #4146, opened 2016-06-13
Include full range of commits when computing relevant release notes

peter-b:
Previously, if there was a series of tags:
- 8.0.2-rc-1
- 8.0.2-rc-2
- 8.1.0-dp-1

Then only the changes between 8.1.0-dp-1 and HEAD would be considered
when computing release notes.  This patch ensures that all changes
between 8.0.2-rc-2 and HEAD are considered.

### DIFF
--- a/builder/release_notes_builder.livecodescript
+++ b/builder/release_notes_builder.livecodescript
@@ -1369,13 +1369,14 @@ private function GitGetRelevantTags pType
    
    set the itemdelimiter to "."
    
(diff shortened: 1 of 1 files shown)


livecodefraser:
@livecode-vulcan review okay 2e0a0ed

livecode-vulcan:
💙 review by @livecodefraser ok 2e0a0edfd86379e828043a367ff8d5268edaa80a

PR #4212, opened 2016-06-29
Fixed typo that caused installation to fail on Windows

livecodepanos:
This is just a quick fix. A more robust fix would be maybe to make the parser to resolve automatically this kind (`this/is/a//path)` of paths


livecodeali:
@livecode-vulcan review ok fe1f65a

livecode-vulcan:
💙 review by @livecodeali ok fe1f65adfb1fd07868ce11c158b75593733bef71

peter-b:
Merging this immediately since the CI does not currently cover the package compiler.

PR #4196, opened 2016-06-23
[[ tsNet ]] Implement changes required to support tsNet libURL driver

montegoulding:
This patch provides support for the inclusion of
tsNet into the livecode package. All additional
externals (mergExt & tsNet) will now be found
in the same `Tools/Ext` directory.

Support has been implemented for a directory within
`Tools/Ext` to contain livecode files that may insert
themselves into the IDE message path and standalone
build process. This was required for tsNet's script
based libURL driver. These files will have
`revLoadLibrary` and `revUnloadLibrary` sent to
them so they may manage their own lifecycle as
back script, front script, library, behavior or
just loaded stack.

In the course of development a bug was identified
and resolved where android externals were potentially
being filtered from the included files. The
`revSBLibrary` was elevated to an always loaded
back script in the IDE due to the many scripts
depending on it.

There are dependencies between this patch and an
associated patch to the livecode-ide repo.

PR #4238, opened 2016-07-05
[[ Bug ]] Support multiple architectures for Ext externals

montegoulding:
Previously Linux -x64.so and -x86.so were being
listed as separate inclusions. This patch allows
for the following naming convention:

`<externalName>[{-|_}{<architecture>|<architecture synonym>].<extension>`

If no architecture is specified then the default
architecture is assumed.

https://github.com/livecode/livecode/pull/4196 Will depend on this

### DIFF
--- a/ide-support/revsblibrary.livecodescript
+++ b/ide-support/revsblibrary.livecodescript
@@ -708,7 +708,7 @@ constant kMacOSXArchitectures = "x86-32,x86-64"
 constant kLinuxArchitectures =  "x86-32,x86-64,armv6-hf"
 constant kAndroidArchitectures =  "armv6"
 constant kiOSArchitectures =  "armv7,arm64"
-constant kEmscriptenArchitectures =  "javascript"
+constant kEmscriptenArchitectures =  "js"
 
 function revSBPlatformArchitectures pPlatform
    switch pPlatform
--- a/ide-support/revsblibrary.livecodescript
+++ b/ide-support/revsblibrary.livecodescript
@@ -1491,6 +1491,17 @@ private function revSBFilterExternalsForDesktop pList, pPlatform
    return pList
 end revSBFilterExternalsForDesktop
 
+function revSBExternalNameFromFile pPlatform, pFile
+   local tName
+   set the itemdel to "."
+   put item 1 of pFile into tName
+   
+   local tArch
+   __ExternalArchitecture pPlatform, tName, tArch
+   
+   return tName
+end revSBExternalNameFromFile
+
 -- This is currently just used for externals that are present in the
 -- 'mergExt' folder in the app bundle or equivalent
 local sExternalsLocationA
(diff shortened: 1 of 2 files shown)


ide-support/revsblibrary.livecodescript
```diff
@@ -702,6 +702,98 @@ function revSBTargetToPlatform pTarget
    end if
 end revSBTargetToPlatform
 
+-- item 1 of the architectures should be the default architecture
+constant kWindowsArchitectures = "x86-32"
+constant kMacOSXArchitectures = "x86-32,x86-64"
+constant kLinuxArchitectures =  "x86-32,x86-64,armv6-hf"
+constant kAndroidArchitectures =  "armv6"
+constant kiOSArchitectures =  "armv7,arm64"
+constant kEmscriptenArchitectures =  "javascript"
```
peter-b: I think we use "js" as the architecture for emscripten buids.

montegoulding: OK, I'll change. This architecture stuff isn't actually used for mobile or html5 but in the event it is one day it would be good if it were correct!

ide-support/revsblibrary.livecodescript
```diff
@@ -702,6 +702,101 @@ function revSBTargetToPlatform pTarget
...
+         set the itemDelimiter to comma
+      end repeat
+      
+      if tFoundArchitecture is not empty then
+         exit repeat
+      end if
+   end repeat
+   
+   if tFoundArchitecture is empty then
+      put item 1 of tArchitectures into tFoundArchitecture
```
livecodeali, line 774: The itemdelimiter won't be comma here if the exit repeat on line 763 is hit, will it? I think it needs to be reset at some point.

ide-support/revsblibrary.livecodescript
```diff
@@ -1430,23 +1536,26 @@ function revSBAdditionalExternalsList pPlatform
          set the itemdelimiter to "."
          set the linedelimiter to slash
          repeat for each element tExtPath in tRawList
-            put tExtPath into sExternalsLocationA[pPlatform][item 1 of line -1 of tExtPath]
+            local tExtension, tExternalName
+            put item 2 of line -1 of tExtPath into tExtension
+            put item 1 of line -1 of tExtPath into tExternalName
```
livecodeali: This was a bug in the existing implementation, but do you think this should be

```
put item -1 of line -1 of tExtPath into tExtension
put item 1 to -2 of line -1 of tExtPath into tExternalName
```

in case the external name contains a `.`?

ide-support/revsblibrary.livecodescript
montegoulding, line 774: good catch, thanks @livecodeali

ide-support/revsblibrary.livecodescript
montegoulding: good idea

livecodeali:
@livecode-vulcan review ok 0c0ff3a

livecode-vulcan:
💙 review by @livecodeali ok 0c0ff3a9af0fa9894196f79cb55f2b69308a8541

PR #4242, opened 2016-07-06
🍄 Add linting framework and CLI tool

peter-b:


montegoulding:
@peter-b This is great. What are the chances of being able to pipe a file to this for use in a pre-commit hook? I guess we could write staged files to a temporary file if necessary. I suspect if we did allow it we would want to specify rules or lints for the file coming on stdin. 

Will this be run using the ide engine because that has better facilities for script linting (handler names, variable names) and also has access to script parser error strings.

montegoulding:
@peter-b I see you are using the standalone engine. Would it be an issue to change?

Also it looks like you have fixed the mac-bin so it points to the correct `_build` directory so perhaps my mac-bin gitignore patch should be closed as the symlink would be no longer necessary.

PR #4247, opened 2016-07-08
Use MCAutoStringRefAsCString for temp const C strings

peter-b:
Refactor several places that were always performing an explicit copy using `MCStringConvertToCString()` so that they use an `MCAutoStringRefAsCString` instead.  This avoids a memory copy if the string is already in a suitable format for use as a C string instead, which will hopefully speed things up.

As a side effect, this fixes some "allocate with `malloc()`, free with `delete`" errors detected by valgrind.

### DIFF
--- a/engine/src/uidc.cpp
+++ b/engine/src/uidc.cpp
@@ -1575,9 +1575,9 @@ void MCUIDC::siguser()
 
 Boolean MCUIDC::lookupcolor(MCStringRef s, MCColor *color)
 {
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok b0a102a

livecode-vulcan:
💙 review by @livecodeali ok b0a102a83ce4c2fc23fb36066aa8c664505177bd

PR #4295, opened 2016-07-16
Build revvideograbber dependencies as universal binary

montegoulding:
This patch fixes compile errors in the revvideograbber
project which is i386 only when building the LiveCode-all
target in Xcode as 64 bit by building its dependences
as universal binaries.

### DIFF
--- a/libcore/libcore.gyp
+++ b/libcore/libcore.gyp
@@ -70,6 +70,17 @@
 					'include',
 				],
 			},
+			
+			'configurations':
+			{
+				'Debug':
+				{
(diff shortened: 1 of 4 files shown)


peter-b:
This needs to be rebased against `develop-8.1` or `develop` (probably the latter).

montegoulding:
You can just edit the PR now

montegoulding:
@peter-b @livecodefraser Should I close this? Is there a better alternative?

montegoulding:
Closing due to an expected 🔥

PR #4312, opened 2016-07-22
build: Ensure removed dictionary entries don't break the build

peter-b:
Both the dictionary info gathering and the release note fragment
gathering steps use lists of changed files in order to determine which
files to consider.  Some of these files have changed because they've
been removed.  It isn't an error or even a warning if these files are
missing; it's just an interesting and perhaps notable fact.


livecodepanos:
@livecode-vulcan review ok 3ea4b19

livecode-vulcan:
💙 review by @livecodepanos ok 3ea4b19ea691958122ac103336b2a6c3f743f62a

PR #4341, opened 2016-08-08
Use GNU-style triples when configuring

livecodefraser:
Work in progress.

The intention is to make it a bit easier to add/remove/change the list of platforms that we support.

### DIFF
--- a/configure
+++ b/configure
@@ -1,0 +1,675 @@
+#!/bin/bash
+# Copyright (C) 2015-2016 LiveCode Ltd.
+#
+# This file is part of LiveCode.
+#
+# LiveCode is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License v3 as published by the Free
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -0,0 +1,589 @@
...
+    # Always use Linux-style makefiles for Android as the Android toolchain
+    # is more Linux-y than Darwin-y
+    linux|android|emscripten) FORMATS="make-linux" ;;
+    mac|ios)       FORMATS="xcode" ;;
+    win)           FORMATS="msvs" ;;
+  esac
+fi
+
+# Default "depth"
+if [test] -z "$DEPTH"; then
```
peter-b: Should this be `if [[ -z ${DEPTH} ]]; then` ?

livecodefraser: Just a silly typo. Fixed locally but I haven't re-pushed.

PR #4374, opened 2016-08-16
[[ Bug 18032 ]] Remove broken parameter evaluation from send script

montegoulding:
This patch removes the parameter evaluation from send when a chunk
of script is being sent rather than a command. In the event that
`message` fails the parameter no longer makes sense or at least
would need to be far more sophisticated than the present
implementation.

### DIFF
--- a/docs/notes/bugfix-18032.md
+++ b/docs/notes/bugfix-18032.md
@@ -1,0 +1,1 @@
+# Resolve poor parsing of command parameters in message box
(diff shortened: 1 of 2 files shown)


livecodeali:
@livecode-vulcan review ok 29b2ba8

livecodeali:
I guess this should go into 8.2.

montegoulding:
yes actually I was going to look into this again because I think this will break some code that someone was posting on the use list this week

montegoulding:
```
on mouseUp
     put the short id of this card into KARD
     put the name of this stack into STAQ
     put the long id of me into DIMENA
     send ("delete DIMENA") to card id KARD of stack STAQ in 0 ticks
end mouseUp
```

livecode-vulcan:
💙 review by @livecodeali ok 29b2ba867b1eb21f14152eaf4f604e0df36d6e4e

montegoulding:
@livecode-vulcan review not ok 29b2ba8

livecode-vulcan:
💔 review by @montegoulding not ok 29b2ba867b1eb21f14152eaf4f604e0df36d6e4e

montegoulding:
I'm thinking we may need to implement `do <script> in object` instead...

PR #4407, opened 2016-08-23
Update current-card.lcdoc

PaulMcClernan:
Added two other synonyms
Added two usage examples
Added OS and Platforms
need to add synonym "currentCard" (or current card) to the dictionary too.
May all three synonyms should to combine with "this stack" reference that is in dictionary.

### DIFF
--- a/docs/glossary/c/current-card.lcdoc
+++ b/docs/glossary/c/current-card.lcdoc
@@ -1,13 +1,22 @@
 Name: current card
 
-Synonyms: current card
+Synonyms: current card, this card, currentcard
 
 Type: glossary
 
 Description:
 The <card> currently being displayed in a <stack window>.
 
+Example:
+put the long id of the current card of this stack
+
+Example:
+set the marked of the current card of this stack to false
+
 References: card (glossary), stack window (glossary)
 
 Tags: navigation
 
+OS: mac, windows, linux, ios, android
+
+Platforms: desktop, server, mobile


livecodeali:
@PaulMcClernan Thanks for the contribution - I think most of this would be better in a new dictionary entry in `property/currentCard.lcdoc`.

Glossary entries are supposed to be explanations of terms used in LiveCode docs that are _not_ syntax. As such, the `Platforms:` and `OS:` elements don't really apply to glossary entries (I should probably update the documentation format reference / contributing guide to say something about glossary entries as I realise there is not much there)

We probably could do with working out a way of using examples better in glossary entries. I think the examples you have added might work with a comment which explain the use of the term "current card" rather than elucidate any particular syntax. For example:

```
Example: 
-- store the long id of the current card in a variable
local tVar
put the long id of this card into tVar
```

As it stands, neither of your examples actually compiles because there is no `current card` syntax.

Also none of "current card", `this card` and `currentCard` are actually synonyms - `currentCard` is a stack property, whereas `this card` is an object chunk reference (for example you can do `put the long id of this card` but _not_ `put the long id of the currentCard`

livecodeali:
Ah, I have just realised that `current` and `this` are actually synonyms, so 
`put the long id of current card` is ok but `put the long id of the current card` is not

PaulMcClernan:
put the long id of the current card
does actually work, I just copy & pasted it in from your comment to the msg box and it gave me the path to the msg box stack in the IDE.

livecodeali:
Ah, that is unfortunate from the point of view of strict chunk parsing :-(

I also note that `put the long id of the this card` works...

PaulMcClernan:
Hah, it seems the interpreter is ok with stuttering! The "the" is just ignored.
put the the the the long id of the current card of the the the the this stack
works!

PaulMcClernan:
So I created property/currentCard.lcdoc , cheers.

PR #4414, opened 2016-08-24
[[ Community Docs ]] Added entry for "open stack"

matthewmaier:
- copied one of the other open-whatever.lcdoc files as a template
- I'm not actually sure about all the details, so left them blank
- mostly I wanted to add the code example for setting the defaultFolder, but first I needed to have the open stack command in the dictionary

### DIFF
--- a/docs/dictionary/command/open-stack.lcdocs
+++ b/docs/dictionary/command/open-stack.lcdocs
@@ -1,0 +1,42 @@
+Name: open stack
+
+Type: command
+
+Syntax: open stack <filePath>
+
+Summary:
+Opens a stack so its contents can be accessed or modified.
+
+Introduced: 1.0
+
+OS: mac, windows, linux, ios, android
+
+Platforms: desktop, server, mobile
+
+Security: disk
+
+Example:
+open stack "temp.livecode"
+
+Example:
+put the filename of me into tMe
+set the itemdelimiter to "/"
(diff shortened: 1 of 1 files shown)


livecode-vulcan:
📝 Hi @matthewmaier, I haven't been able to verify that you've signed our [LiveCode Contributor's Agreement](https://livecode.com/account/developer/contribute).

Please see the [information for contributors](https://github.com/runrev/livecode/blob/develop/CONTRIBUTING.md) for more information.

livecodeali:
@matthewmaier The reason there is no `open-stack.lcdoc` is that this syntax is actually documented in `go.lcdoc`, as `open` and `go` are synonyms.

I think it is worthwhile adding your example about the `defaultFolder` (I did not know that!) but I suggest you use a slightly different example, since you already have the filename! How about something like:

```
set the defaultFolder to specialFolderPath("Documents")
if there is a file "mystack.livecode" then
  -- stack filename relative to the defaultFolder
   open stack "mystack.livecode"
end if
```

Looking at the entry for `go`, I also notice that the syntax for going to a stack directly is missing, so the first bit of syntax should probably be

```
Syntax: go [invisible] [to] {<card>|<stack>} [{as <mode> |in [a] new window|in <window>}]
```

matthewmaier:
Thanks, I killed the open-stack.lcdoc idea and modified go.lcdoc

PR #4392, opened 2016-08-17
[Bug 18293] Use MCObjectHandle for button menu

montegoulding:

PR #4430, opened 2016-08-29
Refactor MCObjectHandle to allow MCButtonHandle, etc

livecodefraser:
This change is [Reviewable](https://reviewable.io/reviews/livecode/livecode/4430)


engine/src/object.h
```diff
@@ -1268,6 +1315,17 @@ class MCObject : public MCDLlist
 	bool getdeletionissuspended(void) const { return m_deletion_is_suspended; }
 	void setdeletionissuspended(bool value) { m_deletion_is_suspended = value; }
 
+    // Because C++11 support is universal amongst our supported platforms, auto-
```
montegoulding: @livecodefraser should there be a `not` before universal... otherwise I don't get this comment... but that could just be me ;-)

livecodefraser: There should indeed! That'll teach me to not proof-read my comments :wink:

montegoulding:
Hmm... does this really need to target 8.2 as I was going to use it in #4392 

I guess we can pretty it up during the merge...

peter-b:
Reviewed 29 of 30 files at r1.
Review status: 29 of 30 files reviewed at latest revision, 4 unresolved discussions.

---

_[engine/src/object.h, line 33 [r2]](https://reviewable.io:443/reviews/livecode/livecode/4430#-KQVHY4HTRmjiVTl94Ue:-KQVHY4HTRmjiVTl94Uf:b-8y0zxs) ([raw file](https://github.com/livecode/livecode/blob/b90fc7cb4c122433ba7fca4a740ec13b0f44cfeb/engine/src/object.h#L33)):_

> ``` objc
> #include "native-layer.h"
> 
> // Disabled until C++11 support is available on all platforms
> ```

Please add a `// TODO(C++11)` here for greppability.

---

_[engine/src/object.h, line 230 [r2]](https://reviewable.io:443/reviews/livecode/livecode/4430#-KQVKTsK23Y_CVJwd5X5:-KQVKTsK23Y_CVJwd5X6:b-gfsskd) ([raw file](https://github.com/livecode/livecode/blob/b90fc7cb4c122433ba7fca4a740ec13b0f44cfeb/engine/src/object.h#L230)):_

> ``` objc
> 
>     // RAII handle class
>     class Handle;
> ```

BTW it might be worthwhile to add a comment explaining where to find the actual definition of this class.

---

_[engine/src/object.h, line 1588 [r2]](https://reviewable.io:443/reviews/livecode/livecode/4430#-KQVHSEML1s_uLQdN2Lm:-KQVHSEML1s_uLQdN2Ln:b-8y0zy5) ([raw file](https://github.com/livecode/livecode/blob/b90fc7cb4c122433ba7fca4a740ec13b0f44cfeb/engine/src/object.h#L1588)):_

> ``` objc
> 
>     // Conversion to handles to compatible types
>     // Currently disabled due to lack of required C++11 support on all platforms
> ```

Please add a `// TODO(C++11)` here for greppability!

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/livecode/livecode/4430)_

PR #4440, opened 2016-09-01
[[ Documentation ]] Fixes to and.lcdoc

asayd:
- Syntax changed to show that more than two operands are permissable.
- Updated summary to reflect possibility of two or more operands.
- Added html5 as an OS.
- Improved examples with comments and variable declarations.
- Supplied missing description for 2nd and repeated parameters.
- Clarified the Description, adding explanations for expressions with  more than 2 operands.
- Fixed reference links and references list.

### DIFF
--- a/docs/dictionary/operator/and.lcdoc
+++ b/docs/dictionary/operator/and.lcdoc
@@ -2,11 +2,11 @@ Name: and
 
 Type: operator
 
-Syntax: <value1> and <value2> [and <value3> ...]
+Syntax: <leftValue> and <rightValue>
 
 Summary:
-Allows the creation of compound <boolean> expressions, which <evaluate> to true 
-if all <operand|operands> are true, false otherwise. 
+Allows the construction of compound <boolean> expressions, which <evaluate> to true 
+if both <operand|operands> are true, false otherwise. 
 
 Introduced: 1.0
 
@@ -27,16 +27,12 @@ local myCount
 if the shiftKey is down and myCount > 1 then exit mouseUp
 
 Parameters:
-value1 (bool):
-Value1 is true or false, or an expression that evaluates to
+leftValue: (bool):
+leftValue is true or false, or an expression that evaluates to
 true or false.
 
-value2: (bool):
-Value2 is true or false, or an expression that evaluates to
-true or false.
-
-value3: (bool):
-Value3 is true or false, or an expression that evaluates to
+rightValue (bool):
+rightValue is true or false, or an expression that evaluates to
 true or false.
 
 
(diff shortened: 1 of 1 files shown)


peter-b:
Test failures:

```
not ok - example 1 of and operator compiles
# Example:
# on mouseUp
# (1 > 0) and (1 = 0)
# -- evaluates to false
# end mouseUp
# Script Parsing Error:
#  - (
not ok - example 2 of and operator compiles
# Example:
# on mouseUp
# (1 > 0) and (1 = 1) and (0 = 0)
# -- evaluates to true
# end mouseUp
# Script Parsing Error:
#  - (
```

livecodeali:
@asayd Another thing to note here is that it's fine to replace the entities `&lt;` and `&gt;` with  `<` and `>` in the examples - they won't get parsed for references anyway.

asayd:
@peter-b @livecodeali I've made those changes. What do you think about the parameter list? Is it too repetitive? I'm documenting each valueN param in the syntax block separately to maintain consistency with our style guidelines. Would it be better to change the syntax to 

```
value and value [ and value... ]
```

and just have one param defined?

livecodeali:
@asayd Actually thinking about it I don't think there should be anything else in the syntax other than 
`<leftValue> and <rightvalue>` - multiple `and`s are not really part of the syntax, it's just there's an implicit grouping of the first pair which evaluates to another value.

I think the example of two ands is good enough to illustrate that explicit grouping is not needed.

livecodeali:
@asayd Just looking back through some pull requests - did you see my previous comment here?

asayd:
@livecodeali Yes, I keep meaning to get back to this.
I guess my reaction is, are we supposed to be documenting what's happening in the engine, or are we supposed to be documenting according to the mental model of what the end user thinks is possible? I suppose the naive end user just perceives the situation as being able to tack on as many AND operands as one wants:

```
condition AND condition AND condition AND condition...
```

They're not thinking about what the engine is actually doing:

```
((condition AND condition) AND condition) AND condition
```

(Or so I presume from your comment.)

So my question would be what tips off the end user that they are not limited to two operands in complex boolean statements conjoined with AND or OR? Do we explicitly express that in the API docu, or do we just let them figure it out by experimenting? Put another way, are the APIs to be purely descriptive of the engine's operations, or are they to be more loose and pedagogical? It's a philosophical question that extends beyond more than just this entry.

livecodeali:
@asayd Personally I think the `Syntax:` element should strictly reflect the implementation of the syntax whereas the examples and (if necessary) description can be expanded to include loose / pedagogical information. 

One reason for this is just because there is no other place that syntax is actually documented, and I feel like the ability to know the exact structure of the abstract syntax tree of some LC code is interesting/useful/important for _someone_.

Another is that it would create an additional judgement call for every binary operator which combines two expressions to give another of the same type as to whether to include a third option. Should we also add to the syntax of the plus operator? 

However, if others agree that the pedagogical value of adding it to the syntax outweighs those concerns then I'm ok with it.

asayd:
@livecodeali Well argued. I think that is the right approach. Keep the Syntax element as a strict representation of what's going on in the engine, and use the Examples and Description to expand with practical/pedagogical examples.

I'll roll back my proposed change to the Syntax.

livecodeali:
@livecode-vulcan review ok 5e9d6a4
Thanks @asayd!

livecode-vulcan:
💙 review by @livecodeali ok 5e9d6a4b950b9cda668d0980439f707a147e6bbd

livecode-vulcan:
😎 test success 5e9d6a4b950b9cda668d0980439f707a147e6bbd
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/282)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/286)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/282)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/282)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/286)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/282)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/282)

PR #4451, opened 2016-09-06
[Bug 18111] Use "DejaVu Sans" instead of "FontAwesome" in guide PDF

jameshale:
Substitute "DejaVu Sans" for "FontAwesome" in PDF of guide.

PR #4452, opened 2016-09-06
bugfix-18111-2.md

jameshale:


peter-b:
Hi @jameshale, I think the existing release note for bug 18111 is probably adequate, so let's go ahead and use #4451 without it.

PR #4468, opened 2016-09-09
Test runner UI for in IDE testing

montegoulding:
This is something I hacked onto the test runner a while back. It probably should be a separate stack and I'm not really expecting it not to fail on Travis. This PR is just here to gage general team interest. For one thing it makes writing and testing tests in the IDE a bit quicker and for another it's an easy way to run commercial tests which can't be run by standalones (well would require a custom standalone anyway).

### DIFF
--- a/tests/_testrunner.livecodescript
+++ b/tests/_testrunner.livecodescript
@@ -19,6 +19,9 @@ along with LiveCode.  If not see <http://www.gnu.org/licenses/>.  */
 -- FIXME provide this on the command line
 constant kLogFilename = "_test_suite.log"
 
+local sFilePathA
+
+
 on startup
    send "TestRunnerMain" to me in 0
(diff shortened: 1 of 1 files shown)


montegoulding:
Here's the generated UI
screen shot 2016-09-09 at 11 00 05 am

montegoulding:
oh and despite the promising PR name it doesn't presently load the IDE tests...

livecodeali:
@montegoulding I think this might be really useful, but *more* for the IDE tests than the engine ones, because currently if I fix a bug in the IDE I have to either run the test suite locally on the command line or create a PR so Travis does it. 

Anyway, just saying I am interested :-)

montegoulding:
ok, I'll try and get this rebased so you can have a play

PR #4473, opened 2016-09-12
Initial simctl wrapper implementation

montegoulding:


peter-b:
@montegoulding Since @livecodemichael has restored life to reviphoneproxy, maybe we don't need this any more.  On the other hand, it seems like this might be a more robust alternative in the long term.  How much work do you think it would be to finish replacing reviphoneproxy with your simctl wrapper?

montegoulding:
@peter-b I think there would be a good chunk of work to do in the deploy library and lots of testing with different xcode + OS combos. My gut feeling is we have enough other stuff to do at the moment but we may want to leave this hanging around as a starting point if we want to look at it down the track.

PR #4497, opened 2016-09-14
[[ Bug 18346 ]] Add bugfix note

techstrategies:


montegoulding:
Sorry @techstrategies I should have mentioned on bugzilla that this one won't need a note as one note should cover the whole release for iOS 10 support

PR #4495, opened 2016-09-14
[[ tsNet ]] Update tsNet pointer to version 1.2.3

techstrategies:

PR #4508, opened 2016-09-15
[[ tsNet ]] Update tsNet pointer to version 1.2.3

peter-b:
The same as #4495, but without the merge conflict.


peter-b:
Merging as is, since #4495 passed its build checks successfully.

PR #4515, opened 2016-09-15
Turn on -Werror=unreachable-code

montegoulding:
Have added the bug label because I think the break in REVDB_QUIT() is a bug

### DIFF
--- a/engine/src/exec-pasteboard.cpp
+++ b/engine/src/exec-pasteboard.cpp
@@ -279,7 +279,7 @@ MCTransferType MCPasteboardFindTransferTypeForLegacyClipboard(const MCClipboard*
     // passed that the behaviour can now be corrected but, until then, it is
     // disabled and the "text over images" method is used instead.
     //
-    if (false)
+    if (/* DISABLES CODE */ (false))
     {
         // If we have both an image and text on the clipboard, whichever one comes
         // first determines what we resolve the clipboard type to be.
--- a/engine/src/opensslsocket.cpp
+++ b/engine/src/opensslsocket.cpp
@@ -1153,6 +1153,98 @@ void MCS_secure_socket(MCSocket *s, Boolean sslverify, MCNameRef end_hostname)
 
 bool MCS_hostaddress(MCStringRef &r_host_address)
 {
+#if defined(_MACOSX)
+    bool t_success;
+    t_success = true;
+    
+    SCDynamicStoreRef t_store;
+    t_store = NULL;
+    if (t_success)
+    {
+        t_store = SCDynamicStoreCreate(kCFAllocatorDefault, CFSTR("JSEvaluator"), NULL, NULL);
+        if (t_store == NULL)
+            t_success = false;
+    }
+    
+    CFStringRef t_network_key;
+    t_network_key = NULL;
+    if (t_success)
+    {
+        t_network_key = SCDynamicStoreKeyCreateNetworkGlobalEntity(kCFAllocatorDefault, kSCDynamicStoreDomainState, kSCEntNetIPv4);
+        if (t_network_key == NULL)
+            t_success = false;
+    }
+    
+    CFDictionaryRef t_network_value;
+    t_network_value = NULL;
+    if (t_success)
+    {
+        t_network_value = (CFDictionaryRef)SCDynamicStoreCopyValue(t_store, t_network_key);
+        if (t_network_value == NULL)
+            t_success = false;
+    }
+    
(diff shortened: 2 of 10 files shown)


montegoulding:
Labelling WIP for merge conflict reasons so we can merge all these in order

engine/src/exec-pasteboard.cpp
```diff
@@ -279,7 +279,7 @@ MCTransferType MCPasteboardFindTransferTypeForLegacyClipboard(const MCClipboard*
     // passed that the behaviour can now be corrected but, until then, it is
     // disabled and the "text over images" method is used instead.
     //
-    if (false)
+    if (/* DISABLES CODE */ (false))
```
peter-b, line 282: Could you please add something about using `/* DISABLES CODE */ (...)` in this way to `CONTRIBUTING.md`?  It's similar to our use of `/* UNCHECKED */` in many ways and if it's done in a consistent way then it becomes similarly greppable.

engine/src/opensslsocket.cpp
```diff
@@ -1153,6 +1153,98 @@ void MCS_secure_socket(MCSocket *s, Boolean sslverify, MCNameRef end_hostname)
...
+    if (t_network_key != NULL)
+        CFRelease(t_network_key);
+    
+    if (t_store != NULL)
+        CFRelease(t_store);
+    
+    return MCStringCreateWithCString(t_result, r_host_address);
+ 
+#else
+
```
peter-b, line 1247: It might be worth going for a code structure like:

``` C++
#if defined(FOO)

#elif defined(BAR)

#elif defined(BAZ)

#else
#   error Function not implemented for this platform
#endif
```

Using this approach makes it much easier to port to new platforms.

revdb/src/revdb.cpp
```diff
@@ -405,8 +405,10 @@ void REVDB_QUIT()
 			}
 		}
 		if (databaserec && databaserec ->releaseconnectionptr)
+        {
 			(*databaserec->releaseconnectionptr)(tconnection);
-		break;
+            break;
+        }
```
peter-b, line 411: Yep, definitely found a bug here.

engine/src/exec-pasteboard.cpp
montegoulding: Sure, it's just something that Xcode suggested ;-)

engine/src/opensslsocket.cpp
montegoulding: Hmm... in this case the code at the end needed to be run on all platforms but `_MACOSX` which is why I used `#else`

peter-b: @montegoulding Derp, I misread your changes. Go on then!

engine/src/exec-pasteboard.cpp
peter-b: It's a good idea and I'm glad that Xcode suggested it, so I think it's worth making official 😉

livecodeali: Is there a good reason to use `if (false)` rather than just `ifdef`fing the block with something undefined?

montegoulding: because it was already `if (false)` ;-)

livecodeali: I meant in general, but that's definitely a good reason in this case :-)

peter-b:
@livecode-vulcan review ok 4cb170a

livecode-vulcan:
💙 review by @peter-b ok 4cb170aed49f208f53133433017be614fb70b3dc

livecode-vulcan:
😎 test success 4cb170aed49f208f53133433017be614fb70b3dc
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/315)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/320)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/315)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/315)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/321)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/315)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/315)

PR #4532, opened 2016-09-21
[Bug 18449] Use xvfb-run to run testsuite on Travis CI

peter-b:
The `xvfb-run` command launches a virtual, framebuffer-based X server
as a background process and then runs a specified comand using the
virtual display.

This should make it possible to run tests on Travis CI without needing an
actual desktop.


peter-b:
It might be worthwhile to rebase this patch against `develop-8.1`.

peter-b:
Marking as WIP pending sorting out the IDE tests.

PR #4546, opened 2016-09-22
Error returned by hostnametoaddress was not being reported to the user

trevordevore:
ulHostNameToAddressCallback needs to report a detailed error rather thant "invalid host address". This fix reports the actual error.

### DIFF
--- a/config/mac.gypi
+++ b/config/mac.gypi
@@ -132,8 +132,11 @@
 							'-Werror=mismatched-new-delete',
 							'-Werror=constant-logical-operand',
 							'-Werror=unknown-pragmas',
+<<<<<<< 183262100ed8e784feb7f5a639c48c340adeccaf
 							'-Werror=missing-field-initializers',
 							'-Werror=objc-literal-compare',
+=======
+>>>>>>> Turn on -Werror=unknown-pragmas
 						],
 					},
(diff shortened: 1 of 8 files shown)


peter-b:
@trevordevore Thanks for this bugfix! I think we should probably include it in the next round of 8.1.x releases.  In order to for us to do this, I would really appreciate it if you could:
- Rebase this PR on the `develop-8.1` branch
- File a bug report describing the problem and how to reproduce it
- Add a release note to this PR
- Ideally, write a unit test to make sure that this bug never happens again

Please let me know if you run into any problems -- I would be delighted to help you backport the bug fix if necessary.

trevordevore:
@peter-b No need to back port. I'm finally moving to 8.1 (I had to modify a lot of code that used the executionContexts) and I want to switch to using the tsNet version of libURL. I'm going to go through and submit some PRs with fixes and additions from my own custom libURL.

peter-b:
@trevordevore Okay, as long as you're happy with them only going into LiveCode 9 ;-)

trevordevore:
@peter-b sorry,  I thought you were referring to backporting to 8.0. I'm with you now. I'm going to move this over to 8.1.

PR #4561, opened 2016-09-23
[Bug 18468] liburl: Add cookie & internet date support

trevordevore:
This commit adds cookie support. Internet date handlers are included as well as one of them is required for cookie support.

### DIFF
--- a/docs/dictionary/command/libURLProcessCookies.lcdoc
+++ b/docs/dictionary/command/libURLProcessCookies.lcdoc
@@ -6,7 +6,7 @@ Syntax: libURLProcessCookies {true|false}
 
 Associations: internet library
 
-Introduced: 8.2
+Introduced: 9.0
 
 OS: mac, windows, linux
 
@@ -23,7 +23,7 @@ libURLProcessCookies false
 Description:
 <libURLProcessCookies> sets whether or not the libURL will transparently parse
 and store cookies sent from a server and then send those cookies back to the server
-with any requests.
+with any requests. By default this setting is off.
 
 >*Important:*  The <libURLProcessCookies> <command> is part of the 
 > <Internet library> on desktop platforms. To ensure that the <command> 
--- a/docs/dictionary/command/libURLProcessCookies.lcdoc
+++ b/docs/dictionary/command/libURLProcessCookies.lcdoc
@@ -1,0 +1,44 @@
+Name: libURLProcessCookies
+
+Type: command
+
+Syntax: libURLProcessCookies {true|false}
+
+Associations: internet library
+
+Introduced: 8.2
+
+OS: mac, windows, linux
+
+Platforms: desktop, server
+
+Security: network
+
+Example:
+libURLProcessCookies true
+
+Example:
+libURLProcessCookies false
+
+Description:
+<libURLProcessCookies> sets whether or not the libURL will transparently parse
+and store cookies sent from a server and then send those cookies back to the server
+with any requests.
+
+>*Important:*  The <libURLProcessCookies> <command> is part of the 
+> <Internet library> on desktop platforms. To ensure that the <command> 
+> works in a desktop <standalone application>, you must include this 
+> <LiveCode custom library|custom library> when you create your 
+> <standalone application|standalone>. In the Inclusions pane of the 
+> <Standalone Application Settings> window, make sure the "Internet" 
+> script library is selected.
+
+References: main stack (glossary), handler (glossary),
+Standalone Application Settings (glossary), message (glossary),
+group (glossary), standalone application (glossary), keyword (glossary),
+command (glossary), application (glossary),
+LiveCode custom library (glossary), URL (keyword),
+Internet library (library), library (library), startup (message),
+openBackground (message), preOpenStack (message), openStack (message),
+preOpenCard (message)
+
--- a/docs/notes/feature-liburl-cookies.md
+++ b/docs/notes/feature-liburl-cookies.md
@@ -1,0 +1,7 @@
+---
+version: 8.2
+---
+# Transparent cookie support in libURL
+
+libURLProcessCookies has been added to libURL. Passing in true for the first parameter 
+will turn on transparent cookie handling both from and to a server.
--- a/docs/notes/feature-liburl-cookies.md
+++ b/docs/notes/feature-liburl-cookies.md
@@ -1,6 +1,3 @@
----
-version: 8.2
----
 # Transparent cookie support in libURL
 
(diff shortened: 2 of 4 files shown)


trevordevore:
I created feature request 18468 for this. These are cookie parsing and internet date functions I've been using in a custom version of libURL for years.

docs/dictionary/command/libURLProcessCookies.lcdoc
```diff
@@ -0,0 +1,44 @@
+Name: libURLProcessCookies
+
+Type: command
+
+Syntax: libURLProcessCookies {true|false}
+
+Associations: internet library
+
+Introduced: 8.2
```
peter-b: Please change this to 9.0.

docs/notes/feature-liburl-cookies.md
```diff
@@ -0,0 +1,7 @@
+---
+version: 8.2
+---
```
peter-b: Please delete this header; the explicit `version` header should only be added when the release notes builder can't get the version right by examining the git history.  I added this to allow me to "fix up" release notes after the first DP they appeared in.

peter-b (changes requested):
@trevordevore Thank you very much for this patch.  I've retargeted this PR at the `develop` branch, because it adds new features rather than fixing known bugs.

Just to check: does libURLProcessCookies default to off?  I think this is quite important in order to make sure that existing code that assumes libURL ignores cookies doesn't start sending them out.  That could potentially cause an information disclosure vulnerability in user apps.

trevordevore:
@peter-b Yes, it defaults to off.

ide-support/revliburl.livecodescript
```diff
@@ -5290,6 +5309,345 @@ private function _proxyfromHTTPProxy pURL, pProxy
...
+
+   -- http://www.faqs.org/rfcs/rfc2965.html
+   -- Set-Cookie: <name>=<value>[; <name>=<value>]...
+   -- [; expires=<date>][; domain=<domain_name>]
+   -- [; path=<some_path>][; secure]
+   ## Cookie names are case insensitive
+   
+   ## Look for a redirect. If present use it as the host.
+   if line 1 of pHeader contains "302" then
+      put lineOffset("Location:", pHeader, tOffset) into tLineNo
```
peter-b, line 5394: Ideally you need to make sure that "location:" is located at the _start_ of the line.  I believe that `lineOffset()` will match anywhere within a line.  However, none of the rest of liburl seems to worry about this insignificant detail, so maybe you shouldn't either.

ide-support/revliburl.livecodescript
```diff
@@ -5290,6 +5309,345 @@ private function _proxyfromHTTPProxy pURL, pProxy
...
+   local tHost
+   local tKey,tMsg
+   local tLineNo
+   local tOffset
+   local tCookieDomain
+
+   -- http://www.faqs.org/rfcs/rfc2965.html
+   -- Set-Cookie: <name>=<value>[; <name>=<value>]...
+   -- [; expires=<date>][; domain=<domain_name>]
+   -- [; path=<some_path>][; secure]
```
peter-b, line 5389: I'm a little confused by this PR.  I looked at [RFC 2965](https://tools.ietf.org/html/rfc2965), but that RFC defines the `Set-Cookie2:` header.  I think you are trying to implement [RFC 2109](https://tools.ietf.org/html/rfc2109), which defines the `Set-Cookie:` header.

Could you please clarify?

Note that RFC 2965 will be obsoleted by [RFC 6265](https://tools.ietf.org/html/rfc6265), although the latter is still in draft status.

ide-support/revliburl.livecodescript
```diff
@@ -5290,6 +5309,345 @@ private function _proxyfromHTTPProxy pURL, pProxy
...
+               put cr & tKey & ":" && tCookieA[tKey] after tMsg
+            end repeat
+            ulLogIt tMsg & cr
+            next repeat
+         end if
+         
+         ## Check that tCookieA["domain"] is part of tHost
+         put tCookieA["domain"] into tCookieDomain
+         if char 1 of tCookieDomain is "." then delete char 1 of tCookieDomain
+         if char -length(tCookieDomain) to -1 of tHost is not tCookieDomain then next repeat
```
peter-b, line 5466: I'm finding it difficult to verify that this implements the whole of [RFC 2109 Section 4.3.2](https://tools.ietf.org/html/rfc2109#section-4.3.2) ("Rejecting Cookies").  I think it's quite important to make sure that the security related parts of the RFC are properly followed.  It would be really helpful if you could refactor this section of the code slightly or add comments to make it clearer how each of the rejection rules are being applied.

Ideally, there would be tests to check these elements but I'm not sure how we'd do that best.

ide-support/revliburl.livecodescript
```diff
@@ -5290,6 +5309,345 @@ private function _proxyfromHTTPProxy pURL, pProxy
...
+         if tCookieA["path"] is empty then put tDefaultPath into tCookieA["path"]
+         if tCookieA["expires"] is not empty then
+            ## Store expires in seconds
+            put _ConvertGMTInternetTimestampToInternetDate(tCookieA["expires"]) into tCookieA["expires"]
+            convert tCookieA["expires"] from internet date to seconds
+         end if
+         put "secure" is among the keys of tCookieA into tCookieA["secure"]
+         
+         ## Store each unique name=value pair on a separate line.
+         ## This allows us to easily update an existing cookie later on
```
peter-b, line 5478: Have you considered making `laCookies` a multi-dimensional array?  For example, `laCookies[<domain>]["expires"]`, `laCookies[<domain>]["path"]`, etc.  Do you think it would make the code simpler than having to encode the information in a string and then parse it again each time you want to check it?  I can see an argument either way, because at the moment you can put the contents of `laCookies[<domain>]` directly into a `Cookie:` header when making a request.

ide-support/revliburl.livecodescript
```diff
@@ -5290,6 +5309,345 @@ private function _proxyfromHTTPProxy pURL, pProxy
...
+
+function libURLGMTTimeStampToInternetDate pTimestamp
+   if char 1 of pTimestamp is an integer then
+      ## gmt timestamp
+      return _ConvertGMTTimestampToInternetDate(pTimestamp)
+   else
+      ## gmt internet timestamp
+      return _ConvertGMTInternetTimestampToInternetDate(pTimestamp)
+   end if
+end libURLGMTTimeStampToInternetDate
```
peter-b, line 5557: It would be great to have some tests for the time-related functions.  Did you figure out how to run the testsuite locally yet?

PR #4564, opened 2016-09-23
[Bug 18488] liburl: Report error from hostnametoaddress to the user

trevordevore:
ulHostNameToAddressCallback needs to report a detailed error rather thant "invalid host address". This fix reports the actual error.

### DIFF
--- a/ide-support/revliburl.livecodescript
+++ b/ide-support/revliburl.livecodescript
@@ -622,7 +622,7 @@ on ulGetFormat pUrl,pCount
       ## I.M. 2010-03-11  Need to check result and handle immediate failure where
       ##  no callback message will be sent
       if the result is not empty then
-         ulHostNameToAddressCallback tResolveHost, empty
+         ulHostNameToAddressCallback tResolveHost, empty, the result
       end if
       
       if laLoadReq[pUrl] is empty or  laAuthRecursCount[pUrl] > 0 then ##dc 170205
@@ -636,7 +636,7 @@ end ulGetFormat
 
 ################################################################################
 
-on ulHostNameToAddressCallback, pHostname, pIP
+on ulHostNameToAddressCallback pHostname, pIP
    local tURL, tOffset
    local tPre, tHost, tPort, tUser, tPass, tFilename, tURLHost, tURLPort
    local tConnectHost, tBadAddress
@@ -664,7 +664,11 @@ on ulHostNameToAddressCallback, pHostname, pIP
    end if
    
    if tBadAddress then
-      put "invalid host address" into laUrlErrorStatus[tURL]
+      if the paramCount >= 3 then
+         put param(3) into laUrlErrorStatus[tURL]
+      else
+         put "invalid host address" into laUrlErrorStatus[tURL]
+      end if
       if laLoadReq[tURL] then
          put "error" into laUrlLoadStatus[tURL]
          delete local laLoadingUrls[tURL]
@@ -712,7 +716,7 @@ on ulHostNameToAddressCallback, pHostname, pIP
       end if
    end if
    
-   ## AL-2013-07-30: [[ Bug 10796 ]]  HTTP "get URL" omits port number from HOST header
+   ## AL-2013-07-30: [[ Bug 10796 ]] HTTP "get URL" omits port number from HOST header
    ## delete this array at the end so that 'get url' can use the port and protocol data.
    subtract 1 from laURLFormat[tURL]["references"]
    if laURLFormat[tURL]["references"] is 0 then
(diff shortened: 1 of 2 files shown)


trevordevore:
@peter-b This still needs a bug report and unit test. I'm having trouble recreating a scenario where `hostnametoaddress` returns an error in `the result`. I had a customer whose network created this scenario at one point and this change is how I was able to a valid error report from libURL. I just can't remember what the scenario was :-)

What would be helpful is if someone on the LiveCode team could tell me under which circumstances `hostnametoaddress` returns an error in `the result`. Also, if someone can confirm that the callback handler passed to `hostnametoaddress` is not called if `the result` is set to a value. This information will help me create the bug report and test.

peter-b:
@trevordevore I think you should file a bug report anyway, even though it's not reproducible.  It helps me a lot to have a bug report for every work item, because it makes sure that things don't slip through the cracks.

What platform was your user on? It'll help me run down the possible code paths a bit faster.

From a brief glance at the code, it appears may be able to get `hostnameToAddress` to set `the result` if you try to resolve a string with garbage in as a hostname, especially something like `invalid:hostname`.

trevordevore:
@peter-b #18488. The issue was most likely on Windows (most issues I troubleshoot with libURL and my customers are on Windows due to more complex internet setups at larger orgs that use Windows). Being able to recreate the error on OS X is easier for me though.

peter-b:
@trevordevore Thanks for filing the bug report :)

peter-b:
@trevordevore Try this in the message box:

```
get hostNameToAddress("invalid:address")
put the result
```

I get "invalid host address" -- is that enough to be able to write a unit test?

trevordevore:
@peter-b The problem with a script like this:

```
set the sockettimeoutinterval to 1000
put hostnametoaddress("invalid:hostname") into theA
put the result
```

is that `the result` is always `invalid host address` which doesn't help us spot the error.

ide-support/revliburl.livecodescript
```diff
@@ -636,7 +636,7 @@ end ulGetFormat
 
 ################################################################################
 
-on ulHostNameToAddressCallback, pHostname, pIP
+on ulHostNameToAddressCallback pHostname, pIP
```
peter-b: Could you please put the optional parameter in the definition of the command here? For example:

```
on ulHostNameToAddressCallback pHostname, pIP, pOptionalError
```

ide-support/revliburl.livecodescript
```diff
@@ -664,7 +664,11 @@ on ulHostNameToAddressCallback, pHostname, pIP
    end if
    
    if tBadAddress then
-      put "invalid host address" into laUrlErrorStatus[tURL]
+      if the paramCount >= 3 then
+         put param(3) into laUrlErrorStatus[tURL]
```
peter-b: If you make the optional parameter explicit in the definition of the handler, then you can use `pOptionalError` instead of `param(3)` here.

peter-b (commented):
@trevordevore I had a look at this with @runrevmark today and we can confirm that `hostNameToAddress()` will never call the callback if it sets the result (and _vice versa_).

When you call `hostNameToAddress()` with a callback, there is only one way in which it can fail: if it can't set up the background thread that does the actual work of address resolution.  I therefore suspect the situation where your user experienced immediate failure of `hostNameToAddress()` in libURL may have been related to a problem on the local machine, rather than a problem with DNS resolution.

By the way, please feel free to go ahead and remove the compatibility code for Revolution < 4.5, if you feel like it!

We don't think it's likely to be possible to create a useful testcase for this, but let's get these changes merged anyway (without a unit test).

trevordevore:
@peter-b Changes have been made and a change log file added.

peter-b:
@trevordevore Thank you!

@livecode-vulcan review ok 0f8b74a

livecode-vulcan:
💙 review by @peter-b ok 0f8b74a908d00998db34c8627a64706068d913e1

livecode-vulcan:
😎 test success 0f8b74a908d00998db34c8627a64706068d913e1
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/300)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/305)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/300)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/300)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/306)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/300)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/300)

PR #4591, opened 2016-09-29
[[ Travis CI ]] Ensure cmd failure results in error

livecodeali:


livecodeali:
Peter's version might actually work. https://github.com/livecode/livecode/pull/4592

PR #4593, opened 2016-09-29
Use `new (nothrow)` everywhere

livecodefraser:
We jump through various hoops in order to get new to be recognised as
having nothrow behaviour but it is fragile and doesn't work when the
engine is hosted inside another process (if that process uses its own
implementation of new, as it is entitled to do).

Instead of dealing with this, just bite the bullet and add the
(std::nothrow) argument to every invocation of new in the engine and
externals.

### DIFF
--- a/revvideograbber/src/revvideograbber.cpp
+++ b/revvideograbber/src/revvideograbber.cpp
@@ -115,7 +115,7 @@ struct MYBITMAP
 
 MYBITMAP *createmybitmap(uint2 depth, uint2 width, uint2 height)
 {
-  MYBITMAP *image = new (std::nothrow) MYBITMAP;
+  MYBITMAP *image = new (nothrow) MYBITMAP;
   image->width = width;
   image->height = height;
   image->depth = (uint1)depth;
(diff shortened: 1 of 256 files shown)


revvideograbber/src/revvideograbber.cpp
```diff
@@ -626,7 +628,7 @@ void REVVideoGrabber(VideoGrabberKeyword whichkeyword,
                 // SN-2015-04-23: [[ Bug 15255 ]] Windows now defaults to directX
 #ifdef WIN32
 				if (_strnicmp(args[1], "vfw",strlen("vfw")) == 0)
-                    gvideograbber = new CWinVideoGrabber(windowid);
+                    gvideograbber = new (nothrow) CWinVideoGrabber(windowid);
```
peter-b, line 631: `std::nothrow`

revvideograbber/src/revvideograbber.cpp
```diff
@@ -647,7 +649,7 @@ void REVVideoGrabber(VideoGrabberKeyword whichkeyword,
                 }
 #ifdef WIN32
                 else
-                    gvideograbber = new CDirectXVideoGrabber(windowid);
+                    gvideograbber = new (nothrow) CDirectXVideoGrabber(windowid);
```
peter-b, line 652: `std::nothrow`

peter-b:
See #4606 for updated PR.

PR #4618, opened 2016-10-04
[[Dictionary]] exitField description display incomplete

seaniepie:
The term “<select command>” appeared to display a combo box and then
cause the rest of the script to fail. After checking other references,
it has been changed to “select <command>” as opposed to “<select>
command” which falls in line with ones such as closeField, openField
and scrollbarDrag.

### DIFF
--- a/docs/dictionary/keyword/codeunits.lcdoc
+++ b/docs/dictionary/keyword/codeunits.lcdoc
@@ -28,7 +28,7 @@ A codeunit refers to the encoded units in which codepoints are stored
 
 Use the expression 
 
-the number of codeunits of tString
+    the number of codeunits of tString
 
 to find out how many codeunits make up a Unicode string.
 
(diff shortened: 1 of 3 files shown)


peter-b:
@seaniepie This PR seems to contain three changes, rather than just the one mentioned in the PR description and title.  You may need to use the `git` command-line tools to fix this (it should be pretty straightforward to do so).  If you need some help with this, please let me know and I'll do my best to guide you through it.

@livecode-vulcan review not ok ee9209e

livecode-vulcan:
💔 review by @peter-b not ok ee9209e31247b945156837abdcf187a533600379

seaniepie:
Yeah, I broke it Peter :/ I was doing the fixes in the desktop app but the pull didn't seem to work so I then started copying them over manually to the web git interface. I tried to be 'clever' part way through the process and this is what happened so I stopped. I will PM for advise, thanks.

PR #4619, opened 2016-10-04
[[Glossary]] object.lcdoc include control & widget

seaniepie:
To make it easier to locate object types and references

### DIFF
--- a/docs/glossary/o/object.lcdoc
+++ b/docs/glossary/o/object.lcdoc
@@ -6,9 +6,13 @@ Type: glossary
 
 Description:
 A LiveCode module capable of receiving <message|messages>. One of the
-building blocks of a LiveCode application.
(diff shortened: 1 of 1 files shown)


livecodeali:
@seaniepie 
How about making it say something like

```
"... building blocks of a LiveCode application. Objects include <stack|stacks> and <card|cards>, as well as any <control>.
```

(and make sure card (glossary) and stack (glossary) are in the references)
Then you could also change the control glossary to include the list of control types.

The ones in control.lcdoc would probably be best rendered as a bulleted list or something. A widget is a control, so it should probably be added to that list and not as an explicit example of an object.

seaniepie:
I had destroyed the repo in my early stages of learning my way around so couldn't seem to edit. Have moved it over to [#4736](https://github.com/livecode/livecode/pull/4736) taking on @livecodeali comments

PR #4620, opened 2016-10-04
[[Dictionary]] Abbreviated.lcdoc - missing 'set'

seaniepie:
Description missing the 'set' command


peter-b:
@seaniepie It would be great to turn "set" into a reference to the `set` command's dictionary entry!

livecodeali:
@seaniepie Also this has `set command (command)` in the references which needs to be removed (or changed to `set (command)` if you go with Peter's suggestion)

seaniepie:
I can't reopen it to edit. I'm out of my depth with GitHub

peter-b:
Oh dear... it looks like you've deleted something and now it's showing this PR as coming from an `unknown repository`.  GitHub is broken. 😭   On the bright side, it looks like this PR is replaced by #4634, so I'm going to go ahead and close it.

PR #4632, opened 2016-10-05
[[Dictionary]] exitField description incomplete

seaniepie:
The term “<select command>” appeared to display a combo box and then
cause the rest of the script to fail. After checking other references,
it has been changed to “select <command>” as opposed to “<select>
command” which falls in line with ones such as closeField, openField
and scrollbarDrag.


seaniepie:
Github is being a git and keeps putting multiple commits into one pull request. I'll try again. Again!

PR #4677, opened 2016-10-06
Revert "Merge pull request #4515 from montegoulding/unreachable-code"

livecodeali:
Not to be merged, just to satisfy my curiosity.


montegoulding:
@livecodeali do you think that helped?

livecodeali:
@montegoulding it does appear that this PR is succeeding consistently (I have only tried it twice though)

montegoulding:
@livecodeali Hmm.... so the mac failure is something else I guess... 

BTW looking at the code again I can't help thinking that the bug is the `break` shouldn't have been there at all, however, I don't think that does anything to help our situation...

PR #4695, opened 2016-10-06
Enforce indentation with tabs & alignment with spaces

peter-b:
This patch introduces some infrastructure for ensuring a whitespace
style where code is indented with horizontal tab characters and
aligned with spaces.

It relies on the `uncrustify` tool [1].  This is packaged for
Ubuntu and Fedora, can be installed on MacOS using `homebrew`, and
there are Windows binaries available too.

This patch adds `indent` and `check-indent` targets to the
top-level Makefile, which apply indentation rules and verify them,
and adds a Travis CI rule which fails the build if there are any
indentation mismatches.

Initially, _all_ existing source files are blacklisted from
having indentation rules applied, apart from libfoundation-system
and the Emscripten-specific source code.

[1] http://uncrustify.sourceforge.net/


montegoulding:
@peter-b do you know if there's a way to integrate this step into git a hook?

montegoulding:
Hmm... I think the best we can do is reject in a pre-commit hook if check-indent fails although that's not ideal as you may not actually have the broken indents staged...

peter-b:
This is now moot -- the agreed-upon coding style is now indentation with spaces only.

PR #4720, opened 2016-10-08
Some coverity defect fixes

montegoulding:


engine/src/hc.cpp
```diff
@@ -1998,10 +1998,16 @@ IO_stat MCHcstak::read(IO_handle stream)
...
 				{
 					if (IO_read_uint1(&byte, stream) != IO_NORMAL)
+					{
+						delete[] tbuffer;
 						return IO_ERROR;
+					}
 					while ((byte == '\n' || byte == '\r') && tsize--)
 						if (IO_read_uint1(&byte, stream) != IO_NORMAL)
+						{
+							delete[] tbuffer;
```
peter-b, line 2008: Is it feasible to turn `tbuffer` into some sort of `MCAutoFoo` so as to get automatic cleanup?

engine/src/paragrafattr.cpp
```diff
@@ -1225,6 +1225,8 @@ void MCParagraph::computerects(int32_t x, int32_t y, int32_t p_layout_width, uin
 		case kMCParagraphTextAlignRight:
 			t_offset = p_layout_width - t_total_width;
 			break;
+		default:
+			MCUnreachableReturn();
```
peter-b, line 1229: @montegoulding I don't think this is enough, unfortunately.  In Release builds, this will be compiled out, so won't catch the error.  There's nothing in the field's implementation that provides compile-time assurance that the return value of `gettextalign()` will be a member of the (anonymous) text alignment enumeration.  I believe that the "right" fix here is to leave the switch how it is, and initialise `int32_t t_offset = 0;`. Possible alternatives would be to explicitly make the left alignment case the `default:` case, or set `t_offset = 0;` just before `MCUnreachabelReturn()`.

engine/src/image.cpp
```diff
@@ -1335,95 +1335,123 @@ IO_stat MCImage::load(IO_handle stream, uint32_t version)
...
+					t_compressed->height = t_pixheight;
+				}
 			}
-			t_compressed->width = t_pixwidth;
-			t_compressed->height = t_pixheight;
-
-			/* UNCHECKED */ setcompressedbitmap(t_compressed);
+			
+			if (IO_NORMAL == stat)
+				/* UNCHECKED */ setcompressedbitmap(t_compressed);
```
peter-b: This patch massively improves this method implementation.  Do you mind just adding this last little check for me, please? 😉

montegoulding: I considered it but I couldn't work out what should happen in the event of a failure here.

engine/src/paragrafattr.cpp
montegoulding: I will init it to 0.

engine/src/hc.cpp
montegoulding: Maybe.. will look into it

montegoulding: @peter-b after looking at this I'd rather avoid touching this code other than to fix the leak as I've done. I don't think it's worth the time or risk of injecting a bug into HC stack import. I did silence a few warnings that were close by ;-)

engine/src/image.cpp
montegoulding: I decided it's probably safe to set stat to IO_ERROR here. There's one more unchecked result in this method but as it's setting the filename I don't think the whole method should error if it fails so I'll leave that for now.

peter-b:
@montegoulding Thanks for making those tweaks!

@livecode-vulcan review ok a74c0bf

livecode-vulcan:
💙 review by @peter-b ok a74c0bf1866be62f96ba16fb50132ba871c95000

livecode-vulcan:
😎 test success a74c0bf1866be62f96ba16fb50132ba871c95000
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/388)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/393)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/388)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/388)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/394)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/388)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/388)

PR #4723, opened 2016-10-09
[[Docs]] Keyword lcdoc's description and syntax

seaniepie:
This brings 4 keyword docs up to the correct standard. Three of them had no Description section and one had regular brackets instead of curly brackets in the Syntax. TravisCL should no longer bring up issues with these files. Finally, exitField had relic html tags and incorrectly set references

### DIFF
--- a/docs/dictionary/keyword/colorPalette.lcdoc
+++ b/docs/dictionary/keyword/colorPalette.lcdoc
@@ -13,6 +13,9 @@ OS: mac, windows, linux
 
 Platforms: desktop, server
 
-References: reserved word (glossary), colors (property),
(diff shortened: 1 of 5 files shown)


livecodeali:
@seaniepie This PR has the unrelated changes to exitField in it too. Perhaps the easiest thing to do would be to create a fresh branch based on develop and do 

```
git cherry-pick 95726d1..dcbd20c
```

and that should apply the last three commits to the branch (note when you cherry-pick a series, the first one is not itself actually cherry-picked)

seaniepie:
Moved this to request #4733

PR #4725, opened 2016-10-10
[[Docs]] lcdoc syntax and param fixes

seaniepie:
- Fixes to Control Structures, Operators and Messages Syntax and Parameters
  - Should help Travis parse and validate these docs

### DIFF
--- a/docs/dictionary/control_st/if.lcdoc
+++ b/docs/dictionary/control_st/if.lcdoc
@@ -41,7 +41,7 @@ statement.
 Description:
 Use the <if> <control structure> to <execute> a <statementList|statement> 
 (or list of <statementList|statements>) only under certain circumstances. 
-#####Form: 
+
 The <if> <control structure> always begins with the <word> if. There are four forms of the <if> <control structure>:
 
     if condition then statementList [else elseStatementList]
--- a/docs/dictionary/control_st/if.lcdoc
+++ b/docs/dictionary/control_st/if.lcdoc
@@ -2,7 +2,7 @@ Name: if
 
 Type: control structure
 
-Syntax: if <condition> then <statement> [else <elseStatement>]
+Syntax: if <condition> then <statementList> [else <elseStatementList>]
 
 Summary:
 <execute|Executes> a list of <statement|statements> if a condition is
@@ -29,50 +29,54 @@ Parameters:
 condition (bool):
 
 
-statement:
-
-
-elseStatement:
+elseStatementList:
 
 
 statementList:
-The statementList or elseStatementList consists of one or more LiveCode
+The <statementList> or <elseStatementList> consists of one or more LiveCode
 statements, and can also include if, switch, try, or repeat control
 structures. The statement or elseStatement consists of a single LiveCode
 statement. 
 
 Description:
-Use the <if> <control structure> to <execute> a <statement> (or list of
-<statement|statements>) only under certain circumstances. Form:The <if>
+Use the <if> <control structure> to <execute> a <statementList|statement> 
+(or list of <statementList|statements>) only under certain circumstances. Form:The <if>
 <control structure> always begins with the <word> if. There are four
 forms of the <if> <control structure>:
 
-if <condition> then <statement> [else <elseStatement>]
+```
+if condition then statementList [else elseStatementList]
+```
 (This form may have a line break before the words then or else or both.)
 
-if <condition> then
-<statementList> 
+```
+if condition then
+   statementList 
 [else
-elseStatementList]
+   elseStatementList]
 end if
-
-if <condition> 
-then <statement> 
+```
+```
+if condition 
+then statement 
 [else
-elseStatementList]
+   elseStatementList]
 end if
+```
+```
+if condition then
+   statementList 
+else elseStatement
+end if
+```
 
-if <condition> then
-<statementList> 
-else <elseStatement> 
-
-If the <condition> <evaluate|evaluates> to true, the <statement> or
+If the <condition> <evaluate|evaluates> to true, the statement or
 <statementList> is <execute|executed>; if the <condition>
-<evaluate|evaluates> to false, the <statement> or <statementList> is
+<evaluate|evaluates> to false, the statement or <statementList> is
 skipped. 
 
 If the <if> <control structure> contains an <else> clause, the
-<elseStatement> or elseStatementList is <execute|executed> if the
+elseStatement or <elseStatementList> is <execute|executed> if the
 <condition> is false.
 
 If one <if> <control structure> is nested inside another, use of the
@@ -82,8 +86,8 @@ cause ambiguities in interpreting which <else> clause belongs with which
 
 The <if> <control structure> is most suitable when you want to check a
 single <condition>. If you need to check for multiple possibilities,
-doing something different for each one, use a <switch> <control
-structure> instead.
+doing something different for each one, use 
+a <switch> <control structure> instead.
 
 >*Note:* The <if> <control structure> is implemented internally as a
 > <command> and appears in the <commandNames>.
--- a/docs/dictionary/control_st/if.lcdoc
+++ b/docs/dictionary/control_st/if.lcdoc
@@ -40,35 +40,32 @@ statement.
 
 Description:
 Use the <if> <control structure> to <execute> a <statementList|statement> 
-(or list of <statementList|statements>) only under certain circumstances. Form:The <if>
-<control structure> always begins with the <word> if. There are four
-forms of the <if> <control structure>:
+(or list of <statementList|statements>) only under certain circumstances. 
+#####Form: 
+The <if> <control structure> always begins with the <word> if. There are four forms of the <if> <control structure>:
 
-```
-if condition then statementList [else elseStatementList]
-```
-(This form may have a line break before the words then or else or both.)
+    if condition then statementList [else elseStatementList]
 
-```
-if condition then
-   statementList 
-[else
-   elseStatementList]
-end if
-```
-```
-if condition 
-then statement 
-[else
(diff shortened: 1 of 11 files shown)


docs/dictionary/control_st/repeat.lcdoc
```diff
@@ -5,6 +5,22 @@ Type: control structure
 Syntax:     repeat <loopForm>
       <statementList>
     end repeat
+    
+Syntax:  repeat {forever | while <condition> | until <condition>} 
+      <statementList> ; end repeat
```
montegoulding: @seaniepie using `;` could be a point of confusion for people so it might be better to move it to its own line. It might be better to remove the original `loopForm` syntax here and delete the associated parameter. Perhaps @livecodeali and or @asayd might want to chime in on that though.

docs/dictionary/message/menuPick.lcdoc
```diff
@@ -2,7 +2,7 @@ Name: menuPick
 
 Type: message
 
-Syntax: menuPick <chosenItem> [| <submenuName>], <previousTab>
+Syntax: menuPick <chosenItem> {| <submenuName>}, <previousTab>
```
montegoulding: @seaniepie I think `submenuName` should be deleted as the pipe delimited submenus are documented in `chosenItem`. `previousTab` should also be in square brackets as it only applies to tab buttons.

seaniepie: I see. 👍

docs/dictionary/control_st/repeat.lcdoc
seaniepie: Wait to hear from @livecodeali

docs/dictionary/message/menuPick.lcdoc
seaniepie: Done

docs/dictionary/control_st/repeat.lcdoc
livecodeali: Yeah I agree with @montegoulding on both counts!

docs/dictionary/control_st/repeat.lcdoc
```diff
@@ -6,20 +6,24 @@ Syntax:     repeat <loopForm>
       <statementList>
     end repeat
     
-Syntax:  repeat {forever | while <condition> | until <condition>} 
+Syntax:     repeat {forever | while <condition> | until <condition>} 
       <statementList> ; end repeat
```
livecodeali: There's still a semicolon here. Also you can delete the 

```
repeat <loopForm>
    <statementList>
end repeat
```

as it's now covered by all the other forms, and delete the `loopForm` parameter description

asayd:
@seaniepie In the Syntax section the param name for the repeat with form is counterVariable, but in the Params section it is counter. They should match.

asayd:
@seaniepie Also note that in the Description section under With counterVariable there is a phrase about "the _counter_ being increased by (line break here) 1." In the Dictionary the 1. on a new line renders as an ordered list. Probably need to break the line somewhere else so that won't happen. Or just spell out "increased by one."

seaniepie:
@asayd, @livecodeali , is this one ready for a merge now?

livecodeali:
@seaniepie The code parts of the Description in the if doc should be tab-indented rather than surrounded by triple-backticks.

I still think the 

```
repeat <loopForm>
    <statementList>
end repeat
```

should be deleted, the loopForm parameter deleted and references to <loopForm> changed to just loop form, but I'd be interested to hear if anyone disagreed!

seaniepie:
I'll wait for more comments on this till I modify the file to remove the triple backticks and review the loopform.

seaniepie:
How about, in an effort to make it more useful and understandable to both newcomers and veterans alike to remove the loopForm from both the Syntax & Params and then to consolidate all the syntax examples into one unit but with comment dividers like:

```
Syntax:
// Repeats until a condition is met:
repeat {forever | while <condition> | until <condition>} 
    <statementList>
end repeat

// Repeats a set number of times:
repeat [for] <number> [times]
    <statementList>
end repeat

// Repeats from one number to another as a variable: 
repeat with <counterVariable> = <startValue> {to | down to} <endValue> [step <increment>]
    <statementList>
end repeat

// Repeats in the listed order for all chunks of the string or sequence as a variable:
repeat for each <chunkType> <labelVariable> in <container>
    <statementList>
end repeat

// Repeats for each element or key of an array as a variable:
repeat for each {element | key} <labelVariable> in <array>
    <statementList>
end repeat
```

asayd:
@seaniepie I think including the various forms in the syntax element is a good idea, assuming the rendering routines handle it properly. I believe that multiple Syntax elements are simply rendered as a multiple variants in the dictionary's Syntax section. (Is that right, @livecodeali ?) So it may not make any difference how you enter it.

Also shouldn't this:
    repeat {forever | while <condition> | until <condition>} 

Be this:
    repeat [forever | while <condition> | until <condition>]

??
A repeat loop does not _require_ a termination condition. If nothing is there, `repeat forever` is assumed.

seaniepie:
I've checked the formatting for Syntax and everything appears well as long as I replace slashes for hyphens.

If a pipe is used to infer OR then I thought curly brackets were required although the dictionary preview stack doesn't seem to mind it. It's very interesting the bit about termination and the assumption of `repeat forever`. With this the case, it is more appropriate to use `repeat [forever | while | until]`. 

And while we remove <loopForm> from the parameters and syntax we would probably still want this in the Description?? :

```
####Loop Form:
The loop form is one of the following forms:

  - `forever`
  - `until `*`condition`*
  - `while `*`condition`*
  - `[for] `*`number`*` [times]`
  - `with `*`counterVariable`*` = `*`startValue`*` {to | down to}
    `*`endValue`*` [step `*`increment`*`]`
  - `for each `*`chunkType`* *`labelVariable`*` in `*`container`*
  - `for each element `*`labelVariable`*` in `*`array`*
  - `for each key `*`labelVariable`*` in `*`array`*

The <repeat> <control structure> always begins with the `repeat` <keyword>.
The last line of a <repeat> <control structure> is always the `end repeat` <keyword>.

How many times the <statementList> is <execute|executed> depends on the
loop form you use. Here are details of the possible forms:
```

is this correct @livecodeali ?

livecodeali:
@seaniepie Yes I think that would be best!

livecodeali:
@seaniepie Sorry, I should have replied to that suggestion about consolidating the `Syntax:` elements - my view is that the syntax should be a pretty strict representation of what is actually the syntax, and so shouldn't have comments in (even if it renders ok). Moreover since these are variants rather than continuations of the same syntax, they should each have their own `Syntax:` element like they did before.

seaniepie:
No problem 😄

seaniepie:
Question: Is it basically only in straight Examples then that we have comments? The suggestion was to make it easier for users to understand at a glance which one does what. I understand, though, the reason for keeping the syntax as clean as possible as it is really ONLY for 'syntax'.

livecodeali:
@seaniepie It's fine for any code blocks to have comments, but I guess I'm not really thinking of the syntax element as a code block- indeed it has various non-code elements so that's probably for the best. @asayd and I have discussed a bit about the presentation of the syntax, initially thinking it would be better rendered as code; but given that (as @asayd pointed out) the parameters now don't get italicised, and that it isn't actually code, that might have been a mistake!

livecodeali:
@seaniepie Ok, I've just noticed in the `if` doc, there are still code blocks done with triple backticks rather than indenting.

I think once these are done then this PR is good to go 😄

seaniepie:
@livecodeali I've changed the layout of the Form section a little more than just the code blocks as I hadn't noticed before that there was a heading for 'Form:'.

docs/dictionary/control_st/if.lcdoc
```diff
@@ -40,35 +40,32 @@ statement.
...
-end if
-```
-```
-if condition then
-   statementList 
-else elseStatement
-end if
-```
+(or list of <statementList|statements>) only under certain circumstances. 
+#####Form:
```
montegoulding: @seaniepie I think this can just be removed. At some point we will probably change lcdoc to use `#Element` headings instead of `Element:` so it is probably better not to use `#` at all until we do that. I don't really thing `Form:` adds anything to the description here.

seaniepie: 👍

montegoulding:
@livecode-vulcan review ok ccd1e4e

livecode-vulcan:
💙 review by @montegoulding ok ccd1e4eedb1203ef6a869d1894d3c569b6485a85

livecode-vulcan:
😎 test success ccd1e4eedb1203ef6a869d1894d3c569b6485a85
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/486)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/491)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/486)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/486)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/492)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/486)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/486)

montegoulding:
Thanks @seaniepie !!! 👍

seaniepie:
😄 Got there finally.

montegoulding:
I'm really looking forward to merging the branch with the extra docs tests!

PR #4759, opened 2016-10-14
Make sure android externals are included in the standalone

livecodepanos:
`tExternals` must **not** include the resolved path, since later the script does:

```
-- Copy across the externals into the libs folder
      repeat for each line tEntry in tExternals
         ...
         put url ("binfile:" & mapFilePath(revMobileRuntimeFolder(pTarget) & slash & tEntry)) into url ("binfile:" & tArmLibsBuildFolder & slash & "lib" & tEntry & ".so")
         // SN-2015-01-22: [[ Bug 13213 ]] Throw an error in case an error occured
         if the result is not empty then
            throw "Could not find " & tEntry && "library"
         end if
      end repeat
```


livecodeali:
@livecode-vulcan review ok ccc49de

livecode-vulcan:
💙 review by @livecodeali ok ccc49de72608c082635d085b9f94dd2efd25b64f

PR #4767, opened 2016-10-17
[Bug 18617] Implement OAuth2 dialog/library

montegoulding:


extensions/script-libraries/oauth2/oauth2.livecodescript
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+
+Present an authorization dialog for any web service that supports OAuth2 Authorization Code Flow
+
+Description:
+In order to handle the redirect the library accepts socket connections on localhost on a
+configurable port. The redirect uri configured when setting up your application with the
+web service should be `http://localhost:<port>` where `<port>` is the port that can be
+configured with `OAuth2Set "port",<port>`. The default port is 8080.
```
peter-b: @montegoulding I recommend removing `OAuth2Set` and the `sport` local variable, and instead always using a random port in the range 49152-65535 (this is [the range assigned for dynamic ports](https://tools.ietf.org/html/rfc6335#section-6)).

Doing this should allow calling `OAuth2()` while another `OAuth2()` call is still waiting to complete.

montegoulding: Hmm.... ill test to see if you can specify the redirect uri on the services I've tested against without a port.

montegoulding: Facebook was the first one I tested as it appears to be the strictest. It didn't like specifying the redirect uri in the app settings without a port then calling it with one.

screen shot 2016-10-18 at 7 42 49 am

montegoulding: @peter-b with regard to multiple dialogs perhaps the simplest is if there is already a dialog open then `return "dialog open" for error` or something...

peter-b: @montegoulding I'm not sure what you mean by "specifying the redirect URI in the app settings".

montegoulding:
When you setup an application for a web service you often need to specify a redirect uri in their settings. This must match the one you redirect to.

peter-b:
@montegoulding Okay, that makes sense.  Sorry I didn't realise this! :-) I still recommend using a private use area port as the default.

montegoulding:
OK, do you think I should force it in that range?

peter-b:
@montegoulding No, I don't think so -- need to keep options open to users. I recommend using port in private use area as the default and recommending the private use range to users in the docs, but if they want to choose to use port 1 then good luck to them :)

peter-b:
@montegoulding I also think it's a good idea to make sure that either multiple simultaneous OAuth sessions work or that the library prevents them by throwing an error.

montegoulding:
@peter-b OK, I'll look into what can be done. It's feasible I could use the state parameter which is just a uuid that gets passed back and forth to work out which dialog the socket is handling and which stack to close etc.

montegoulding:
@peter-b should this actually have a release note and be included in the IDE & loaded at this stage or should it just sit in the repo until we want to do something in the IDE with it.

peter-b:
@montegoulding This is a pretty neat little feature. Let's do the full works (give it a detailed release note, ship it with the IDE, load it by default and make sure it can be included with standalones)!

montegoulding:
@peter-b I think I need to put some things in place to generate lcdoc and include it in the dictionary. I also need to document security issues around app secrets.

montegoulding:
The docs extraction PR doesn't depend on but is closely related to https://github.com/livecode/livecode-ide/pull/1457

montegoulding:
Have added need-tests because the extracted docs should be tested

montegoulding:
@peter-b does the release note go in `extensions/script-libraries/oauth2/notes/` or in `docs/notes` ?

peter-b:
The former, please!

montegoulding:
Trying to work out why the extracted docs tests don't appear to be finding the directory. I note that there's a COPY of everything but `extracted_docs`...

peter-b:
Could it be because script only extensions don't have a manifest?

montegoulding:
hmm... I was thinking something like write privileges to the products dir... how would the manifest impact it?

montegoulding:
I expect so... it doesn't appear to be that anyway as the file exists... I'll throw some diagnostics into the test...

montegoulding:
And... I forgot there's no way to get diagnostics from travis...

montegoulding:
Yep.. files are there when it extracts docs but the folder doesn't exit when it's doing the test... I'm suspicious of the section where everything appears to be copied but my new output directory... I think I need to do something to get mine copied there.

```
  COPY ../../linux-x86_64-bin/server-revzip.so
  COPY ../../linux-x86_64-bin/revpdfprinter.so
  COPY ../../linux-x86_64-bin/modules
  COPY ../../linux-x86_64-bin/packaged_extensions
```

montegoulding:
Yay... found the required gyp incantation:

```
'all_dependent_settings':
            {
                'variables':
                {
                    'dist_aux_files':
                    [
                        # Gyp will only use a recursive xcopy on Windows if the path ends with '/'
                        '<(PRODUCT_DIR)/extracted_docs/',
                    ],
                },
            },
```

montegoulding:
Retrying travis after an IDE test crash

engine/engine.gyp
```diff
@@ -22,6 +22,88 @@
...
+				{
+					'action_name': 'extract_docs_from_stacks',
+					'message': 'Extracting docs from stacks',
+					
+					'inputs':
+					[
+						'../util/extract-docs.livecodescript',
+						'../ide-support/revdocsparser.livecodescript',
+						'<@(_sources)',
+					],
```
peter-b, line 87: I think you need to list `'<(engine)'` in the `inputs` so that extracting the documentation gets ordered after compiling the engine by the build process.

peter-b:
I remain a little concerned over using a predictable port, but if the commonly-used server-side implementations of OAuth require it then there's not much we can do... 😞 

@livecode-vulcan review ok c102dc5

livecode-vulcan:
💙 review by @peter-b ok c102dc5bd21c62726744a299bd7bf890ed13d789

livecode-vulcan:
😎 test success c102dc5bd21c62726744a299bd7bf890ed13d789
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/499)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/504)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/499)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/499)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/505)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/499)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/499)

engine/engine.gyp
montegoulding: Hmm... I was fairly sure the server dependency would do that

montegoulding:
@peter-b when/if we support accepting on random ports properly in the engine we can add that as an option here for the web services that don't force you to enter a redirect uri into the app config.

PR #4805, opened 2016-10-25
[Bug 18642] Fix crash on iOS 10 when calling mobileGetNotificationDetails

livecodepanos:
1. Using `mobileCreateLocalNotification alertBody, alertButtonMessage, alertPayload, alertTime, playSound [, badgeValue]` successfully creates a local notification on both iOS 9 and 10.
2. The LC param `alertButtonMessage` corresponds to `[t_local_notification alertAction]`
3. Setting this param works as expected on both iOS 9 and 10.

However, getting this param on iOS 10 ( `[t_local_notification alertAction]`) returns `nil`, which causes a crash when calling `mobileGetNotificationDetails`

This patch fixes the crash. However, if one examines the array returned by `mobileGetNotificationDetails`:

iOS 9: `alertButtonMessage = myAlertButtonMessage`
iOS 10: `alertButtonMessage = ""`

### DIFF
--- a/docs/notes/bugfix-18642.md
+++ b/docs/notes/bugfix-18642.md
@@ -1,0 +1,1 @@
+# Fixed crash on iOS 10 when trying to read local notifications
--- a/engine/src/mbliphonenotification.mm
+++ b/engine/src/mbliphonenotification.mm
@@ -119,9 +119,20 @@ bool MCSystemGetNotificationDetails(int32_t p_id, MCNotification &r_notification
         UILocalNotification* t_local_notification = [t_scheduled_local_notifications objectAtIndex:i];
         if (atoi ([[t_local_notification.userInfo objectForKey:@"notificationId"] cStringUsingEncoding:NSMacOSRomanStringEncoding]) == p_id) 
         {
-            MCStringCreateWithCFString ((CFStringRef)[t_local_notification alertBody], r_notification.body); // or ""
-            MCStringCreateWithCFString ((CFStringRef)[t_local_notification alertAction], r_notification.action); // or ""
-            MCStringCreateWithCFString ((CFStringRef)[t_local_notification.userInfo objectForKey:@"payload"], r_notification.user_info);
+			if ([t_local_notification alertBody] == nil)
+				r_notification.body = MCValueRetain(kMCEmptyString);
+			else
+				MCStringCreateWithCFString ((CFStringRef)[t_local_notification alertBody], r_notification.body); // or ""
+			
+			if ([t_local_notification alertAction] == nil)
+				r_notification.action = MCValueRetain(kMCEmptyString);
+			else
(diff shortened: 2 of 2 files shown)


montegoulding:
@livecodepanos I wonder if `MCStringCreateWithCFString` should do the check for nil? Seems quite similar to the problem in #4806

peter-b:
@montegoulding What's the usual pattern with Objective C APIs?  Do they usually accept `nil` for strings?

montegoulding:
No, but this is a conversion of nil -> LC value. The main reason I suggested it is because we either need to remember to do what @livecodepanos has done here on APIs that might return nil or Apple might decide one day to change to occasionally return nil or we just handle the case in one location and always convert nil to the empty string.

peter-b:
@montegoulding My feeling is that "the argument is a valid `CFString`" is a necessary precondition for calling `MCStringCreateWithCFString`, and should be asserted.

In the long run we could probably have something similar to [`g_return_val_if_fail()`](https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-val-if-fail) which aborts in debug builds and prints a critical message and attempts to stagger along without killing the program in release mode

montegoulding:
OK then

peter-b:
@montegoulding I wish C++ had a type which was something along the lines of "a pointer that is not allowed to ever be null", but sadly it doesn't :-(

peter-b:
@montegoulding @livecodepanos On further investigation, it looks like the reason iOS is handing us a null pointer here instead of a valid `CFString` is because in iOS 10 the `UILocalNotification` API is deprecated, and one is supposed to use the `UNLocalNotification` API instead.  My guess is that we're getting null pointers because the values we gave iOS when setting up the notification aren't relevant to the new system and are just getting thrown away by iOS.

I will open a new bug report for updating to the `UNLocalNotification` API.

@livecode-vulcan review ok 07b3fcf

livecode-vulcan:
💙 review by @peter-b ok 07b3fcff2bdfd2a9b3c23f807aee656589c0398d

livecode-vulcan:
😎 test success 07b3fcff2bdfd2a9b3c23f807aee656589c0398d
- try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/502)
- try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/507)
- try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/502)
- try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/502)
- try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/508)
- try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/502)
- try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/502)

PR #4826, opened 2016-10-29
Turn on -Werror=writable-strings

montegoulding:


peter-b:
@montegoulding I'm not sure this is a good idea.  Several places in this patch are exposing places where APIs should take `const char*` argument rather than `char *` arguments, and I feel like adding casts to remove `const` is really just papering over the cracks rather than actually fixing the problem that the compiler is complaining about.

I personally would be much happier with doing a minor refactor in each of these places to remove the writable-pointer-to-immutable-string entirely, and only _then_ adding `-Werror=writable-strings`. I'm not totally sure what the best option is here, though... maybe what's needed is a second opinion.

/cc @livecodefraser

montegoulding:
@peter-b fair enough... There's probably a mix of things that should be refactored here and things that are OK being cast. `replaceString` can probably just be changed to `const char *` params as it is only called in one place. Some things it isn't worthwhile to refactor like the `AddError` in video grabber. Some things I'm not really sure about like initing `DBString` with an empty string....

PR #4840, opened 2016-11-03
[Bug 18765] Implement script library for common MIME functions

montegoulding:


peter-b:
@montegoulding This looks neat!  Is there any reasonable way to store the mimetype info in a separate file that's either loaded at run time or used in a build step to generate the script-only stack?  It would make it easier to update the data in the future.

peter-b:
@montegoulding Also, could you please file a tracking ticket in Bugzilla for this enhancement?

montegoulding:
@peter-b sure I was planning to do the bug once I gauged if people thought this was worthwile making generally available. Basically I need this stuff in what I'm working on so...

There will be a bit of work involved in getting libraries to travel around and into standalones with resource files and I doubt we have resources for that right now so a separate file is not as easy as it seems.

I wonder if we need a general update stuff automation moomin and this could plug into that? It would be easy enough to add some markers into the file so it could be swapped out with the latest.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+   if tMIMEType is empty then
+      return "application/octet-stream" for value
+   else
+      return tMIMEType for value
+   end if
+end MIMETypeForExtension
+
+/**
+
+List extensins that match a MIME type or MIME type wildcard pattern
```
peter-b: "extensins" -> "extensions"

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+      pass revUnloadLibrary
+   end if
+   
+   remove the script of me from back
+end revUnloadLibrary
+
+/**
+
+This script library implements portions of Multipurpose Internet Mail Extensions (MIME)
+as described in RFCs 2045 - 2049
```
peter-b: It might be helpful to add a brief explanation of what sort of things this library will help you with.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+Example:
+put "MIME-Version:" && MIMEVersion() & crlf after tHeaders
+
+Description:
+Use the <MIMEVersion> function create a `MIME-Version` header.
+
+*/
+
+function MIMEVersion
+   return 1.0
```
peter-b: Consider returning a string rather than a floating-point number here.  I'm aware that there's currently no distinction in LiveCode Script when it comes to literals (they're all "names") but semantically this isn't a number, it's two period-delimited numbers.

Please name this function with a verb.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+
+function MIMETypeForExtension pExtension
+   local tMIMEType
+   put sMIMETypes[pExtension] into tMIMEType
+   if tMIMEType is empty then
+      return "application/octet-stream" for value
+   else
+      return tMIMEType for value
+   end if
+end MIMETypeForExtension
```
peter-b: Is it feasible to get this to first do a case-sensitive lookup of the extension, and then a case-insensitive one if the first failed?  Some people have, in the past, thought it would be a great idea to use `.C` for C++ source files and `.c` for C source files... 😖 

Please name this function with a verb.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+   switch pEncoding
+      case "quoted-printable"
+         return __EncodeQuotedPrintable(pData) for value
+         break
+      case "base64"
+         return __EncodeBase64(pData) for value
+         break
+   end switch
+   
+   return "unsupported encoding" for error
```
peter-b: Perhaps it would be clearer to move this into a `default:` clause in the `switch` above.

In LiveCode, is it acceptable coding style to omit the `break` from a switch case if it ends with a `return`?

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+
+Returns:
+The encoded content
+
+References:
+textEncode (function)
+
+*/
+
+function MIMEEncode pEncoding, pData, pTextEncoding
```
peter-b: Consider re-ordering the arguments to this function.  It feels a bit weird to have the two encodings on either side of the input data parameter.

What about either making the `pTextEncoding` parameter optional or even removing it altogether?  We can now automatically detect whether the argument is binary data or a string by using the `is strictly a binary string` predicate, and nowadays it seems reasonable to default to encoding strings as UTF-8 for e-mail.  If the prototype was:

```
function MIMEEncode pData, pEncoding
```

then it doesn't seem to awkward to write:

```
-- The defaults are fine
MIMEEncode("my text", "quoted-printable")
-- For some reason my heart's desire is UTF-16-BE
MIMEEncode(textEncode("my text", "UTF-16-BE"), "quoted-printable")
```

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+      add 1 to tIndex
+   end repeat
+   return tData
+end __EncodeQuotedPrintable
+
+/**
+
+Build a multipart mime content type from an array of pre-encoded content
+
+Description:
```
peter-b, line 258: Please flesh out the description and add an example.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+pContent (array):
+A numerically indexed array of pre-encoded parts including headers. If no headers are present
+a leading crlf should be present.
+
+Returns:
+The multipart encoded content
+
+*/
+
+function MIMEMultipart pType, pContent
```
peter-b: I think it might be slightly more ergonomic to make this:

```
function MIMEMultipart pContent, pType
```

and allow `pType` to default to `mixed` if not specified.

Please name this function with a verb.  What would you call a function that parses a multipart MIME blob?

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1271 @@
...
+Description:
+List the extensions associated with a given MIME type or with a MIME type matching
+a wildcard pattern.
+
+Returns:
+A list of extensions or empty if none were found.
+
+*/
+
+function MIMETypeExtensions pMimeType
```
peter-b: Please name this handler with a verb.

extensions/script-libraries/mime/mime.livecodescript
montegoulding: OK

extensions/script-libraries/mime/mime.livecodescript
montegoulding: OK and Hmm... I generally avoid `Get` for functions as it's both redundant and less english like to write `put GetSomething() into container` rather than `put Something() into container`. It does make more sense in other languages where its `something = GetSomething()`. But assuming it's policy then do you think I should stick with the standardised `MIME` prefix for the library making this `MIMEGetVersion()`

extensions/script-libraries/mime/mime.livecodescript
montegoulding: Sure

extensions/script-libraries/mime/mime.livecodescript
montegoulding: good catch

extensions/script-libraries/mime/mime.livecodescript
montegoulding: Sure that seems reasonable

extensions/script-libraries/mime/mime.livecodescript
montegoulding: OK and probably yes I just didn't think about it

extensions/script-libraries/mime/mime.livecodescript
montegoulding: Sure

extensions/script-libraries/mime/mime.livecodescript
montegoulding: OK & OK & `MIMEParseMultipart()`?

montegoulding:
@peter-b this is probably done to the extent that I need it done at the moment. I hope you don't mind I opened up the discussion of functions being verbs in our language that already has a verb in the statement the function is in to the use list as it seemed an interesting difference to other languages. I think there was only one response so far that thought they must be verbs. BTW do we have our own scripting style guide for LCS anywhere? The LiveCode Script guide points to the FourthWorld one which doesn't mention handler naming other than title casing handler names with the exception of engine messages. I'll leave this WIP while we ponder this ;-)

peter-b (commented):
@montegoulding Thanks for taking the time to go through and address my comments.  I agree that it's unclear what the "idiomatic" naming for LCS functions should be, so let's not worry about those for now.

The implementation looks fairly sensible, although there are a few larger handlers that would ideally be broken down into a set of helper functions.

I've asked @livecodefraser to take a quick look through the API just for a sanity check, but overall i think this is good to go.

peter-b:
I've removed the WIP label because I think we can cope with the handler names that you've currently got. 👍

peter-b:
@montegoulding Note that there's a problem in one of your doc comments that's causing the Travis CI testsuite to fail.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1558 @@
...
+Parameters:
+pContentA (array):
+A numerically indexed array of pre-encoded parts including headers. If no headers are present
+a leading crlf should be present.
+
+pMultipartType (optional string):
+The type of multipart content. If not provided the default type will be `mixed`.
+
+pParamA (optional array):
+Additional parameter array in key : value pairs
```
livecodefraser: This could maybe do with a little clarification about what is meant by "additional parameters" - upon first reading, I interpreted it as indicating they were additional settings controlling the behaviour of the function rather than key=value options added to the Content-Type header.

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1558 @@
...
+Generate multipart content from a LiveCode field
+
+Example:
+MIMECreateMultipartContentFromField the long id of field "email"
+if the result is empty then
+   SendEmail it
+end if
+
+Description:
+Parses the content of a field to create multipart conten
```
livecodefraser: Missing t in "content"

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1558 @@
...
+
+Parameters:
+pField (string):
+Field object long id
```
livecodefraser: Subjective nitpick: I think this might read more easily as "long id of the field to encode" or similar (especially as the other parameter descriptions in the PR are more prose-like).

extensions/script-libraries/mime/mime.livecodescript
```diff
@@ -0,0 +1,1558 @@
...
+   
+   put crlf & "--" & tBoundary & "--" & crlf & crlf after tData
+   
+   return tData for value
+end MIMECreateMultipartContent
+
+
+/**
+
+Generate multipart content from a LiveCode field
```
livecodefraser: Maybe "Generate multipart content from a field control"?

extensions/script-libraries/mime/mime.livecodescript
montegoulding: ok

extensions/script-libraries/mime/mime.livecodescript
montegoulding: thanks!

extensions/script-libraries/mime/mime.livecodescript
montegoulding: ok

extensions/script-libraries/mime/mime.livecodescript
montegoulding: ok

montegoulding:
Thanks @peter-b & @livecodefraser I think I've addressed the issues you found now and I'm fairly sure I got the example compile error

montegoulding:
I've just updated the API based on the recommendations from @runrevmark

peter-b:
@montegoulding This PR now has conflicts :-(

montegoulding:
@peter-b I have rebased this now

peter-b:
@livecode-vulcan review ok d62b78d

livecode-vulcan:
💙 review by @peter-b ok d62b78d78b384715eb5fbba01de9c2a6cae8ac2d

livecode-vulcan:
😎 test success d62b78d78b384715eb5fbba01de9c2a6cae8ac2d

* try-community-linux-x86: [success](https://vulcan.livecode.com/builders/try-community-linux-x86/builds/757)
* try-community-mac: [success](https://vulcan.livecode.com/builders/try-community-mac/builds/762)
* try-community-linux-x86_64: [success](https://vulcan.livecode.com/builders/try-community-linux-x86_64/builds/755)
* try-community-emscripten: [success](https://vulcan.livecode.com/builders/try-community-emscripten/builds/756)
* try-community-ios: [success](https://vulcan.livecode.com/builders/try-community-ios/builds/762)
* try-community-win-x86: [success](https://vulcan.livecode.com/builders/try-community-win-x86/builds/754)
* try-community-android-armv6: [success](https://vulcan.livecode.com/builders/try-community-android-armv6/builds/754)

PR #4864, opened 2016-11-11
Remove QuickTime and revvideograbber

livecodefraser:
This is not yet ready for merging. The replacement code is not available yet and it needs release notes and documentation updates to reflect the changes.

All QT-related properties now throw an error with the exception of dontUseQT and dontUseQTEffects - these always return true and throw an error if an attempt is made to set them to false.

Pulling this PR locally is useful if you need to build against the MacOS 10.12 SDK which removes all QuickTime-related headers.


peter-b:
@livecodefraser Per comments by @runrevmark, we need to keep revvideograbber around for use on Windows with the QuickTime parts expunged.

@livecode-vulcan review not ok 2ce9b0d

livecode-vulcan:
💔 review by @peter-b not ok 2ce9b0da7062aec6e8f9ebf3a9b40f80970c78da

MartinKoob:
It looks like removing docs is part of this task.  This looks like something that is fairly easy that I could do to help out.   Let me know if I could help out on this one.

PR #4869, opened 2016-11-14
Travis CI: Run tests with both en_US and de_DE locales set.

peter-b:


peter-b:
This doesn't seem to be particularly informative.

PR #4951, opened 2016-12-06
[Bug 18917] Maximum text length for a native ios field

livecodepanos:


livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/4975, which is against `develop` branch

PR #5018, opened 2017-01-02
Add Android Toast Function

HedgeHao:
Add toast on Android. Toast is a handy function to display short messages. User can toast a message and set the duration for it.


livecodeali:
@HedgeHao Thanks for this, we transferred your code to an LCB library instead, which will be included in 9 dp 11: https://github.com/livecode/livecode/pull/5860

PR #5065, opened 2017-01-12
[[ Bug 19032 ]] Sort out object deletion

livecodeali:


engine/src/control.cpp
```diff
@@ -454,6 +454,59 @@ void MCControl::deselect()
...
+	MCObject::removefromobjecttree(p_is_child);
+    
+    switch (parent->gettype())
+    {
+        case CT_CARD:
+        {
+            if (!p_is_child)
+            {
+                MCCard *cptr = parent.GetAs<MCCard>();
+                cptr->removecontrol(this, False, True);
```
peter-b: This is a pretty sweet-looking refactor. 💯 

I suggest `parent.GetAs<MCCard>()->removecontrol(this, False, True);` but YMMV.

livecodeali: 👍 

Not sure if this patch is quite correct yet anyway!

livecodeali:
I can't seem to write a test that reproduces 19032. We can't replicate the original recipe in the test system because currently data grids require user-interaction when deleted.

engine/src/control.cpp
```diff
@@ -454,6 +454,60 @@ void MCControl::deselect()
...
+            if (!p_is_child)
+            {
+                parent.GetAs<MCGroup>()->removecontrol(this, True);
+            }
+            break;
+        }
+        default:
+        {
+            parent.GetAs<MCStack>()->removecontrol(this);
+        }
```
peter-b, line 500: I feel like this should be:

```C++
case CT_STACK:
    parent.GetAs<MCStack>()->removecontrol(this);
    break;
default:
    MCUnreachable();
```

This is because you're switching on the parents object type, and it makes sense to have a stack case for stacks. :)

livecodeali:
As this is pretty invasive, here is a patch which prevents the crash with much less churn.
https://github.com/livecode/livecode/pull/5071

livecodeali:
@peter-b Thanks, I have incorporated both your observations in the replacement PR!

PR #5099, opened 2017-01-22
[[ Bug 17664 ]] Enables localization of iOS system UI text

HedgeHao:
According to [Apple's Document:The Bundle Search Pattern](https://developer.apple.com/library/content/documentation/CoreFoundation/Conceptual/CFBundles/AccessingaBundlesContents/AccessingaBundlesContents.html#//apple_ref/doc/uid/10000123i-CH104-SW7). If a .lproj folder exists in the bundle, that localization is used. This patch create .lproj folders for many language so the system can recognize and use it as default localization. Functions display text using iOS system UI (e.g. mobilePick, mobilePickPhoto) will be localized.

I've tested in simulator (10.2) and it works fine. But when testing on my iPhone (10.2) I need to change "CFBundleDevelopmentRegion" in Setting.plist to country name (e.g. US or Taiwan) rather than language name. So besides this patch, I change "CFBundleDevelopmentRegion" from "English" to "Taiwan".

Since I don't have any other iOS devices for testing. This patch should be test in some more versions of iOS devices.

### DIFF
--- a/docs/notes/bugfix-17664.md
+++ b/docs/notes/bugfix-17664.md
@@ -1,0 +1,1 @@
+# This patch enables localization of iOS system UI text.
--- a/ide-support/revsaveasiosstandalone.livecodescript
+++ b/ide-support/revsaveasiosstandalone.livecodescript
@@ -266,6 +266,12 @@ private command revSaveAsMobileStandaloneMain pStack, pAppBundle, pTarget, pSett
    
    -- Generate the plist
    revCreateMobilePlist pSettings, pAppBundle, pTarget, tCopiedFonts, tPlist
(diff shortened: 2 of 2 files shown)


montegoulding:
@HedgeHao after a bit of googling I *think* the issue is that if there is a `Base.lproj` and your preferred language settings includes the development language (even if it's not the preferred I think???) the app will display in the development language.

livecodepanos:
After some internal discussion with the team, it turns out this PR is not entirely correct. See more details in comment 9 of the original report (https://quality.livecode.com/show_bug.cgi?id=17664).

PR #5148, opened 2017-02-03
[[ Foundation ]] Add module loading API to libfoundation

livecodeali:
This patch removes the engine dependency from our dynamic
library loading implementations. The MCU_loadmodule
per-platform implementations now (where possible) defer to
the foundation implementations.

Note that it is not possible to move all of this code into
libfoundation, as we need to do some special things with the
module path in the engine, such as trying to load relative to
MCcmd, or in a location specified by deploy parameters.

### DIFF
--- a/engine/src/dsklnx.cpp
+++ b/engine/src/dsklnx.cpp
@@ -1270,33 +1270,20 @@ public:
         return MCStringCreateWithSysString(tmpnam(NULL), r_tmp_name);
     }
 
-    virtual MCSysModuleHandle LoadModule(MCStringRef p_path)
-    {
-        // dlopen loads whole 4-byte words when accessing the filename. This causes valgrind to make
-        // spurious noise - so in DEBUG mode we make sure we allocate a 4-byte aligned block of memory.
-        //
-        // Because converting to a sys string allocates memory, this alignment will always be satisfied.
-        // (malloc/new always return alignment >= int/pointer alignment)
-        MCAutoStringRefAsSysString t_filename_sys;
-        /* UNCHECKED */ t_filename_sys.Lock(p_path);
-
-        MCSysModuleHandle t_result;
-        t_result = (MCSysModuleHandle)dlopen(*t_filename_sys, (RTLD_NOW | RTLD_LOCAL));
-
-        return t_result ;
-    }
-
-    virtual MCSysModuleHandle ResolveModuleSymbol(MCSysModuleHandle p_module, MCStringRef p_symbol)
-    {
-        MCAutoStringRefAsSysString t_symbol_sys;
-        /* UNCHECKED */ t_symbol_sys.Lock(p_symbol);
-        return (MCSysModuleHandle)(dlsym(p_module, *t_symbol_sys));
-    }
-
-    virtual void UnloadModule(MCSysModuleHandle p_module)
-    {
-        dlclose(p_module);
-    }
+	virtual MCSysModuleHandle LoadModule(MCStringRef p_filename)
+	{
+		return (MCSysModuleHandle)MCSDylibLoadModule(p_filename);
+	}
+	
+	virtual MCSysModuleHandle ResolveModuleSymbol(MCSysModuleHandle p_module, MCStringRef p_symbol)
+	{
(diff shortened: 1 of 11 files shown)


libfoundation/include/system-dylib.h
```diff
@@ -0,0 +1,61 @@
...
+
+MC_DLLEXPORT extern MCTypeInfoRef kMCSDylibModuleNotFoundErrorTypeInfo;
+MC_DLLEXPORT extern MCTypeInfoRef kMCSDylibSymbolNotFoundErrorTypeInfo;
+
+/* ================================================================
+ * Dynamic library operations
+ * ================================================================ */
+
+/* Load the module at path. */
+MC_DLLEXPORT extern "C" void *MCSDylibLoadModuleCString(const char *p_path);
```
peter-b, line 34: You can safely remove all of the the `extern "C"` qualifiers from this header file; it's always included from within an `extern "C"` block.

I think it might be helpful for explanatory purposes to introduce something like:

```C++
typedef void * MCSDylibModuleRef;
```

And then use it through the API rather than `void *`.

For full type safety in the API, you could even create a distinct type:

```c++
class MCSDylibModule { void *m_handle; };
```

libfoundation/include/system-dylib.h
```diff
@@ -0,0 +1,61 @@
...
+
+/* Resolve the symbol with given name in module. */
+MC_DLLEXPORT extern "C" void *MCSDylibResolveModuleSymbolCString(void* p_module, const char *p_symbol);
+MC_DLLEXPORT void *MCSDylibResolveModuleSymbol(void* p_module, MCStringRef p_symbol);
+
+#ifdef __MCS_INTERNAL_API__
+
+void *_MCSDylibLoadModule(MCStringRef p_path);
+void _MCSDylibUnloadModule(void *p_module);
+void *_MCSDylibResolveModuleSymbol(void* p_module, MCStringRef p_symbol);
```
peter-b, line 48: For consistency, please use `__MCS` prefixes for `__MCS_INTERNAL_API__` symbols!

engine/src/dsklnx.cpp
```diff
@@ -1270,33 +1270,20 @@ class MCLinuxDesktop: public MCSystemInterface
...
-    }
+	virtual MCSysModuleHandle LoadModule(MCStringRef p_filename)
+	{
+		return (MCSysModuleHandle)MCSDylibLoadModule(p_filename);
+	}
+	
+	virtual MCSysModuleHandle ResolveModuleSymbol(MCSysModuleHandle p_module, MCStringRef p_symbol)
+	{
+        return (MCSysModuleHandle)MCSDylibResolveModuleSymbol(p_module, p_symbol);
+	}
```
peter-b, line 1281: This has a bit of unsoundness to it; a module handle and the result of resolving a symbol have distinct types.  They have different operations valid on them: for example, you can pass a module handle to `MCSDylibResolveModuleSymbol()` but things will go wrong if you pass it a symbol pointer.

Is it viable to redefine `MCSysModuleHandle` to a type that doesn't require casts to use the new `MCSDylib` API?

As a side point, please avoid C-style casts.

engine/src/dskmac.cpp
```diff
@@ -4140,71 +4140,53 @@ struct MCMacDesktop: public MCSystemInterface, public MCMacSystemService
...
-#endif
+		if (t_handle != nil)
+			return t_handle;
+		
+		MCAutoStringRefAsSysString t_filename_sys;
+		if (!t_filename_sys.Lock(p_filename))
+			return nil;
+		
+		CFURLRef t_url;
+		t_url = CFURLCreateFromFileSystemRepresentation(NULL, (const UInt8 *)*t_filename_sys, strlen(*t_filename_sys), false);
```
peter-b, line 4154: I think that `t_filename_sys.Size()` should Do The Right Thing here.

libfoundation/include/system-dylib.h
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright (C) 2016 LiveCode Ltd.
```
peter-b, line 2: Might as well make this "2016-2017" 😉

PR #5159, opened 2017-02-07
[[ Bug 16714 ]] Fix object layout problem caused by IME on Android

HedgeHao:
In previous versions, when keyboard shows up and the acceleratedrendering property set to true on Android. The whole screen will automatically scrolls up to make a room for keyboard whether the focused field is out of sight or not. This patch add a field property "autoReposition" to let user decide whether a field need to scroll up when showing the keyboard. This patch also fix black area when hiding keyboard (un-focus field or app resumed).


montegoulding:
@HedgeHao I'm sorry this has taken so long to investigate. I'm looking into it now and will push up a PR which just fixes bug 10881. We can then look at how best to implement properties to control the panning behavior as it requires an accurate rect of the focused object which will need some work to implement.

PR #5164, opened 2017-02-08
[[ Bug 15094 ]] Add 'Documents' parameter for specialFolderPath() on Linux

HedgeHao:


peter-b:
@HedgeHao This looks good.

Maybe, it would be better if LiveCode used `g_get_user_special_dir(G_USER_DIRECTORY_DOCUMENTS)` here [1], in order to correctly check the environment variables and to handle the case that the "Documents" directory has a localised name.

You could do this if you wanted to; you'd need to add a stub for `g_get_user_special_dir()` to [engine/src/linux.stubs](https://github.com/livecode/livecode/blob/develop/engine/src/linux.stubs), and then use it to construct the return value.

While you were at it, you could also update the "desktop" special directory to use `g_get_user_special_dir()` as well.

Let me know what you want to do here.

[1] https://developer.gnome.org/glib/2.50/glib-Miscellaneous-Utility-Functions.html#GUserDirectory

HedgeHao:
@peter-b I think you are right.  If I recall correctly, the name of document folder in Linux can be represented in other languages when you set system language other than English (e.g. "文件" in Traditional Chinese). So it might not be a good idea to always return "Documents". My patch now is ill-considered. I will try to follow your suggestions. Thanks for your advice.

peter-b:
@HedgeHao I'll mark this as requiring further work!

@livecode-vulcan review not ok 5881802

livecode-vulcan:
💔 review by @peter-b not ok 5881802b0d2e90d04cc6982a03cc65f5ef9469ab

HedgeHao:
I think the latest commit depends on some thridparty change. I've made a [PR](https://github.com/livecode/livecode-thirdparty/pull/93)

peter-b:
@HedgeHao I'm sorry -- I didn't realise that the thirdparty headers we use were so out of date.

There are two options:

1. Go back to the original plan -- hardcoding a `$HOME/Documents` directory -- and file a bug report that it does not respect environment variables and/or localization
2. Get [bug 19236](http://quality.livecode.com/show_bug.cgi?id=19236) fixed and then go with your new patch with `g_get_user_special_dir()`

Which do you think is best?

HedgeHao:
@peter-b I will go fix bug 19236!

PR #5187, opened 2017-02-14
[Bug 18912] Fixes to object deletion when target or handling a message in behaviors

montegoulding:


montegoulding:
It turns out that the ability to delete the target in a frontScript was a tested feature `TestDeleteTargetInFrontScript` however, as the test stack was -ui there was no window to cause the crash recently fixed. As this leaves an invalid target and there's lots of code that expects it to be valid I'll go ahead and modify the tests to fail with an execution error and we can then discuss what we think is right.

peter-b:
Replaced by #5235

PR #5168, opened 2017-02-09
Update Skia to a more recent version (December 2016)

livecodefraser:
Currently a work-in-progress. Depends on using libc++ on macOS as Skia now uses modern C++ features.

Per-platform status:

- MacOS: done
- iOS: done
- Android: done
- Emscripten: done
- Windows: in progress
- Linux: in progress

There is a corresponding PR for this in the third-party repo; these are just the source code changes required to adapt to the changes in Skia's API.

Various bits are still broken. In particular, legacy blend modes have been removed from libgraphics (though incompletely; assertion failures will be triggered when used). It is not possible to re-add legacy blend modes due to major changes in how Skia handles blending. Additionally, the gradient code (in legacygradients.cpp) does not compile due to a change in the Skia filtering API; this will need to be refactored (probably to use Skia's built-in gradient support just like the widget graphics library provides).

PR #5207, opened 2017-02-22
Use Skia's CoreText functionality to draw text on macOS/iOS

livecodefraser:
Depends on the Skia updates (https://github.com/livecode/livecode/pull/5168)

Using Skia to render the individual glyphs provides a couple of benefits: firstly, fonts containing embedded colour bitmaps (e.g. emoji) now render properly in colour. Secondly, sub-pixel anti-aliasing works properly(!)

We still use CoreText directly for doing the layout, so nothing has changed on that front (Skia doesn't really do layout or things like font substitution, so that's not going to change).


engine/src/eventqueue.cpp
```diff
@@ -495,13 +495,15 @@ static void MCEventQueueDispatchEvent(MCEvent *p_event)
 			}
 
 			// Otherwise 'char_code' is the unicode codepoint, so first map to
-			// UTF-16 (not done properly yet...)
-			unichar_t t_unichar;
-			t_unichar = (unichar_t)t_event -> key . press . char_code;
+			// UTF-16 codeunits
+			unichar_t t_unichars[2];
+            uindex_t t_length = 1;
+            if (MCUnicodeCodepointToSurrogates(t_event->key.press.char_code, t_unichars))
```
livecodeian, line 501: This looks like it needs a fallback for when MCUnicodeCodePointToSurrogate fails, otherwise t_unichars will be left uninitialised.

libgraphics/src/coretext-skia.cpp
```diff
@@ -0,0 +1,418 @@
...
+    if (s_colour_space != NULL)
+        CGColorSpaceRelease(s_colour_space);
+    if (s_measure_context != NULL)
+        CGContextRelease(s_measure_context);
+    MCMemoryDelete(s_measure_data);
+    s_colour_space = NULL;
+    s_text_colour = NULL;
+    s_measure_context = NULL;
+    s_measure_data = NULL;
+}
```
livecodeian, line 72: These need to be renamed really, as they only set up the environment for text rendering.

libgraphics/src/coretext-skia.cpp
```diff
@@ -0,0 +1,418 @@
...
+        // Measure the run
+        // This line depends on the assumption that CGGlyph and SkGlyphID are the same type
+        MCGFloat t_width = t_paint.measureText(t_glyphs, t_glyph_count * sizeof(CGGlyph));
+        
+        // Check for extension of the left-hand bound of the run
+        t_left = MCMin(t_left, t_x);
+        
+        // Check for extension of the right-hand bound of the run
+        t_right = MCMax(t_right, t_x + t_width);
+    });
```
livecodeian, line 393: Perhaps t_left & t_right should always be set on the first iteration in case the line does not begin at x<=0.

libgraphics/src/coretext-skia.cpp
```diff
@@ -0,0 +1,418 @@
...
+    
+    // Release typeset line
+    CFRelease(t_line);
+    
+    // Return the width of the run
+    return t_right - t_left;
+}
+
+// TODO: this seems to have no callers
+bool MCGContextMeasurePlatformTextImageBounds(MCGContextRef self, const unichar_t *p_text, uindex_t p_length, const MCGFont &p_font, const MCGAffineTransform &p_transform, MCGRectangle &r_bounds)
```
livecodeian, line 403: this is called (eventually) by the LCB canvas module in the implementation of "the image bounds of ..."

PR #5249, opened 2017-03-07
Remove support for i386 iPhoneSimulator

peter-b:
As planned, we are dropping support for iPhoneSimulator < 8.2 in LiveCode 9.  This gives us the opportunity to also drop i386 iPhoneSimulator support, which is great because it means our iPhoneSimulator engines will be quicker to compile and take up much less space in the installer.

### DIFF
--- a/config.py
+++ b/config.py
@@ -207,10 +207,7 @@ def guess_xcode_arch(target_sdk):
         else:
             return 'armv7 arm64'
     if sdk == 'iphonesimulator':
-        if int(ver) < 8:
-            return 'i386'
-        else:
-            return 'i386 x86_64'
+        return 'i386 x86_64'
 
 def validate_target_arch(opts):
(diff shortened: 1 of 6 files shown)


livecodepanos:
There are also some references to `6.1` and `7.1` simulators in `prebuilt/scripts/ios.inc` that probably have to be removed.

Other than that, this PR looks fine to me.

config.py
```diff
@@ -207,10 +207,7 @@ def guess_xcode_arch(target_sdk):
         else:
             return 'armv7 arm64'
     if sdk == 'iphonesimulator':
-        if int(ver) < 8:
-            return 'i386'
-        else:
-            return 'i386 x86_64'
+        return 'i386 x86_64'
```
livecodeali, line 210: This needs to be just `return 'x86_64'` for the Skia update to build

PR #5269, opened 2017-03-13
Update libgraphics to account for Skia API changes

livecodeali:
The biggest impact is the removal of custom blend modes in Skia. This
forces us to drop support for legacy blend modes as we can no longer
emulate them.

On the plus side, only using the built-in Skia blend modes will result
in faster blending.

(cherry picked from commit 2ffc8ef900168f5ce0d791b9d6b72d3273a50dc6)


libgraphics/src/context.cpp
```diff
@@ -1350,34 +1463,42 @@ void MCGContextEnd(MCGContextRef self)
...
+		self->layer->canvas->restore();
+	}
 	else
-		self -> layer -> canvas -> drawSprite(t_child_bitmap, t_child_layer -> origin_x, t_child_layer -> origin_y, &t_paint);
+	{
+		// Hardware-backed layer - copy it into RAM
+		sk_sp<SkImage> t_image = t_child_layer->m_surface->makeImageSnapshot();
+		self->layer->canvas->save();
+		self->layer->canvas->resetMatrix();
+		self->layer->canvas->drawImage(t_image, t_child_layer->origin_x, t_child_layer->origin_y, &t_paint);
```
livecodeian: Could this be used instead? http://skia-doc.commondatastorage.googleapis.com/doxygen/doxygen/html/classSkSurface.html#a788d7e9a9eb7e93aef3dce9f684ea09f

libgraphics/src/context.cpp
```diff
@@ -2127,8 +2246,8 @@ static bool MCGContextApplyPaintSettingsToSkPaint(MCGContextRef self, MCGColor p
 		}
 		else if (p_paint_style == kMCGPaintStyleStippled)
 		{
-			t_stipple = new (nothrow) SkStippleMaskFilter();
-			t_success = t_stipple != NULL;			
+			//t_stipple = new (nothrow) SkStippleMaskFilter();
```
livecodeian, line 2248: If the stipple filter is no longer available then I suggest removing this code along with kMCPaintStyleStippled

libgraphics/src/w32text.cpp
```diff
@@ -117,7 +119,7 @@ static bool w32_draw_text_process_transparent_buffer(uint32_t p_fill_color, bool
 			// If the destination pixel is opaque, then set it to transparent black
 			// and flag for opacity. Otherwise; take an average of the three RGB
 			// components and blend with the fill color.
-			if (((*t_context_pxl >> 24) & 0xFF) == 0xFF)
+			if (false && ((*t_context_pxl >> 24) & 0xFF) == 0xFF)
```
livecodeian, line 122: Should this dead code be removed?

libgraphics/src/w32text.cpp
```diff
@@ -249,25 +251,24 @@ static bool w32_draw_text_using_mask_to_context_at_device_location(MCGContextRef
...
+	// We want all drawing to ignore the transform matrix within this function
+	p_context->layer->canvas->save();
+	p_context->layer->canvas->resetMatrix();
+
 	// Now process depending on the first pixel - if transparent, then first process
 	// transparency then process opacity; otherwise vice-versa.
 	if (t_success)
 	{
-		if ((*t_context_pxls >> 24) != 0xff)
+		if (true || (*t_context_pxls >> 24) != 0xff)
```
livecodeian, line 271: Can this if be removed now?

livecodeian (changes requested):

libgraphics/src/w32text.cpp
livecodefraser: The dead code gets removed in the update to use DirectWrite instead of GDI text rendering.

libgraphics/src/context.cpp
livecodefraser: Unfortunately, the resetMatrix bit is required - I did try it that way before and it caused issues.

livecodeian: The resetMatrix stuff is fine, but the docs for SkSurface::draw implies that it could be more efficient than snapshot followed by drawImage

engine/src/mblandroidtypeface.cpp
```diff
@@ -87,10 +90,12 @@ void MCAndroidTypefaceRelease(MCAndroidTypefaceRef p_typeface)
 
 bool MCAndroidTypefaceGetMetrics(MCAndroidTypefaceRef p_typeface, uint32_t p_size, float &r_ascent, float &r_descent, float &r_leading, float &r_xheight)
 {
-	bool t_success = true;
-	
+    // Skia APIs expect typefaces to be passed as shared pointers
+    sk_sp<SkTypeface> t_typeface((SkTypeface*)p_typeface);
+    t_typeface->ref();
```
livecodeian: looks like there's a convenience function for this - sk_ref_sp(T* obj) which wraps the ptr and increments the ref count

livecodeian:
@livecode-vulcan review ok db257d5

livecode-vulcan:
💙 review by @livecodeian ok db257d5bbdbea99a186e8d768634e5a151ebe5a1

PR #5281, opened 2017-03-15
[[ Bug 19416 ]] Refactor object message clear

livecodeali:
In order to fully purge the pending messages of messages for a given
object when the object is removed from the object tree, we must
recursively remove these messages for all children of the deleted
object.


livecodeali:
@livecode-vulcan review by @peter-b ok 1485ed4

PR #5285, opened 2017-03-15
[[ Docs ]] Changes to listStyle.lcdoc

asayd:
- Wrapped list bullet text was being formatted incorrectly; unwrapped those lines to fix formatting.
- Changed param names to avoid confusion with Livecode tokens.
- Tightened up syntax statement; 'line' is a required keyword and the line number is a param.
- Added example.
- Added germane references and deleted unrelated ones.

### DIFF
--- a/docs/dictionary/property/listStyle.lcdoc
+++ b/docs/dictionary/property/listStyle.lcdoc
@@ -2,7 +2,7 @@ Name: listStyle
 
 Type: property
 
-Syntax: set the listStyle of <line> of <field> to <style>
+Syntax: set the listStyle of line <lineNumber> of <fieldRef> to <styleType>
 
 Summary:
 Specifies what type of list style is used for the line of text.
@@ -18,9 +18,19 @@ Platforms: desktop, server, mobile
 Example:
 set the listStyle of line 1 of field 1 to "disc"
 
+Example: 
+# clear all list formatting in a field
+set the listStyle of line 1 to -1 of field 1 to empty
+
 Parameters:
-style (enum):
-The style of bullet to use for the line of text in the list. One of:
+lineNumber (integer):
+A number corresponding to one of the lines in the field.
+
+fieldRef:
+Any valid <object reference|field reference>.
+
+styleType (enum):
+The style of bullet to use for the line of text in the list.
 
 -   "disc": A solid round bullet.
 -   "circle": A ring.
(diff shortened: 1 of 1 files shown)


livecodeali:
@asayd I wonder if `lineNumber` might be better as `lineChunk`, just to cover the case where it is an expression for multiple lines. I'm not sure though, what do you think?

asayd:
@livecodeali Good point. Of course this same should then apply to any property that affects only lines (paragraphs) of a field. There are maybe 8 to 10 of these, and currently the syntax for all of them read:

    set the <property> of line of field to <value>

Syntax statements for properties that can apply either to a whole field or a line tend to look like this:

    set the <property> [of line lineNumber] of field to <value>

So it's inconsistent. I might propose something like this for line-specific properties:

    set the <property> of line lineNumber [to lineNumber] of fieldRef to <value>

And by analogy, for field, optional line properties:

    set the <property> [of line lineNumber [to lineNumber]] of fieldRef to <value>

Alternatively, I could take the approach you suggest:

    set the <property> of lineChunk of fieldRef to <value> 

Which treats line chunks as a a fundamental concept that we represent in some syntax statements and define in the Parameters element. Regardless of which approach we choose, it seems clear that I'd want to change a fair number of these entries to make them consistent. What do you think?

livecodeali:
@asayd I think there may be too many to change and too much syntax to add if we make all the chunk properties explicit, so personally I'd prefer to treat the line chunk as a fundamental concept (for reference, here's a list I got from the source code of all field properties that accept a char or line chunk:
```
Text
UnicodeText
PlainText
UnicodePlainText
FormattedText
UnicodeFormattedText
RtfText
HtmlText
StyledText
FormattedStyledText
CharIndex
LineIndex
FormattedTop
FormattedLeft
FormattedWidth
FormattedHeight
FormattedRect
LinkText
ImageSource
Metadata
Visited
Encoding
Flagged
FlaggedRanges
ListStyle
ListDepth
ListIndent
ListIndex
FirstIndent
LeftIndent
RightIndent
TextAlign
SpaceAbove
SpaceBelow
TabStops
TabAlignments
BorderWidth
BackColor
BorderColor
HGrid
VGrid
DontWrap
Padding
TabWidths
Invisible
ForeColor
BackColor
TextFont
TextSize
TextStyle
TextShift
TextStyleElement
```
)

I think the easiest thing to do here would be to make `line chunk` a valid _type_ for a parameter, and then ensure that parameter types are linked to a glossary entry which describe them. I think we need to do this anyway for more complicated parameter types (eg `object reference`, which we've discussed before, and `lines of string` and that sort of thing).

If you agree, I can probably fairly quickly add some stuff to the renderer to make it work. Anyway interested to know your thoughts, and sorry for the tardiness of my reply!

livecodeali:
Another reason to do it like that is that we could feasibly replace `line chunk` parameters on the fly in the syntax with `line lineNumber [to lineNumber]`, whereas it would be much harder to do the converse.

asayd:
@livecodeali Sorry for the late reply. I do like your idea of making `line chunk` a valid type for a parameter. Are you talking about running some kind of batch job to change all of these at once, or would the change need to happen manually?

PR #5353, opened 2017-04-05
[[ Bug 19537 ]] Allow currently debugging stack to execute

livecodeali:
This patch allows the stack being debugged line-by-line to have
its handlers executed in the course of debugging.

### DIFF
--- a/docs/notes/bugfix-19537.md
+++ b/docs/notes/bugfix-19537.md
@@ -1,0 +1,1 @@
+# Allow currently debugging stack to execute other handlers and remain debugging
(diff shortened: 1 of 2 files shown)


montegoulding:
@livecodeali There's a similar block in `execparenthandler` which possibly needs the same treatment, however, I'm fairly sure this patch will break the profiler which depends on the setting of `MCtrace` to true if the handler is on the traceStack. The only other place it is set to true is after a debug message if the tracestack is valid. It could be that `MCtrace` could be refactored out in favour of `MCtracestackptr` being valid and being the stack of the target. Actually that could solve the issue I mentioned while I was doing the profiler of frontScripts not being traced even when the target is on the traceStack.

PR #5465, opened 2017-05-15
[[ Coverity Scan ]] Ensure gcc 4.9 is used

livecodeali:
Putting the EXPORTs in the `script` step of the travis build script
means it gets skipped by the coverity_scan rules. Try putting them
in `before_install` instead

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ matrix:
 
 # Install any required tools
 before_install:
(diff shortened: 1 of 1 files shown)


livecodeali:
Even though this causes the coverity scan build to timeout on Travis, it's probably worth merging anyway as it's correct and will still be needed if we cut build time down using more prebuilts.

PR #5472, opened 2017-05-16
Text should be textDecoded for transfer into LiveCode

trevordevore:
- Rename ulDecodeData to ulDecodeURLData
- Add ulDecodeData which takes headers and data as a parameter.
- Update ulDecodeData to look at Content-Type header when deciding how to decode
- Move ulDecodeURLData calls to locations where “cached” state is set. This works for both get and load url.

Notes:
- Did not add ulDecodeData call for Ftp
- Downloading data directly to a file will not convert encoding.

### DIFF
--- a/ide-support/revliburl.livecodescript
+++ b/ide-support/revliburl.livecodescript
@@ -194,9 +194,6 @@ on loadUrl x,y
          
          ulGetFormat newUrl,lvCount
          
-         -- SN 2014-02-21 Decode the URL read according to the encoding specified in the head part
-         ulDecodeData newUrl
-         
          if laUrlLoadStatus[newUrl] is "error" and not laCancelled[newurl] then
             ulSendMessage newUrl ##send message now only if error occurred befoe block point
             return "error"
@@ -267,9 +264,7 @@ on getUrl x
       put "getData" into laAction[newUrl]
       put empty into laData[newUrl]
       ulGetFormat newUrl,lvCount # convert url to components
-      
-      -- SN 2014-02-21 Decode the URL read according to the encoding specified in the head part
-      ulDecodeData newUrl
+
       ##final clean up here
       
       if laAuthRecursCount[newUrl] < 1 then
@@ -1240,6 +1235,7 @@ on ulDoProcess x,y
                put empty into laUrlErrorStatus[laUrl[x]]
                if laFile[laUrl[x]] is empty then
                   put "cached" into tStatus
+                  ulDecodeUrlData laUrl[x]
                else
                   put "downloaded" into tStatus
                end if
@@ -1343,6 +1339,7 @@ on ulDoProcessLength x,y
       if char 1 of laStatusCode[laUrl[x]] = 2 then ##in 200 range--OK
          if laFile[laUrl[x]] is empty then
             put "cached" into tStatus
+            ulDecodeUrlData laUrl[x]
          else
             put "downloaded" into tStatus
          end if
@@ -1433,6 +1430,7 @@ on ulDoProcessChunked x,y
             if char 1 of laStatusCode[laUrl[x]] = 2 then ##in 200 range --OK
                if laFile[laUrl[x]] is empty then
                   put "cached" into tStatus
+                  ulDecodeUrlData laUrl[x]
                else
                   put "downloaded" into tStatus
                end if
@@ -1547,6 +1545,7 @@ on ulDoProcessNoLength x,y
                if char 1 of laStatusCode[laUrl[x]] = 2 then
                   if laFile[laUrl[x]] is empty then
                      put "cached" into tStatus
+                     ulDecodeUrlData laUrl[x]
                   else
                      put "downloaded" into tStatus
                   end if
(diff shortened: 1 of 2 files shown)


montegoulding:
Hi @trevordevore it looks to me like server responses with non `2*` codes will not be decoded now when they would have been previously. Also I double-checked that the header is meant to take precedence over the meta tag and [this](http://stackoverflow.com/questions/7102925/prefer-charset-declaration-in-html-meta-tag-or-http-header) indicates it does but that a BOM should take precedence of the header.

What do you think about adding the BOM check to this PR?

trevordevore:
@montegoulding I'm traveling but will look into this when I return. Makes sense to add a BOM check. Is there any code available that already implements a BOM check?

montegoulding:
@trevordevore https://github.com/livecode/livecode/blob/develop/engine/src/sysspec.cpp#L1315

It would be a reasonable thing to have exposed to LCS somehow. Ideally something that returned the encoding and length. However, it would be relatively easy to translate the above into a function to return the encoding or empty if no BOM is found and the length to strip off.

montegoulding:
@trevordevore as I just bumped into libURL not decoding something correctly I thought I'd tickle this PR to see if you had any more thoughts on it

trevordevore:
@montegoulding I haven't had a chance to look into this further. Unfortunately it will probably be a while before I can look at adding and testing BOM support.

bwmilby:
So, would something like this work?

```
function CheckForBOM pInput
   put numtobyte(0xFF) & numtobyte(0xFE) into kUTF16LE
   put numtobyte(0xFE) & numtobyte(0xFF) into kUTF16BE
   put kUTF16LE & numtobyte(0x00) & numtobyte(0x00) into kUTF32LE
   put numtobyte(0x00) & numtobyte(0x00) & kUTF16BE into kUTF32BE
   put numtobyte(0xEF) & numtobyte(0xBB) & numtobyte(0xBF) into kUTF8
   
   switch byte 1 to 4 of pInput
      case kUTF32LE
         return "UTF-32LE"
         break
      case kUTF32BE
         return "UTF-32BE"
         break
   end switch
   switch byte 1 to 2 of pInput
      case kUTF16LE
         return "UTF-16LE"
         break
      case kUTF16BE
         return "UTF-16BE"
         break
   end switch
   if byte 1 to 3 of pInput is kUTF8 then
      return "UTF-8"
   end if
   
   return empty
end CheckForBOM
```

ide-support/revliburl.livecodescript
```diff
@@ -1760,50 +1760,58 @@ function ulNormaliseEncoding pEncoding
...
-      if tEncoding is not empty then
-         local tDecoded
-         put uniEncode(laData[pUrl], tEncoding) into tDecoded
-         put uniDecode(tDecoded) into laData[pUrl]   
-      end if
-   end if
+command ulDecodeData pHeaders, @pData
+  local tEncoding, tCharset, tLineNo
+
+  put empty into tEncoding
```
bwmilby, line 1766: One option would be to change this line to `put CheckForBOM(pData) into tEncoding` and then check to see if it is not empty.  If not, trim the necessary bytes from pData and skip the remainder of the checks and just decode.  Either UTF32 is 4 bytes, either UTF16 is 2 bytes, and UTF8 is 3 bytes.

montegoulding:
@bwmilby @trevordevore I think even simpler than that could be something like this:

```
function CheckForBOM pInput
       constant kEncodings = "UTF-8,UTF-16,UTF-16BE,UTF-16LE,UTF-32,UTF-32BE,UTF-32LE"

       local tCodepoint, tEncoding
       put numToCodepoint(0xFEFF) into tCodepoint
       repeat for each item tEncoding in kEncodings
                if pInput begins with  textEncode(tCodepoint, tEncoding) then
                         return tEncoding
                end if
       end repeat
       return empty
end CheckForBom
```

montegoulding:
BTW ideally we would have a file and folder handling script library with tests etc in `extensions/script-libraries` and it can include stuff like this, normalizing line endings, recursing directories, copying, deleting and so forth.

bwmilby:
I think you may need to reverse the order.  UTF16LE and UTF32LE start with the same 2 bytes.  I'm not seeing how it works for UTF8 yet though.

montegoulding:
ah ok ... I wrote it on the fly ;-)

Yes reverse `kEncodings`!

montegoulding:
Also probably remove `UTF-16` and `UTF-32` from the list as they are redundant.

bwmilby:
Just checked it and it does work for UTF 8.  Now to figure out why :)

montegoulding:
Yes it should work barring my stuffed up ordering ;-)

bwmilby:
I'm getting something odd though.  I can see the bytes shift from UTF32 LE/BE, but for UTF16 BE is returning blank in LC9DP8.  Works as expected in 8.1.6

montegoulding:
What is returning blank? The function?

bwmilby:
Never mind, must have been a typo (was getting blank when just having the textencode print in the msg box)

PR #5490, opened 2017-05-19
[[ Tests ]] Add tests for widgets in built standalones

livecodeali:
This patch adds tests to the IDE test suite where standalones are built which
simply create an instance of a given widget and ensure no errorDialog message
is sent.

(cherry picked from commit 5408d03d3bb5a1afcb3d88c127ed2c6fece79208)


livecodeali:
This is just the develop version of https://github.com/livecode/livecode/pull/5487

PR #5573, opened 2017-06-07
[[ Bug 19810, 19811, 19812, 11773 ]] Try issues

montegoulding:
This PR is a placeholder for try fixes. It can be split up as needed.

Based on https://github.com/livecode/livecode/pull/5572

### DIFF
--- a/docs/dictionary/control_st/pass.lcdoc
+++ b/docs/dictionary/control_st/pass.lcdoc
@@ -2,7 +2,7 @@ Name: pass
 
 Type: control structure
 
-Syntax: pass <messageName> [to top]
+Syntax: pass { <messageName> [to top] | error }
 
 Summary:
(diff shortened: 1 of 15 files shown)


montegoulding:
I just realised that `pass error` should have some script parser tests to ensure it errors outside of catch and finally statement blocks. I won't worry about that until we are looking more seriously at this PR though.

montegoulding:
I just thought of another test. Considering the `finally` block is executed regardless of `return` and `exit` statements we should test what happens in the event of:

```
try
   get something
catch tError
   return tError for error
finally
   return it for value
end try
```
I expect that the result will always be empty and it will always be it but it but we should have a test for it.

There's also [this](http://quality.livecode.com/show_bug.cgi?id=14618) which I'm not sure what can be done about. My gut feeling is the result should be copied at the end of the try block and reset at the end of the finally block but maybe that's a bit scary.

PR #5600, opened 2017-06-14
[[ Scriptifier ]] Implement scriptifier

montegoulding:
This patch adds a scriptifier tool that parses a stackFile and exports
non-trivial scripts as script only behavior stacks. It sets the stackFiles
property ensuring the scripts can be found when the stack is loaded. The
script in the original stack is set to empty. For simplicity any control
with an existing behavior is not modified.

### DIFF
--- a/tools/Scriptifier.livecodescript
+++ b/tools/Scriptifier.livecodescript
@@ -1,0 +1,269 @@
+﻿script "Scriptifier"
+local sInit = "false"
+
+---------------------------------------------
+--- EVENTS
+
+on openStack
+   __InitUI
+end openStack
+
+on resumeStack
+  __UpdateStackList
+end resumeStack
+
+on menuPick pWhich
+   switch short name of the target
+      case "mainstacks"
+         break
+   end switch
+   pass menuPick
+end menuPick
+
+on mouseUp pButton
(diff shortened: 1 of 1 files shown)


montegoulding:
Do we think the file naming scheme we ended up with in the environment stack PR is general enough to be hard coded into this. At the moment behaviors are created in a `behaviors` dir with the stack name as a file prefix. That PR uses a stack name dir for the behaviors instead therefore keeping binary stack and behaviors dir next to each other in an alphabetically sorted list.

livecodeali:
@livecode-vulcan review ok 811fa6e
Did you end up changing the naming scheme here? TBH I think either is probably fine.

livecode-vulcan:
💙 review by @livecodeali ok 811fa6e77e514c7e3ccae150af5fd856199719e1

montegoulding:
@livecodeali yes I did. At some stage it might be nice to add in an option to make the behavior stack name be reverse domain with a given prefix but I haven't done anything about that yet.

montegoulding:
Woops forgot to wait for vulcan... 🙄

livecodeali:
Oh well, vulcan couldn't have done anything anyway

PR #5609, opened 2017-06-15
[[ Bug 19877 ]] Accept connections in ephemeral port range

montegoulding:
This patch adds support for accepting connections in the ephermeral
port range by setting the socket name to the bound port when accepting
connections on port 0 and returning the socket name in the it variable.

### DIFF
--- a/engine/src/opensslsocket.cpp
+++ b/engine/src/opensslsocket.cpp
@@ -1198,6 +1198,23 @@ MCSocket *MCS_accept(uint2 port, MCObject *object, MCNameRef message, Boolean da
 		listen(sock, SOMAXCONN);
 #endif
     
+    if (0 == port)
+    {
+        socklen_t addrsize = sizeof(addr);
+        if (getsockname(sock, (sockaddr *)&addr, &addrsize) < 0)
+        {
+            MCresult->sets("can't get bound port");
+#if defined(_WINDOWS_DESKTOP) || defined(_WINDOWS_SERVER)
+            closesocket(sock);
+#else
+            close(sock);
+#endif
+            return NULL;
+        }
+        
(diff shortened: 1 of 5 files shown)


engine/src/opensslsocket.cpp
```diff
@@ -1198,6 +1198,13 @@ MCSocket *MCS_accept(uint2 port, MCObject *object, MCNameRef message, Boolean da
 		listen(sock, SOMAXCONN);
 #endif
     
+    if (0 == port)
+    {
+        socklen_t addrsize = sizeof(addr);
+        getsockname(sock, (sockaddr *)&addr, &addrsize);
```
livecodemichael: We should probably check the return value from `getsockname()`.

engine/src/opensslsocket.cpp
```diff
@@ -1198,6 +1198,13 @@ MCSocket *MCS_accept(uint2 port, MCObject *object, MCNameRef message, Boolean da
...
+        port = ((struct sockaddr_in *)&addr)->sin_port;
```
livecodemichael: Do we need to convert the port number from network byte order?

montegoulding: ah, yes we do! thanks!

engine/src/opensslsocket.cpp
montegoulding: yes OK, in the event of failure here I guess we should close, set a result error and return NULL.

livecodemichael:
@livecode-vulcan review ok d3e2be1

livecode-vulcan:
💙 review by @livecodemichael ok d3e2be1e3a58daa0bdf867850dad2754fd7cf7ae

PR #5774, opened 2017-08-10
[[ Embedded ]] Add support for building LiveCode as an embedded library

livecodeali:


livecodeali:
Depends on https://github.com/livecode/livecode-private/pull/190, but is essentially reviewable now.

PR #5778, opened 2017-08-11
Update allowableDragActions.lcdoc

livecodealice:
Changed a single word in the move value line of the allowableDragActions entry to make the sentence make sense.


livecodepanos:
@livecodesam it seems that your branch `patch-17` was accidentally branched-off from `develop-8.0`, so if we now try to merge it into branch `develop-8.1` or `develop` there will be conflicts.

The easiest way to resolve this would be to close this PR, and submit a new one against `develop-8.1`, after first making sure you have branched-off from `develop-8.1`

PR #5781, opened 2017-08-13
[[ Bug 16941 ]] Implement escaped literals

montegoulding:
This patch implements escaped literals wrapped in single
quotes. For example:

    put 'foo\n"bar"'

Equals:

    foo
    "bar"

This patch has the side effect that `'` is no longer an
acceptable char in an identifier.

Just leaving this here for discussion at the moment. Paging @runrevmark

### DIFF
--- a/engine/src/ide.cpp
+++ b/engine/src/ide.cpp
@@ -849,8 +849,9 @@ static void tokenize(const unsigned char *p_text, uint4 p_length, uint4 p_in_nes
 		}
 		else
 		{
-			switch(type_table[t_char])
-			{
+            uint8_t t_current_type = type_table[t_char];
+            switch(t_current_type)
+            {
 				case ST_SPC:
 					t_class = COLOURIZE_CLASS_WHITESPACE;
 					while(t_index < p_length && type_table[(t_char = next_valid_char(p_text, t_index))] == ST_SPC)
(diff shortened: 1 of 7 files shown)


montegoulding:
Probably should add support for `\u` and `\U` to this...

montegoulding:
Added `\u` and `\U`, however, I still need to work out how to handle parser errors... eg `\U` without 8 hex digits after etc. Ideally it would throw a parser error but at the moment it's just setting the token to empty.

montegoulding:
The implementation of `\U` is wrong as it should use codepoint rather than unichar...

montegoulding:
escapedliteral
Tis good fun ;-)

mwieder:
OK - I gotta admit that's impresive. I do echo Mike's concern, though, that things like Windows UNC formats like "C:\user\docs\textLetter.txt" not get get mangled in transit.

montegoulding:
Well `"C:\user\docs\textLetter.txt"` would remain untouched of course and if you needed to hard code such a string in single quotes then you would need to double escape the backslash like in every other language (`'C:\\user\\docs\\textLetter.txt'`).

montegoulding:
escapedliteral

Like this

mwieder:
Ah... I missed that. Escaping is within *single* quotes, not double quotes. I expect there may be some user confusion ahead since the two aren't completely interchangeable, but it does very neatly get around any backwards compatibility issues and allow embedding at the same time. Nice.

montegoulding:
@mwieder as it's not likely this PR will be accepted then there really won't be any chance for user confusion... having said that the whole point of single quotes is to create strings that aren't the same as can be in double quotes even if that is _just_ the double quote char permitted and not escapes. I don't really see the point of offering a solution only for quotes though. I'd rather leave things as they are until I can have a proper solution allowing me to embed return & tab & quote and so on in a constant declaration.

montegoulding:
escapedreturn
A bit more fun because it came up in discussion ;-)

mwieder:
Can we revive this PR and discussion now (three years later)?

montegoulding:
I'd be surprised if we did ;-)

PR #5804, opened 2017-08-17
[Dictionary] Update answer effect

livecodealice:
It: Put `<custom property>` on one line so that it no longer goes to undefined.
References: Changed capitalisation on `qtEffects (function)` to fix the link.


livecodeali:
@livecode-vulcan review ok 0aa9fc3

PR #5806, opened 2017-08-17
[Dictionary] Update answer folder

livecodealice:
General: Added angle brackets to terms where other instances of those terms had them.
Description: Put `<file dialog box|standard file dialog>` on one line to fix the link.


livecodeali:
@livecode-vulcan review ok 0dd2188

PR #5813, opened 2017-08-18
[Dictionary] Update ask file with type

livecodealice:
Fixed links that were split across lines.


livecodeali:
@livecode-vulcan review ok 9c22261

PR #5817, opened 2017-08-18
[Dictionary] Update audioClip

livecodealice:
Changed an example at the bottom of the entry so that it half of it isn't absent when viewed in the dictionary.


livecodeali:
@livecode-vulcan review ok 0117635

PR #5845, opened 2017-08-25
[Dictionary] Update bottomLeft.lcdoc

livecodealice:
Changed `<absolute (screen) coordinates(glossary)>` to `<absolute coordinates(glossary)|absolute (screen) coordinates>` as the original does not display or link correctly and absolute coordinates is a synonym.
Changed `<relative (window) coordinates(glossary)>` to `<relative coordinates(glossary)|relative (window) coordinates>` as the original does not display or link correctly and relative coordinates is a synonym.


livecodeali:
@livecode-vulcan review ok 8de6c1d

PR #5884, opened 2017-08-31
Fix bug 20302

macMikey:
• Documentation for dropboxGetMetadata incorrectly states that pPath will be deleted.
• dropboxGetMetadata parameter pIncludeMediaInfo is supposed to be optional, and default to "false".


macMikey:
UGH!  Wrong QR - it's 20306 not 20302

PR #5885, opened 2017-08-31
Fix bug 20306

macMikey:
dropboxGetMetadata documentation incorrectly states that pPath will be deleted.
dropboxGetMetadata parameter pIncludeMediaInfo is supposed to be optional and default to "false".

### DIFF
--- a/extensions/script-libraries/dropbox/notes/20302.md
+++ b/extensions/script-libraries/dropbox/notes/20302.md
@@ -1,0 +1,2 @@
+# [20302] Documentation for dropboxGetMetadata incorrectly states that pPath will be deleted
+# [20302] dropboxGetMetadata pIncludeMediaInfo is supposed to be optional, defaulting to "false"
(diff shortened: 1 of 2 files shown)


extensions/script-libraries/dropbox/notes/20302.md
```diff
@@ -0,0 +1,2 @@
+# [20302] Documentation for dropboxGetMetadata incorrectly states that pPath will be deleted
+# [20302] dropboxGetMetadata pIncludeMediaInfo is supposed to be optional, defaulting to "false"
```
montegoulding, line 2: @macMikey I'm fairly sure that the second line here will clobber the first when release notes are generated. I suggest either reword them both to a single line or open a new report for one of them.

Perhaps:
```
# [20302] Ensure `dropboxGetMetadata pIncludeMediaInfo` defaults to `false` and fix copy/paste error in description
```

PR #5887, opened 2017-09-01
Edition components

montegoulding:
This is largely for comment at this stage. It is on top of my release notes stuff so please look largely at the last commit that refactors edition info


montegoulding:
I think I've picked the bits I wanted from this

PR #5927, opened 2017-09-09
[[ DON'T MERGE ]] Test explicit vars IDE test

montegoulding:


montegoulding:
@livecodeali seems like it ran ok... I wonder if we should change the filter to show which things are not ok to strip those with an XFAIL reason?

montegoulding:
Note this is too long to be shown in the travis log viewer because the IDE tests are being run twice. My other PR sorts that as well as improving the output but as to do that we lose the IDE crashed test then I'm not sure if it's a good idea. I had a play with adding a `processClosed` message but haven't managed to get that working yet.

PR #5974, opened 2017-09-21
[[ Scriptifier ]] If parent name is not set then skip

montegoulding:


livecodeali:
Do you think it's worth using 
```
get the properties of pObject
if it["name"] is not empty then
```
That seems to be the only reliable way of getting empty for an object with no name

montegoulding:
Good idea

PR #5998, opened 2017-09-28
[Dictionary] Update ceiling.lcdoc

livecodealice:
Changed a link to `<return(constant)>` to link to `<return>` instead and removed the reference to `return (constant)` as it has nothing to do with this function.


livecodeali:
@livecode-vulcan review ok 6af0dc4

livecode-vulcan:
💙 review by @livecodeali ok 6af0dc43dc310ea076da7ce141cc12562f5e62dc

PR #5993, opened 2017-09-27
[[ ScriptExtensions ]] Make script extensions work like lcb extensions

livecodeali:
To provide a bit of uniformity in the way script-library extensions
are built, packaged, and included in the IDE and standalones, this
patch rejigs some gyp files and the installer manifest to build the
script libraries into the packaged_extensions folder and include them
in the IDE Extensions folder in the installed LiveCode.

Additionally this patch:
- Tweaks the docs parser to accept title, version and author metadata
  at the library level. This info is used as the extension metadata
  in lieu of a manifest
- Adds the library level docs as metadata for the current set of
  script libraries
- Adds a script library extension packaging script, which builds
  optionally protected script libraries into the packaged_extensions
  folder

PR #6081, opened 2017-10-25
[[ Bug 20603 ]] Include extensions' resources folder in standalones

livecodeali:


livecodeali:
This sort of thing is pretty difficult to test in develop-8.1 - and will be *much* easier to test in develop once https://github.com/livecode/livecode/pull/5993 is merged, so I'm going to omit the tests here and add them once this is merged up.

I'm not sure what branch this should be against anyhow... it's technically a bug 😄

PR #6087, opened 2017-10-31
[[ NO MERGE ]] Test if indent patch

montegoulding:


BerndN:
@montegoulding now what?

montegoulding:
OK so from the test results I can see the failure is not related to your patch. The indent tests passed so I can review your patch and we can sort out the other failures separately.

PR #6079, opened 2017-10-24
[Bug 16551]  Ensure "hilitedButtonName" returns the name of the hilited button

livecodepanos:

PR #6100, opened 2017-11-04
Revert "[Bug 16551]  Ensure "hilitedButtonName" returns the name of the hilited button"

livecodepanos:
Reverts livecode/livecode#6079

Reverting this patch because it broke the standalone settings:

`put the hilitedButtonName of group "inclusions"` now returns:

`button "Search"` instead of just `"Search"`


livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/6106

PR #6133, opened 2017-11-15
[[ JDK ]] Find JDK 9+ on windows

montegoulding:
This could be rebased back to 8.1 unless anyone can think of a reason not to. I'm also wondering about moving all the java home finding to the initialisation library and just use `$JAVA_HOME` here.


montegoulding:
There appears to be other issues to deal with so I'm making this WIP for now. Minimum source and target options for javac is 1.6 (we use 1.5).

montegoulding:
We are unfortunately stuck I think. It seems that the android sdk doesn’t support using > 8 because `dx.bat` uses now no longer supported `-Djava.ext.dirs` option. On Mac and linux the dx script does not use that option.

PR #6135, opened 2017-11-15
[[ Inclusions ]] Keep existing inclusions when searching

montegoulding:
This patch will ensure that any inclusions that are already set
are not lost when searching for inclusions. This will mean that
deploy inclusions and dependencies set in `revSBGetSettings` are
not lost when searching for inclusions.

### DIFF
--- a/ide-support/revsblibrary.livecodescript
+++ b/ide-support/revsblibrary.livecodescript
@@ -2527,7 +2527,7 @@ private command revSmartChecking pStackFilesList, @xSettings
       
       revStandaloneProgress "Auto detecting library inclusions..."
       
-      put empty into tNewSettings["scriptLibraries"]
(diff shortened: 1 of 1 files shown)


livecodeali:
@montegoulding Do you think we need a way of clearing all inclusion settings as well as this?

montegoulding:
Hmm... so the concern is someone adds an inclusion on the inclusions pane then switches to search for inclusions and the inclusion keeps on being added? Perhaps choosing search should clear them. The problem that I was trying to resolve here is the call to get settings updates some of the inclusions then they get lost if there's a search. I actually think given the patches you are making to inclusions it's time to have a way to both search and explicitly choose. Whether that's a third choice or just making search a checkbox I'm not sure.

PR #6155, opened 2017-11-22
[[ Bug ]] Fix crash where parentscript is invalid after msg

montegoulding:
This patch uses an `MCObjectHandle` to ensure that the parent script
object is valid when tryign to add its long id to the error stack
as a substitue it uses the executing object's long id instead.

### DIFF
--- a/engine/src/object.cpp
+++ b/engine/src/object.cpp
@@ -979,9 +979,11 @@ Exec_stat MCObject::execparenthandler(MCHandler *hptr, MCParameter *params, MCPa
 	if (MCmessagemessages)
 		sendmessage(hptr -> gettype(), hptr -> getname(), True);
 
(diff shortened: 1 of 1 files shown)


livecodeali:
@montegoulding Is it possible to add a test for this?

montegoulding:
@livecodeali hmm... I'll have to see if I can even reproduce it outside filemaker

montegoulding:
well I found a different crash while trying to find a recipe for this one... yay ;-)

montegoulding:
got it! will fix the other too as it's related

montegoulding:
https://github.com/livecode/livecode/pull/6171 covers this crash and a different one if the debugger is running by throwing an error if mutating a behavior that may be executing https://github.com/livecode/livecode/pull/6171

PR #6171, opened 2017-11-27
[[ Bug ]] Throw error if changing behavior of executing object

montegoulding:
This patch ensures that an error will be thrown when setting a
behavior on an executing object if new behavior is a different
object. There is no error if the object did not have a behavior
because it is not possible that the behavior script is executing.


montegoulding:
Actually this appears to have been reworked in develop and I _think_ the only issue there will be when setting to empty. Will discuss internally.

montegoulding:
See https://github.com/livecode/livecode/pull/6172

PR #6174, opened 2017-11-28
[[ Bug 20663 ]] Adjust segmented control corner radius based on orientation

bwmilby:
The segmented control is hard coded for the corner radius to be 1/5 of the height.  When changing to a vertical orientation, width should be used.

### DIFF
--- a/extensions/widgets/segmented/notes/20663.md
+++ b/extensions/widgets/segmented/notes/20663.md
@@ -1,0 +1,1 @@
+# [20663] Adjust corner radius based on control orientation
--- a/extensions/widgets/segmented/segmented.lcb
+++ b/extensions/widgets/segmented/segmented.lcb
@@ -295,7 +295,11 @@ public handler OnPaint()
 	if mGeometryIsChanged then
 		calculateExtents()
 		-- update mPerimeter and mLines variables if the geometry has changed
-		put (the trunc of my height)/5 into mRadius
+		if mIsHorizontal then
+			put (the trunc of my height)/5 into mRadius
+		else
+			put (the trunc of my width)/5 into mRadius
+		end if
 		updatePerimeter()
 		updateLines()
 	end if


bwmilby:
Just saw @jameshale 's suggestion about making this a user editable property.  If that is the route that we want to take, I will cancel this PR.

jameshale:
After playing with both options I would indeed vote for a user settable property with a default of say 15. Having the radius a ratio of either the width or height puts a constraint on how one sets the dimensions of the widget. With the user being able to set the radius they are free to set the dimensions how they see fit.

jameshale:
If the adjustable radius is the way chosen please ensure the radius metadata only has a default and step value set. If you set a min/max value the PI will also present a scroller that ignores the step value.
See http://quality.livecode.com/show_bug.cgi?id=20686

livecodeali:
I agree it should probably be a property!

bwmilby:
@jameshale since Ali agrees, do you want to submit a PR with your fix?  I'm sure I could figure it out, but since you already have it done...

jameshale:
@bwmilby sure.

jameshale:
@bwmilby or @livecodeali 
not knowing the protocol here. does this require a bump in the version number?

livecodeali:
@jameshale We generally try to go by  https://semver.org/ - the relevant section being

>Given a version number MAJOR.MINOR.PATCH, increment the:
>1. MAJOR version when you make incompatible API changes,
>2. MINOR version when you add functionality in a backwards-compatible manner, and
>3. PATCH version when you make backwards-compatible bug fixes.

So I would increment the minor version here.

jameshale:
@livecodeali thanks

bwmilby:
Issue is being addressed in https://github.com/livecode/livecode/pull/6204
Closing this PR

PR #6229, opened 2017-12-18
remove any generated doc files after running the IDE

mwieder:
After compiling and launching the compiled build, the data directory is generated inside the resources directory. Git is not happy about this, so this patch deletes the generated files and the directory as part of the cleanup procedure before rebuilding on linux.

Note: apparently there were some stray trailing whitespaces cleaned up as well.


montegoulding:
@mwieder these would be better added to gitignore wouldn't they?

mwieder:
Don't know about "better", but that's another way to do it, and I like it. New PR submitted to the ide repo.

PR #6231, opened 2017-12-18
[[ Bug 20640 ]] [Docs] Update orientation docs for iPhone X

seaniepie:
iPhone X no longer supports upside-down portrait mode.


seaniepie:
For bug http://quality.livecode.com/show_bug.cgi?id=20640

montegoulding:
Thanks @seaniepie could you add a release note? Also as iphone X support was added to 8.1 we should rebase this onto develop-8.1. If you don't know how to do that one of us can do it once the release note is added.

PR #6305, opened 2018-01-30
[[ WIP ]] Add linux native button implementation

livecodeali:


livecodeali:
There are a couple of issues with this at the moment:
- Focus seems a little broken, perhaps something to do with the plug/socket setup. In particular, mousing over the widget causes a lighter appearance as expected, but then it never returns to normal after moving the mouse out. This seems related to another issue: clicking and then moving the mouse out of the rect causes the button to go back to normal, but then it does not respond to subsequent mouse over or clicks.
- Dragging the widget out when in browse mode causes an additional application window to appear at the top left, and LC crashes if you close it
- Switching to edit mode causes 'holes' in the stack briefly where the buttons are, before they get painted again.
- Not sure how relevant this is, but when closing a stack with button widget, I get 
```
Gtk-Critical **: IA_gtk_style_detach: assertion 'style->attach_count > 0' failed
```

PR #6320, opened 2018-02-05
Refactor MCStringGetCString into MCAutoStringRefAsCString

abrisan:


abrisan:
@runrevmark this should be ready for review, as we discussed offline

PR #6329, opened 2018-02-15
[Bug 20897] Ensure htmlText with listStyle=skip roundtrips

livecodepanos:


tests/lcs/core/field/htmlText.livecodescript
```diff
@@ -55,6 +55,31 @@ on TestSurrogateRoundtrip
...
+on TestListStyleRoundtrip
+   create stack "Test"
+   set the defaultStack to "Test"
+   
+   create field "TestField"
+   
+   set the text of field "TestField" to "one" & CR & "two" & CR & "three"
+   set the listStyle of line 1 of field "TestField" to "skip"
+   set the listStyle of line 2 of field "TestField" to "disc"
+   TestAssert "htmlText roundtrips listStyle beginning with skip", the htmlText of field "TestField" is the htmlText of field  "TestField"
```
montegoulding: @livecodepanos I'm not sure I get this test. Isn't the htmlText of the field always going to equal the htmlText of the field?

engine/src/fieldhtml.cpp
```diff
@@ -545,6 +551,10 @@ static void export_html_begin_lists(export_html_t& ctxt, uint32_t p_new_style, u
 		else
 			/* UNCHECKED */ MCStringAppendFormat(ctxt.m_text, "<li value=\"%d\">", p_index);
 	}
+	
+	// listStyle = "skip" does need a <li> tag as well
+	if (p_new_style == kMCParagraphListStyleSkip)
```
montegoulding, line 580: @livecodepanos is there a reason not to just drop the `p_new_style != kMCParagraphListStyleSkip` above?

engine/src/fieldhtml.cpp
```diff
@@ -513,10 +514,12 @@ static void export_html_end_lists(export_html_t& ctxt, uint32_t p_new_style, uin
 	if (ctxt . list_depth > 0 && !(p_new_depth == ctxt . list_depth && p_new_style == kMCParagraphListStyleSkip))
 		/* UNCHECKED */ MCStringAppendFormat(ctxt.m_text, "</li>\n");
 	
+	
 	while(p_new_depth < ctxt . list_depth)
 	{
 		ctxt . list_depth -= 1;
-		MCStringAppendFormat(ctxt.m_text, ctxt.list_styles[ctxt.list_depth] < kMCParagraphListStyleNumeric ? "</ul>" : "</ol>");
+		// listStyle = "skip" is unordered, thus needs </ul> tag
+		MCStringAppendFormat(ctxt.m_text, (ctxt.list_styles[ctxt.list_depth] < kMCParagraphListStyleNumeric || ctxt.list_styles[ctxt.list_depth] == kMCParagraphListStyleSkip) ? "</ul>" : "</ol>");
```
montegoulding: @livecodepanos these are now really long so perhaps a good opportunity to refactor out the ternary operator?

tests/lcs/core/field/htmlText.livecodescript
livecodepanos: Lol, you are right, this was definitely not what I meant to write!

engine/src/fieldhtml.cpp
livecodepanos: @montegoulding just dropping `p_new_style != kMCParagraphListStyleSkip ` causes data loss in this use-case:
```
put "one" & cr & "two" & cr & "three" into field 1
set the listStyle of line 1 of field 1 to "skip"
set the listStyle of line 2 of field 1 to "disc"
set the htmltext of field 1 to the htmltext
```
as in the first pass `ctxt . list_depth` equals `p_new_depth` thus the code in the `if` block will not be executed, but we do need a `<li>` tag in this case anyway I think

montegoulding: hmm... isn't that a general style of list changing but depth remaining the same issue? So style changes we should end the previous list and start a new list?

PR #6352, opened 2018-02-26
[[ Bug 20959 ]] Fixed bug cuasing crash when ungrouping

abrisan:


engine/src/stack2.cpp
```diff
@@ -816,6 +816,7 @@ void MCStack::ungroup(MCGroup *source)
 	cptr->remove(controls);
 	source->insertto(MCsavegroupptr);
 	source->setparent(this);
+    source->removereferences();
```
montegoulding, line 819: @livecodealex I'm fairly sure you don't want to call `removereferences` here. Probably what's needed here is in `X_close` there's a teardown of `MCsavegroupptr` so we would need this there.

```
        while (MCsavegroupptr != NULL)
	{
		MCControl *gptr = MCsavegroupptr->remove(MCsavegroupptr);
                 gptr->removereferences();
		delete gptr;
	}
```

PR #6357, opened 2018-02-27
[[ WidgetFocus ]] Implement widget focus

montegoulding:
This patch implements `focus {in | out | next | previous}`
in addition to re-implementing `OnFocusEnter` and `OnFocusLeave`
messages.


montegoulding:
This is going to need to remain WIP until we can do more complete native view focus management.

montegoulding:
Once https://github.com/livecode/livecode/pull/6506 is complete and merged then we need to grab the browser specific focus in when the JS forced it from here.

PR #6375, opened 2018-03-05
Extension examples

livecodeali:
Some extensions are merely for documentation purposes and there
are no current plans to bundle them with the IDE or put them in the
store. However we still want to build them under make check, as
they should be prevented from becoming bit-rotten.


montegoulding:
@livecodeali good idea... the ini lib can be put in there I think

livecodeali:
@montegoulding Only thing I'm not sure about with the ini lib is if dependencies are suitably worked out that it gets built for testing

livecodeali:
Can but try I guess!

PR #6430, opened 2018-03-16
 [[ ExtensionUtils ]] Move code to validate package to extension utils

livecodeali:

PR #6432, opened 2018-03-19
[[ ExtensionUtils ]] Add package validation

livecodeali:
Cherry-picked commits from PR targeted at wrong branch.


livecodeali:
Already checked here: https://github.com/livecode/livecode/pull/6430, as release-9.0.0 was identical to develop at time of checking.

PR #6496, opened 2018-04-21
refresh windows after gdk calls to fix linux backdrop problem

mwieder:
One-line change (moving the refresh call to after the gdk calls instead of before it) fixed the backdrop issue for me on linux mint 17.3. If this works for other distros then this should fix bug #17323.

### DIFF
--- a/engine/src/lnxdcs.cpp
+++ b/engine/src/lnxdcs.cpp
@@ -163,7 +163,7 @@ void gdk_event_fn(GdkEvent *p_event, gpointer dc)
     // Let GTK know about the event (needed so it can drive its Save As
     // dialogues and the like)
     gtk_main_do_event(p_event);
-    
+
     //fprintf(stderr, "gdk_event_fn\n");
     ((MCScreenDC*)dc)->EnqueueEvent(gdk_event_copy(p_event));
 }
@@ -207,19 +207,19 @@ Boolean MCScreenDC::open()
     initialise_required_weak_link_gdk();
     initialise_required_weak_link_gdk_pixbuf();
     initialise_required_weak_link_cairo();
-    
+
     gdk_init(0, NULL);
     //gdk_threads_init();
-    
+
     // Check to see if we are in a UTF8 locale
 	// TS : Changed 2008-01-08 as a more relaible way of testing for UTF-8
 	MCutf8 = (strcmp(nl_langinfo(CODESET), "UTF-8") == 0)	;
-	
+
 	MCimagecache = new (nothrow) MCXImageCache ;
-	
+
     if (MCdisplayname == NULL)
         MCdisplayname = gdk_get_display();
-    
+
 	if ((dpy = gdk_display_open(MCdisplayname)) == NULL)
 	{
         MCAutoStringRefAsSysString t_cmd;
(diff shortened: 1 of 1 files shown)


engine/src/lnxdcs.cpp
```diff
@@ -1375,30 +1375,33 @@ void MCScreenDC::enablebackdrop(bool p_hard)
...
+        // MDW-2018-04-21 [[ bugfix_17323 ]]
+        //MCstacks -> refresh();
         gdk_window_set_functions(backdrop, GdkWMFunction(0));
         gdk_window_set_decorations(backdrop, GdkWMDecoration(0));
         gdk_window_set_skip_taskbar_hint(backdrop, TRUE);
         gdk_window_set_skip_pager_hint(backdrop, TRUE);
 	gdk_window_move_resize(backdrop, 0, 0, device_getwidth(), device_getheight());
         gdk_window_lower(backdrop);
 	gdk_window_show_unraised(backdrop);
+    // MDW - refresh *after* the gdk calls
```
mwieder: Here's the one-line change.

montegoulding: @mwieder great! We could probably just move it after the else block and delete the call to refresh there.

montegoulding:
@mwieder thanks for looking into this! I'm wondering if you can amend this commit in the following ways:

- drop all the whitespace diffs
- delete rather than comment out the old call
- see my note about just calling refresh once after the else block to clean the code up
- add a release note
- change the commit message to something like:

```
[[ Bug 17323 ]] refresh windows backdrop created

This patch moves the call to `MCstacks->refresh()` to after the creation of the
presentation of the backdrop window to ensure that stack windows are raised
above it.
```

montegoulding:
BTW the following are your friends to do the above:
```
git reset --mixed HEAD^
git add -p
```

mwieder:
What's with all that whitespace? Seems like there was a bunch of trailing garbage after many lines.

bwmilby:
@mwieder Yes, many of the source files have extra white space.  The problem is that fixing all of the white space makes seeing what actually changed difficult and also can cause issues merging later on.

I just built this in my VM and I'm not seeing success (Ubuntu 16.04.3).  On launch, the menu is hidden.  I can make it appear by clicking on the Tools palette and then clicking on the background.

It also does not seem to address at least one issue that I've been able to make repeatable.  If I launch LiveCode, create a new stack, drag a new button onto it from the Tools palette, click on the stack then the tools palette disappears.

I have a Debian 9.4.0 VM with Gnome and the backdrop isn't usable at all there.  LC starts with just the backdrop.  Once I get the UI to display, the above steps will end up with just the new stack showing.  I have not been able to build LC on that VM (libicu:  `relocation R_X86_64_32S against symbol ... can not be used ... recompile with -fPIC`)  (Also, `uname -p` didn't work in the Makefile so I changed it to `-m` which does resolve to 64bit)

mwieder:
Brian- that's why I added the one-line comment to emphasize what had been changed <g>

I'm wondering why you're not seeing the backdrop change. The fix is working repeatably here for me. But I'm thinking this has more to do with the desktop manager than with the kernel. I'm using cinnamon over mint 17.3, which has an ubuntu 14.04 base.

I can verify that bug you're seeing, where clicking on the new stack hides the other LC windows, and clicking in the taskbar for LC toggles them back. But not with the fix I implemented.

And I am, of course, wondering why revmenubar is toplevel instead of palette, which would allow it to float at the same level as the tool palette instead of hiding behind other windows.

mwieder:
Heh - never mind that last bit. If the menubar is a palette it disappears from the taskbar.

mwieder:
unrelated tests failed.

mwieder:
travis build failed with unrelated errors.

PR #6502, opened 2018-04-23
[[ LCB ]] Additions to widget module functions

livecodeian:
Additions to builtin widget & emscripten modules to aid widget development:

- modify HandlerAsJSFunction to accept handlers with different signatures
- add optional `set` parameter to `property <name> of <scriptobject>` expression allowing access to values from specific custom property sets
- add `trigger all in <widget>` syntax to specify the widget object whose properties have changed
- add `this widget` expression to obtain a reference to the current widget object

Also fixes issues with resolving builtin foreign handlers in emscripten engine.

### DIFF
--- a/engine/src/module-engine.cpp
+++ b/engine/src/module-engine.cpp
@@ -265,7 +265,7 @@ bool MCEngineEvalObjectOfScriptObject(MCScriptObjectRef p_object, MCObject *&r_o
 	return true;
 }
 
-MCValueRef MCEngineGetPropertyOfObject(MCExecContext &ctxt, MCStringRef p_property, MCObject *p_object, uint32_t p_part_id)
+MCValueRef MCEngineGetPropertyOfObject(MCExecContext &ctxt, MCStringRef p_set, MCStringRef p_property, MCObject *p_object, uint32_t p_part_id)
 {
 	Properties t_prop;
 	t_prop = parse_property_name(p_property);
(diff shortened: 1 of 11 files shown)


engine/src/module-engine.cpp
```diff
@@ -316,11 +319,16 @@ extern "C" MC_DLLEXPORT_DEF void MCEngineExecGetPropertyOfScriptObject(MCStringR
...
+}
+
+extern "C" MC_DLLEXPORT_DEF void MCEngineExecGetPropertyOfScriptObject(MCStringRef p_property, MCScriptObjectRef p_object, MCValueRef& r_value)
+{
+	MCEngineExecGetPropertyOfSetOfScriptObject(p_property, kMCEmptyString, p_object, r_value);
 }
 
 // IM-2015-02-23: [[ WidgetPopup ]] Factored-out function for setting the named property of an object to a value.
-void MCEngineSetPropertyOfObject(MCExecContext &ctxt, MCStringRef p_property, MCObject *p_object, uint32_t p_part_id, MCValueRef p_value)
+void MCEngineSetPropertyOfObject(MCExecContext &ctxt, MCStringRef p_set, MCStringRef p_property, MCObject *p_object, uint32_t p_part_id, MCValueRef p_value)
```
livecodeali, line 331: I was going to ask if it was possible to restructure this so that we don't need to call `MCValueRelease` in each failure branch, but actually I think it's quite awkward either way :-(

engine/src/widget.lcb
```diff
@@ -1260,6 +1273,42 @@ end syntax
...
+-- may be called from another module library
+private handler EventCallback() returns nothing
+	-- update internal variables
+
+	-- notify ide of changes
+	trigger all in mSelf
+end handler
+
+Description:
+This widget evaluates to the current widget. This can be used to retain a reference for occasions where widget handlers may be called from another module, where the current widget may not be valid.
```
livecodeali, line 1301: Can you wrap this description so the lines don't exceed 80 chars please? It doesn't affect the rendering.

livecodeali:
I think most of this needs to go into develop except for 
[[ Emscripten ]] Fix unresolved foreign handlers in certain modules (6ec6c88)
[[ Emscripten ]] Skip foreign type constructor check on emscripten (0fb8d9d) 
which are fixes that we should probably have in develop-9.0. Could you cherry pick those and PR against develop-9.0 and then drop them from this patch and re-target at develop?

livecodeian:
develop-9.0 patches: https://github.com/livecode/livecode/pull/6541/files

livecodeian:
Split develop-9.0 patches into PR https://github.com/livecode/livecode/pull/6541
and develop additions into PR https://github.com/livecode/livecode/pull/6542

PR #6518, opened 2018-05-08
[15443] Ensure case-insensitive filter returns correct results

livecodepanos:


livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/6519

PR #6542, opened 2018-05-23
[[ LCB ]] Additions to widget module functions

livecodeian:
Additions to builtin widget & emscripten modules to aid widget development:

- modify HandlerAsJSFunction to accept handlers with different signatures
- add optional `set` parameter to `property <name> of <scriptobject>` expression allowing access to values from specific custom property sets
- add `trigger all in <widget>` syntax to specify the widget object whose properties have changed
- add `this widget` expression to obtain a reference to the current widget object


engine/src/engine.lcb
```diff
@@ -176,6 +178,7 @@ end syntax
 /**
 Summary:	The property of a script object.
 Property: 	The name of the property to manipulate
+Set:		The name of the custom property set to access
```
montegoulding, line 181: @livecodeian doesn't this need to be moved down to a doc block for `PropertyOfSetOfScriptObject`?

montegoulding: @livecodeian did the new syntax need to be documented separately or is this fine as is?

PR #6570, opened 2018-06-08
[19420] Prevent crash on Android when quitting and starting the app

livecodepanos:
`t_name` was `NULL`, causing null pointer dereference (and crash) when doing `t_name -> next`

### DIFF
--- a/libfoundation/src/foundation-name.cpp
+++ b/libfoundation/src/foundation-name.cpp
@@ -298,7 +298,7 @@ void __MCNameDestroy(__MCName *self)
 	// Find the previous link in the chain
 	__MCName *t_previous;
 	t_previous = nil;
-	for(__MCName *t_name = s_name_table[t_index]; t_name != self; t_name = t_name -> next)
+	for(__MCName *t_name = s_name_table[t_index]; t_name != nullptr && t_name != self; t_name = t_name -> next)
 		t_previous = t_name;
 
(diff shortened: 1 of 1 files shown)


livecodeali:
I'm on the fence about whether we should fix in this way, given it fixes a symptom rather than a cause - the cause being that the name table is destroyed while some MCName references hang around. OTOH, it is a crash. 

In any case, we need to fix the logic here slightly if we are going to merge it, looking at this as 'the incoming MCName is already gone from the name table'. As it stands, this might do something really bad
```
 	// Update the previous name's next field
 	if (t_previous == nil)
 		s_name_table[t_index] = self -> next;
 	else
 		t_previous -> next = self -> next;
```
And this shouldn't be done either:
```
 		// Reduce occupancy of the table
 		s_name_table_occupancy -= 1;
 
 		// If the table is too sparse, reduce its size (current heuristic has the
 		// threshold at 33%).
 		if (s_name_table_capacity > 1024 && s_name_table_occupancy * 16 / s_name_table_capacity < 5)
 			__MCNameShrinkTable();
```

I think it would also be worth keeping in an `MCLog` of the incoming name that's not found, as that will give us more clues of what needs to be cleaned up that isn't being.

livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/6574

PR #6605, opened 2018-07-04
Do not split up .framework folders between MacOS and Resources/_MacOS…

trevordevore:
… folder


montegoulding:
@trevordevore could you rebase this against develop-9.0?

PR #6603, opened 2018-07-03
[[ PlayerPanning ]] Add stereo balance & panning properties to mac player

livecodeian:

PR #6607, opened 2018-07-05
 [[ Bug 20211 ]] Fix crash in Mac player when changing filename repeatedly in a loop

livecodeian:


livecodepanos:
This PR was automatically merged once I merged https://github.com/livecode/livecode/pull/6603, since the latter already includes these commits

PR #6619, opened 2018-08-01
[[ Bug 10881 ]] Fix keyboard panning view when acceleratedRendering true

montegoulding:
This patch changes the use of `SOFT_INPUT_ADJUST_RESIZE` to
`SOFT_INPUT_ADJUST_NOTHING` when the keyboard is presented. The OpenGLView was
not resizing and as a result was being pushed up by the keyboard and then
not returning. Additionally it is unnecesary to resize and therefore reconfigure
the bitmap when the keyboard is presented.

### DIFF
--- a/docs/notes/bugfix-10881.md
+++ b/docs/notes/bugfix-10881.md
@@ -1,0 +1,1 @@
+# Fix view moving up when acceleratedRendering is true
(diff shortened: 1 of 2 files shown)


montegoulding:
Due to the fact that some apps may depend on the panning of the view when acceleratedRendering is true I hesitate to suggest inclusion of this patch without the inclusion of a patch to control it the behavior. Bug 10881 is nasty enough to warrant inclusion of a fix ASAP though so perhaps both should target 9.0.2?

https://quality.livecode.com/show_bug.cgi?id=21467

montegoulding:
Using adjust nothing causes issues with working out if the keyboard was displayed. This patch should be reverted and updated to `getWindow().setGravity(Gravity.TOP)`

montegoulding:
The gravity thing unfortunately does not work correctly. I have experimented with implementing an `onMeasure` method in `OpenGLView` and there is some hope there but it will require some time to get right. Closing this patch for now.

PR #6628, opened 2018-08-06
[[ Bug 21467 ]] Implement mobileSetKeyboardDisplay

montegoulding:
This patch implements a new command `mobileSetKeyboardDisplay` and function
`mobileGetKeyboardDisplay` to govern the behavior of the keyboard on mobile
to pan the view up if necessary so the focused object is visible.

### DIFF
--- a/docs/dictionary/command/mobileSetKeyboardDisplay.lcdoc
+++ b/docs/dictionary/command/mobileSetKeyboardDisplay.lcdoc
@@ -1,0 +1,40 @@
+Name: mobileSetKeyboardDisplay
+
+Type: command
+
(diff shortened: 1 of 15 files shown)


montegoulding:
Should be merged into the same branch as https://github.com/livecode/livecode/pull/6619

montegoulding:
This patch needs to be use adjust resize. See the comment on #6619. Additionally on android pan mode we will not be able calculate the `effective working screenrect` correctly as the window is not actually resized.

montegoulding:
Removing WIP as pan mode is not impacted by bug 10881 it is a helpful workaround.

PR #6693, opened 2018-09-18
[[ BUILD ]] Disable PIE when linking executables on Linux

livecodeian:
This option is enabled by default when building on Ubuntu 18.04LTS,
which causes problems when linking against libraries built without PIE
support.

### DIFF
--- a/config/linux-settings.gypi
+++ b/config/linux-settings.gypi
@@ -39,6 +39,15 @@
 			},
 		],
 		[
+			'_type == "executable"',
+			{
+				'ldflags':
+				[
+					'-no-pie',
+				],
+			},
+		],
+		[
 			'server_mode == 0',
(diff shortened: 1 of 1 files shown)


montegoulding:
@livecodeian you might need to set an environment variable in the config script and then use that in the gyp to govern whether to add the flag. See https://stackoverflow.com/questions/12584936/gcc-ignore-unrecognized-option

livecodeian:
Yeah, I think you're right @montegoulding - if I just pass that flag to gcc < 6 it'll just error out.

bwmilby:
So this is why I couldn’t build (link) on the laptop I configured.

livecodeali:
Superseded by  https://github.com/livecode/livecode/pull/6953

PR #6738, opened 2018-10-18
[[ Tree View Widget ]] Auto unfold to show selected row

bwmilby:
The tree view widget now has the ability to automatically expand to reveal
a row when it is selected. If **scrollHilitedElementIntoView** is not `true`
then the scroll position will be adjusted to maintain the currently visible
top row.

The tree view fold state can now be reset in two ways. To set the array
data in a folded state then use the **foldedArrayData** property. To
collapse the tree without changing the data, get the **foldedArrayData**.

Added 2 module variables to control redraw and determine when fold state
changes (`mRedraw` and `mFoldChanged`).

Added `pathIsValid` handler to ensure that a user supplied path actually
exists.  Validation is now done in the `setSelectedElement` handler so the
requirement to test for an empty path in `selectPath` is removed.

Added `unfoldFullPath` handler to take care of ensuring the full selection
path is unfolded.  Code is included to maintain the top visible row if
`mScrollHilitedElementIntoView` is not `true`.

Updated `applyFoldStateKey` handler to set `mFoldChanged` when an expansion
is performed.  Updated to check for the need to expand.

### DIFF
--- a/extensions/widgets/treeview/notes/feature-auto_expand.md
+++ b/extensions/widgets/treeview/notes/feature-auto_expand.md
@@ -1,0 +1,10 @@
+# Properties
+
+The tree view widget now has the ability to automatically expand to reveal
+a row when it is selected. If **scrollHilitedElementIntoView** is not `true`
+then the scroll position will be adjusted to maintain the currently visible
+top row.
+
+The tree view fold state can now be reset in two ways. To set the array
+data in a folded state then use the **foldedArrayData** property. To
+collapse the tree without changing the data, get the **foldedArrayData**.
--- a/extensions/widgets/treeview/notes/21361.md
+++ b/extensions/widgets/treeview/notes/21361.md
@@ -1,0 +1,1 @@
+# [21361] Add option to reset fold state when setting arrayData
--- a/extensions/widgets/treeview/treeview.lcb
+++ b/extensions/widgets/treeview/treeview.lcb
@@ -83,7 +83,7 @@ use com.livecode.library.widgetutils
 use com.livecode.foreign
 
 metadata author is "LiveCode"
-metadata version is "2.1.0"
+metadata version is "2.2.0"
 metadata title is "Tree View"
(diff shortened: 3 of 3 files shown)


extensions/widgets/treeview/notes/feature-auto_expand.md
```diff
@@ -0,0 +1,10 @@
+# Properties
+
+The tree view widget now has the ability to automatically expand to reveal
+a row when it is selected. If **scrollHilitedElementIntoView** is not `true`
+then the scroll position will be adjusted to maintain the currently visible
+top row.
+
+The tree view fold state can now be reset in two ways. To set the array
```
montegoulding, line 8: @bwmilby My first inclination was to think this property allowed control over the folded state over the entire tree but it seems it just clears the folding. I'm wondering if this should be a boolean property. Either everything is folded or it's not. So:
```
set the folded of widget "tree" to true -- folds it all
get the folded of widget "tree" -- returns true if it's entirely folded
```
The only issue is what to do with `set the folded of widget "tree" to false` if there isn't currently an unfolded node... perhaps just unfold the first node in sort order?

bwmilby: This is actually kind of a hack to get around the lack of directly calling handlers.  So I had to encode what would be a parameter into the property name.  The reasoning is to allow the array data to be replace and also reset the fold state in a single operation.  This prevents the applying of the fold state and the related calculations when the end desire is to have it folded.  The bug report was requesting the ability to reset the fold state when setting the array data.

bwmilby: Possibly `arrayDataAndFoldAll` for the property name?

bwmilby:
This PR can be closed.  Updates made to PR #6764 that already included these commits due to the way I branched.  **foldedArrayData** removed and **autoFoldStateReset** added.  PR #6770 addresses manual setting of the fold state array.

bwmilby:
Replaced by PR #6764

PR #6757, opened 2018-10-23
[[Docs]] Update Filter to make better distinction for regEx

seaniepie:


docs/dictionary/command/filter.lcdoc
```diff
@@ -111,12 +108,17 @@ You can match instances of special chars as follows:
...
+bracket is not a special character.
 
 The three bracketed forms can be combined to create more complex
 character classes, for example the pattern [!abcA-C] matches any
 character which is not a, b or c (upper or lower case)
 
+If 'regex pattern' is specified, the <filterPattern> evaluates to a
+regular expression.
+
 If no <filterPattern> type is specified, the <filterPattern> is handled
```
montegoulding: @seaniepie what do you think of highlighting this by making it a `>**Note:**`

docs/dictionary/command/filter.lcdoc
```diff
@@ -71,20 +71,17 @@ Description:
 Use the <filter> command to pick specific lines, items, keys or elements
 in a container or expression.
 
-The <filter>...with form and the <filter>...matching form retain the
+The '<filter>...with' form and the '<filter>...matching' form retain the
```
montegoulding: @seaniepie could you use back ticks to make these render as code. Would need to drop the `<>` around filter as that causes the parser to mess with the word.

docs/dictionary/command/filter.lcdoc
```diff
@@ -71,20 +71,17 @@ Description:
...
+The '<filter>...without' form and the '<filter>...not matching' form discard
 the lines, items, keys or elements that do not contain a match for the
 specified <filterPattern>. 
 
 If you don't specify lines, items, keys or elements then lines are
 filtered by default.
 
-If a regex pattern is specified, the <filterPattern> evaluates to a
-regular expression.
-
```
montegoulding: @seaniepie in the section below describing the wildcard special characters could you wrap them all in back ticks? The `- *` one in particular is causing the markdown renderer to render an extra indent on the unordered list.

docs/dictionary/command/filter.lcdoc
```diff
@@ -111,17 +113,22 @@ You can match instances of special chars as follows:
...
 character classes, for example the pattern [!abcA-C] matches any
 character which is not a, b or c (upper or lower case)
 
-If no <filterPattern> type is specified, the <filterPattern> is handled
-as a wildcard.
+>**Note:** If the <paternType> `regex pattern` is specified, 
+the <filterPattern> evaluates to a <regular expression>.
+>
+>If no <patternType> is specified, the <filterPattern> is handled
+as a `wildcard pattern` type.
```
seaniepie: I've joined the two together as part of the Note

docs/dictionary/command/filter.lcdoc
```diff
@@ -56,49 +56,51 @@ Parameters:
 filterSource:
 An expression that evaluates to a string, array, or a container.
 
+patternType:
+Either `wildcard pattern` or `regex pattern`.
```
seaniepie: @montegoulding , is the wording here ok to you? Having it here in the parameters should draw more attention to it and it's explicit nature perhaps.

bwmilby: I think that the original suggestion for the syntax is more accurate.  The patternType is only optional if `wildcard pattern` is the desired value.  Otherwise it is a required value.  A secondary consideration is that AutoComplete may no longer be able to suggest the pattern type with this change.

```
Syntax: filter [{lines | items | keys | elements} of] <filterSource> {with | without | [not] matching} {[wildcard pattern] | regex pattern} <filterPattern> [into <targetContainer>]
```

seaniepie: Hmm 🤔. I’d like to defer up the chain on this one. It’s a good point and may well be the correct one. What are the thoughts. @runrevmark @livecodepanos @livecodeali @montegoulding anyone who wants to pipe in.

livecodeali: I can definitely see the logic, although I'm not exactly convinced the syntax element should be used to imply something. Personally in terms of technical specification I find the existing syntax less confusing than the proposed new one, and would prefer the default behavior to be documented elsewhere (which I think you've done)

seaniepie: Thanks @livecodeali. What about the difference between the original `[{wildcard pattern | regex pattern}]` and @bwmilby's suggestion of `{[wildcard pattern] | regex pattern}` which seems to have a more accurate logic to it?

livecodeali: That's what I mean - to me the syntax element is just an expression of valid syntax in the language and shouldn't necessarily try to convey meaning in the way suggested - there's no actual difference between them in terms of what syntax is valid. 

I guess if people find `{[wildcard pattern] | regex pattern}` more useful than `[{wildcard pattern | regex pattern}]` then it's fair enough to make the change! As long as it doesn't have any negative consequences for autocomplete (I don't see why it would)

livecodeali: On the other hand I guess we do split out syntactic variants into 'types' in some entries.
```
Syntax: filter [{lines | items | keys | elements} of] <filterSource> {with | without | [not] matching} [wildcard pattern] <filterPattern> [into <targetContainer>]
Syntax: filter [{lines | items | keys | elements} of] <filterSource> {with | without | [not] matching} regex pattern <filterPattern> [into <targetContainer>]
```
is also a possibility

seaniepie: Ooo, now that is much better.

montegoulding: @seaniepie I agree that splitting them is better. The filter command will only get more complicated when my patch for the `where` clause using the `each` global is merged https://github.com/livecode/livecode/pull/6626

docs/dictionary/command/filter.lcdoc
```diff
@@ -108,22 +113,22 @@ You can match instances of special chars as follows:
...
 The three bracketed forms can be combined to create more complex
 character classes, for example the pattern [!abcA-C] matches any
 character which is not a, b or c (upper or lower case)
 
-If 'regex pattern' is specified, the <filterPattern> evaluates to a
-regular expression.
-
-If no <filterPattern> type is specified, the <filterPattern> is handled
-as a wildcard.
+>**Note:** If the <paternType> `regex pattern` is specified,
```
montegoulding: @seaniepie the `<>` should only be used for parameters or doc referenced docs. I'm thinking it's better to make just the bit about the no type specified the note. That's the bit that is easy to miss.

docs/dictionary/command/filter.lcdoc
montegoulding: Actually splitting them gives us the option to use `<wildcardPattern>` and `<regularExpression>` on the different lines as the parameter names and document them separately.

docs/dictionary/command/filter.lcdoc
seaniepie: @montegoulding I think your looking at an older commit. Jump to the latest and mark which ones you think are still valid. Particularly the one about having wildcard and regex as parameters. Are they really parameters as they are more like keywords than a placeholder for a value.

docs/dictionary/command/filter.lcdoc
montegoulding: Aha... I must have been looking at an older commit. I still think the block quoted note should _just_ highlight the no pattern type thing. Also I think the idea of splitting `<filterPattern>` into separate `<wildcardPattern>` and `<regularExpression>` parameters will help significantly. In that case you might change the note to:
```
If neither `wildcard pattern` or `regex pattern` are specified then the pattern is handled as a `wildcard pattern`.
```

montegoulding: Perhaps `<regexPattern>` would be better to match `regex pattern` ?

bwmilby: I like `<regexPattern>` to go with `<wildcardPattern>` as the two options unless you want to be more verbose with `<regexFilterPattern>` and `<wildcardFilterPattern>`.  I agree that since the syntax is now split into two forms, that the parameter should get unique names.

You could also refer to the syntax as the `regex pattern form` and `wildcard pattern form` throughout the document.

Here is another alternative for the note:
```
If the regex pattern form is used (`regex pattern`) then the filter pattern is a <regexPattern>.  
If the wildcard pattern form is used (`wildcard pattern` or the pattern is not specified) then the 
filter pattern is a <wildcardPattern>.
```

seaniepie: Ohhh 😂I see what you mean now. I thought you were talking about my now redundant patternType. But you mean have, instead of filterPattern, these other two parameters listed. I’ll draft it up in the morning (uk time) for everyone to look at. Thanks Monte.

seaniepie: @bwmilby, the verbose names in the already long syntax lines will likely be too long and unnecessary. The parameter description can probably make it clearer that they are filter patterns. 
I’ll look at the note with fresh eyes later and see if there is a simple approach that suits all of these ideas and bears in mind the upcoming changes to the command. Thanks Brian.

bwmilby:
Looks good!  Thanks for working on this.

docs/dictionary/command/filter.lcdoc
```diff
@@ -56,49 +57,60 @@ Parameters:
...
 An expression that evaluates to a string, array, or a container.
 
-filterPattern:
+wildcardPattern:
 An expression used to match certain lines, items, keys or elements.
+See description.
+
+regexPattern:
+An regular expression used to match certain lines, items, keys or elements.
+See [Microsofts RegEx Quick Reference](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference)
```
montegoulding: @seaniepie Microsoft's

seaniepie: Thanks :)

montegoulding:
@livecode-vulcan review ok 6f65b43

Thanks @seaniepie!

livecode-vulcan:
💙 review by @montegoulding ok 6f65b436d34d0ea4fd262f87b56b0cd3316bc7cf

livecode-vulcan:
😎 test success 6f65b436d34d0ea4fd262f87b56b0cd3316bc7cf

* try-community-armv6-android-sdk26_ndk16r15: [success](https://vulcan.livecode.com/builders/try-community-armv6-android-sdk26_ndk16r15/builds/192)
* try-community-armv7-android-sdk26_ndk16r15: [success](https://vulcan.livecode.com/builders/try-community-armv7-android-sdk26_ndk16r15/builds/184)
* try-community-arm64-android-sdk26_ndk16r15: [success](https://vulcan.livecode.com/builders/try-community-arm64-android-sdk26_ndk16r15/builds/184)
* try-community-x86-android-sdk26_ndk16r15: [success](https://vulcan.livecode.com/builders/try-community-x86-android-sdk26_ndk16r15/builds/184)
* try-community-x86_64-android-sdk26_ndk16r15: [success](https://vulcan.livecode.com/builders/try-community-x86_64-android-sdk26_ndk16r15/builds/184)
* try-community-js-emscripten-sdk1.35: [success](https://vulcan.livecode.com/builders/try-community-js-emscripten-sdk1.35/builds/1706)
* try-community-universal-ios-iphoneos12.0: [success](https://vulcan.livecode.com/builders/try-community-universal-ios-iphoneos12.0/builds/88)
* try-community-universal-ios-iphonesimulator12.0: [success](https://vulcan.livecode.com/builders/try-community-universal-ios-iphonesimulator12.0/builds/88)
* try-community-universal-mac-macosx10.9: [success](https://vulcan.livecode.com/builders/try-community-universal-mac-macosx10.9/builds/1736)
* try-community-x86-linux-debian8: [success](https://vulcan.livecode.com/builders/try-community-x86-linux-debian8/builds/1729)
* try-community-x86_64-linux-debian8: [success](https://vulcan.livecode.com/builders/try-community-x86_64-linux-debian8/builds/1726)
* try-community-x86-win32: [success](https://vulcan.livecode.com/builders/try-community-x86-win32/builds/1762)
* try-community-x86_64-win32: [success](https://vulcan.livecode.com/builders/try-community-x86_64-win32/builds/1703)

PR #6774, opened 2018-11-01
Ensure "mailto:" works in the browser widget on iOS

livecodepanos:
`canHandleRequest` returns false for requests with the `mailto:` scheme on WebViews, so this patch handles this case separately. Tested on an iPad.


montegoulding:
Hmm... is that a bug? I would have thought it is appropriate to send `browserUnhandledLoadRequest` in this case and then the user can decide to use `launch url` if they want to handle urls with apps outside the browser.

livecodepanos:
You are right @montegoulding. I am closing this PR as the current behavior of the browser widget on iOS is the expected one.

PR #6787, opened 2018-11-12
[[ Is Strictly Equals ]] Implement is strictly equals

montegoulding:
This patch enhances the `is strictly` operator to allow for strict value
equivalence. Previously `is strictly` could be used to only test if the
expression evaluated to either a particular type or nothing. This patch
allows `is strictly` to also compare exact value equivalence.

### DIFF
--- a/engine/src/exec-logic.cpp
+++ b/engine/src/exec-logic.cpp
@@ -233,6 +233,42 @@ void MCLogicEvalIsNotEqualTo(MCExecContext& ctxt, MCValueRef p_left, MCValueRef 
 	ctxt . Throw();
 }
 
+static bool MCLogicIsStrictlyEqualTo(MCValueRef p_left, MCValueRef p_right)
+{
+    /* Take into account names */
+    MCValueTypeCode t_left_code, t_right_code;
+    t_left_code = MCValueGetTypeCode(p_left);
+    t_right_code = MCValueGetTypeCode(p_right);
+    
+    if (t_left_code != t_right_code)
+    {
+        if (t_left_code == kMCValueTypeCodeName)
+        {
+            p_left = MCNameGetString((MCNameRef)p_left);
+        }
+        else if (t_right_code == kMCValueTypeCodeName)
+        {
+            p_right = MCNameGetString((MCNameRef)p_right);
+        }
+        else
+        {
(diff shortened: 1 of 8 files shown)


engine/src/exec-logic.cpp
```diff
@@ -233,6 +233,16 @@ void MCLogicEvalIsNotEqualTo(MCExecContext& ctxt, MCValueRef p_left, MCValueRef
 	ctxt . Throw();
 }
 
+void MCLogicEvalIsStrictlyEqualTo(MCExecContext& ctxt, MCValueRef p_left, MCValueRef p_right, bool& r_result)
+{
+    r_result = MCValueIsEqualTo(p_left, p_right);
```
runrevmark: This actually needs to be:
```

static bool MCLogicIsStrictlyEqualTo(MCValueRef p_left, MCValueRef p_right)
{
    /* Take into account names */
    MCValueTypeCode t_left_code, t_right_code;
    t_left_code = MCValueGetTypeCode(p_left);
    t_right_code = MCValueGetTypeCode(p_right);

    if (t_left_code != t_right_code)
    {
        if (t_left_code == kMCValueTypeCodeName)
        {
            p_left = MCNameGetString((MCNameRef)p_left);
        }
        else if (t_right_code == kMCValueTypeCodeName)
        {
            p_right = MCNameGetString((MCNameRef)p_right);
        }
        else
        {
            return false;
        }
    }
    
    return MCValueIsEqualTo(p_left, p_right);
}

void MCLogicEvalIsStrictlyEqualTo(MCExecContext& ctxt, MCValueRef p_left, MCValueRef p_right, bool& r_result)
{
    r_result = MCLogicIsStrictlyEqualTo(p_left, p_right);
}

void MCLogicEvalIsNotStrictlyEqualTo(MCExecContext& ctxt, MCValueRef p_left, MCValueRef p_right, bool& r_result)
{
    r_result = !MCLogicIsStrictlyEqualTo(p_left, p_right);
}
```
Otherwise you get oddities such as `"0" is not strictly (0 & "")` (the engine tries to store constant strings as names internally wherever it can).

montegoulding:
Rebased, added docs and release notes

PR #6790, opened 2018-11-13
 [[Bugfix 21183]] [Docs] Resources SpecialFolderPath behaviour

seaniepie:
Added a note for specialFolderPath("resources") behaviour on macOS before and after the current stack has been saved.

### DIFF
--- a/docs/dictionary/function/specialFolderPath.lcdoc
+++ b/docs/dictionary/function/specialFolderPath.lcdoc
@@ -90,9 +90,9 @@ The following <folderIdentifier> values are supported:
   * "temporary": The folder where temporary files can be placed
   * "engine": The folder containing the LiveCode engine and the
     executable files copied in the standalone application
-  * "resources": In development mode, the current stack's folder.  In a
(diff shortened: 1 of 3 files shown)


docs/dictionary/function/specialFolderPath.lcdoc
```diff
@@ -102,8 +102,13 @@ The following <folderIdentifier> values are supported:
   * 0x000a: The "Recycle Bin".
 
 
-**On Mac OS systems**, the following <folderIdentifier> values 
-are supported: 
+**On Mac OS systems**, when there are no user stacks loaded into memory or
```
seaniepie, line 105: We need to check if this only affects the macOS platform or if it also affects Windows and Linux environments also.

montegoulding: @seaniepie yes it will be the same on all platforms. When there are no user stacks one of the IDE stacks will be `the defaultStack` so `specialFolderPath("resources")` will be relative to that. So probably this should be a non-platform specific note.

PR #6793, opened 2018-11-19
[[ ByteOffset ]] Fixed bug causing incorrect results

abrisan:


livecodepanos:
@livecodealex I think this PR should target `develop-9.0`, as it fixes a regression bug

abrisan:
@livecodepanos tadaa https://github.com/livecode/livecode/pull/6794

PR #6807, opened 2018-11-29
Reverted IDE pointer to check if it passes Travis tests

livecodepanos:


livecodepanos:
Ok, this passed the test, so we now know the commit in the IDE repo that caused the problem.

PR #6843, opened 2019-01-08
[21778] Fix sort international when system locale is non-en

livecodepanos:


livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/6846

PR #6844, opened 2019-01-09
[[ Intents ]] Implement OnActivityResultListener

montegoulding:
This patch implements an `OnActivityResultListener` interface so it is possible to
handle `onActivityResult` for a particular result code.

This implementation should be icnsidered temporary until bug 19920 (EngineAPI class)
is implemented at which point `OnActivityResultListener` would likely become an
interface of `EngineAPI` rather than `LiveCodeActivity`.

### DIFF
--- a/docs/lcb/notes/feature-jobjectarray.md
+++ b/docs/lcb/notes/feature-jobjectarray.md
@@ -1,0 +1,10 @@
+# LiveCode Builder Standard Library
+
+## Java Utilities
+
+Syntax for converting between a JObjectArray and a List of JObjects have been
+added to the Java utility library.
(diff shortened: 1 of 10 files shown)


montegoulding:
@livecodeali Do you think creating an abstraction for the activity listener and object array to list in `com.livecode.library.androidutils` is sufficient for now and then we can move stuff to an engine api class later? That also gives me somewhere to document these.

montegoulding:
Perhaps object array to list should go in `com.livecode.java`

montegoulding:
@livecodeali have gone ahead with the above and added docs & notes

PR #6848, opened 2019-01-10
[21562] Ensure the dragimage is shown on Windows

livecodepanos:


livecodepanos:
This patch is incorrect because it is likely to introduce a memory leak, and more importantly, does not make the dragImage to survive outside the LC stack (this worked in previous versions)

PR #6865, opened 2019-01-31
[15157] Replace '-' with '_' in android standalone identifier

livecodepanos:


docs/notes/bugfix-15157.md
```diff
@@ -0,0 +1 @@
+# Replace '-' with '_' in android standalone identifier
```
montegoulding, line 1: @livecodepanos The markdown isn't rendering correctly. I think you need to replace `'` with back-tick otherwise the `_` is recognised as the beginning of an emphasised chunk

ide-support/revsaveasandroidstandalone.livecodescript
```diff
@@ -1034,6 +1036,11 @@ private function escapeLabel pLabel
    replace quote with "\" & quote in pLabel
    return pLabel
 end escapeLabel
+
+private function legalizeIdentifier pIdentifier
```
montegoulding, line 1040: @livecodepanos The docs say:
> A full Java-language-style package name for the Android app. The name may contain uppercase or lowercase letters ('A' through 'Z'), numbers, and underscores ('_'). However, individual package name parts may only start with letters.

So my gut feeling is that we should have a validation function which is checked on entry and throws up an error dialog if invalid and then when building a standalone we should re-check and just throw an error and stop building.

I think the following would correctly check the identifier is valid:
```
function validIdentifier pIdentifier
   return matchText(pIdentifier, "(?i)^[a-z][a-z0-9_]*(\.[a-z][a-z0-9_]*)+$")
end validIdentifier
```

livecodepanos: @montegoulding it seems we already had a function that checked if the identifier is valid, but the check was not sufficient, as it allowed `-`. I have modified this function as you suggested. A PR in the IDE repo is coming

livecodepanos:
Replaced by https://github.com/livecode/livecode-ide/pull/2039

PR #6873, opened 2019-02-12
[21836] Update mouseLoc,mouseH,mouseV in scrollerBeginDrag/scrollerEndDrag

livecodepanos:


engine/src/mbldc.cpp
```diff
@@ -908,11 +908,21 @@ uint16_t MCScreenDC::platform_getheight()
...
+        // IM-2014-03-03: [[ Bug 11836 ]] Mouse loc and window position now stored in logical coords
+        r_x = m_mouse_x + m_window_left;
+        r_y = m_mouse_y + m_window_top;
+    }
+    else
+    {
+        r_x = t_x;
+        r_y = t_y;
+    }
 }
```
livecodeian, line 922: @livecodepanos I think there's an issue with this approach - we want mousex & mousey to represent the x & y coordinates of the last touch event. With the logic as implemented once the scroll is over, those values will revert to what they were before the scroll began.

 I suggest doing away with MCSystemUpdateMouseCoords (the name is misleading) and instead updating MCScreenDC::m_mouse_x and MCScreenDC::m_mouse_y in mbldc.h by adding an update method to MCScreenDC that accepts the new coords and calling that from the platform-specific scroller implementations

livecodepanos: Thanks @livecodeian.

I am setting this to WIP

engine/src/mbliphonescroller.mm
```diff
@@ -883,21 +888,32 @@ - (id)initWithInstance:(MCiOSScrollerControl*)instance
...
 - (void)scrollViewWillBeginDragging: (UIScrollView*)scrollView
 {
 	MCCustomEvent *t_event;
 	t_event = new MCiOSScrollerEvent(m_instance, MCM_scroller_begin_drag);
+    
+    CGPoint touchPoint = [scrollView.panGestureRecognizer locationInView:scrollView];
+
+    MCscreen -> device_updatemouse((int32_t)touchPoint.x,(int32_t)touchPoint.y);
+    
 	MCEventQueuePostCustom(t_event);
```
livecodeian, line 904: @livecodepanos can you make sure the coordinates you get are relative to the top-left of the main view.

engine/src/java/com/runrev/android/nativecontrol/ScrollerControl.java
```diff
@@ -153,7 +153,10 @@ public boolean dispatchTouchEvent(MotionEvent e)
                     {
                         m_dragging = true;
                         m_tracking = false;
-                        onScrollBeginDrag();
+                        
+                        int x = (int)e.getX();
+                        int y = (int)e.getY();
+                        onScrollBeginDrag(x,y);
                         e.setAction(MotionEvent.ACTION_CANCEL);
```
livecodeian, line 160: @livecodepanos can you check if these coordinates are relative to the top-left of the main view?

livecodepanos: Hmm, looks like they are not. I will investigate.

engine/src/mbliphonescroller.mm
livecodepanos: Same here, it looks like they are not. I will investigate.

PR #6874, opened 2019-02-14
[Bugfix-13506] Remove outdated information from create entry

livecodealice:
Removed a paragraph regarding the template object reseting after using the create command as it does not appear to be true.

Note: It was commented on (privately) that the template object *does* reset after the next event dispatch point and example code was given but that wasn't reproduced in LC 9.0.2.

### DIFF
--- a/docs/dictionary/command/create.lcdoc
+++ b/docs/dictionary/command/create.lcdoc
@@ -66,13 +66,6 @@ If you use the invisible form, the object is created with its visible
 property set to false, so it cannot be seen. Use this form to create a 
 hidden object, change its appearance or position, then make it visible.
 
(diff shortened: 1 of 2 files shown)


docs/dictionary/command/create.lcdoc
```diff
@@ -69,26 +69,9 @@ hidden object, change its appearance or position, then make it visible.
...
-    create button
-    end repeat
-    unlock messages
-
-
-LiveCode resets the template only when in the development environment, 
-not in standalone applications.
+tool:
+
+    send "choose browse tool" to me in 1 tick
```
montegoulding: @livecodesam I think this is also no longer needed.

PR #6897, opened 2019-03-07
added -fPIC and -fno-pie options to allow compiling on modern linux d…

mwieder:
Changes to two files to allow compiling on modern linux systems. Gcc since version 6.2 (my system now defaults to version 7.3.0) now defaults to the "-enable-default-pie" option for position independence, which prevents compilation. The patch to config.py forces "-fno-pie" as a compiler option instead and "-no-pie" as a linker option. The cef library compilation still fails unless the second patch to linux-settings.gpyi to force the "-fPIC" option is also installed.

### DIFF
--- a/config.py
+++ b/config.py
@@ -379,13 +379,13 @@ def configure_toolchain(opts):
     if opts['CC_PREFIX'] is not None:
         ccprefix = opts['CC_PREFIX']
     if opts['CC'] is None:
-        opts['CC'] = ccprefix + 'cc'
+        opts['CC'] = ccprefix + 'cc -fno-pie'
     if opts['CXX'] is None:
-        opts['CXX'] = ccprefix + 'c++'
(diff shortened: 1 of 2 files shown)


livecodeian:
Hi @mwieder, I have a similar PR here - https://github.com/livecode/livecode/pull/6693 however it can't be merged right now as the `-fno-pie` option is not recognised by the version of GCC used to build our releases.

One thing that we may be able to do to be more backwards-compatible is to make sure our prebuilt libraries are compiled with PIC enabled so they can be linked with versions of GCC with PIE enabled by default.

Ian.

mwieder:
Ian- thanks. Yes, your patch does the trick nicely and changes only one file instead of two. I'm going to close this PR in favor of #6693. Is there a reason the team is still building with outdated software?

PR #6899, opened 2019-03-08
The tagged words disappear in the Dictionary

PaulMcClernan:
Condition & ExpectsPreconditionWithReason disappear when viewed in the dictionary.
Not sure what the reason for Condition not working as a tag is so I just removed the enclosure.
ExpectsPreconditionWithReason has an 's' after Expect.
ExpectPreconditionWithReason should fix it.

### DIFF
--- a/libscript/src/assert.lcb
+++ b/libscript/src/assert.lcb
@@ -61,12 +61,12 @@ may require that its parameters have a particular structure or fall
 within a particular range, or may need the system to be in a specific
 state.
 
-Note that if the <Condition> is false, this statement will throw an
+Note that if the Condition is false, this statement will throw an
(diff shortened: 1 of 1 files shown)


libscript/src/assert.lcb
```diff
@@ -61,12 +61,12 @@ may require that its parameters have a particular structure or fall
 within a particular range, or may need the system to be in a specific
 state.
 
-Note that if the <Condition> is false, this statement will throw an
```
montegoulding, line 64: @PaulMcClernan the issue is in the `Parameters` section there is a tab before `Condition`. It looks like the same issue is in the `ExpectPreconditionWithReason` docs below.

PR #7017, opened 2019-05-08
[[ NO MERGE ]] Try Travis with Windows config

livecodeali:


.travis.yml
```diff
@@ -88,9 +90,15 @@ script: |
       export XCODEBUILD_FILTER="| xcpretty"
       export JAVA_HOME=$(/usr/libexec/java_home)
       ;;
+    win)
```
montegoulding: `windows` here I think

montegoulding:
probably need to do something about the compiler matrix so it only builds win once

PR #7050, opened 2019-05-15
[[ Widget ]] Add ability to specify property set when accessing properties

livecodeian:
This patch adds the syntax `property <prop> of set <set> of <object>`
to specify which object property set to fetch a value from.


engine/src/engine.lcb
```diff
@@ -181,6 +183,7 @@ end syntax
 /**
 Summary:	The property of a script object.
 Property: 	The name of the property to manipulate
+Set:		The name of the custom property set to access
```
montegoulding, line 186: @livecodeian I don't think a single doc for multiple syntax blocks is supported by the docs parser. I don't think it will show the syntax properly in the dictionary so you may need to duplicate the docs here for the two variants.

PR #7073, opened 2019-05-30
[[ Cleanup ]] Cleanup unexpected passes

livecodeali:
This patch cleans up some unexpected test passes by changing the relevant calls to `TestAssertBroken` to calls to `TestAssert`


tests/lcs/core/interface/interfaceui.livecodescript
```diff
@@ -255,39 +255,36 @@ TestAssert "test", (the resizeable of this stack is false)
 end TestInterface63
 
 on TestBug11799Part1
- 
+   create stack "Test"
```
montegoulding, line 258: @livecodeali This and the next should probably have `TestSkipIfNot "wait"`

PR #7090, opened 2019-06-07
[[ LCIDLC ]] Fix build issue

montegoulding:
This patch removes the sources block of the lcidlc `encode_support` target
due to MSVC getting perplexed by `Support.java` due to our java source
rules. The change has no impact on the functioning of the target.

### DIFF
--- a/lcidlc/lcidlc.gyp
+++ b/lcidlc/lcidlc.gyp
@@ -75,13 +75,7 @@
     {
       'target_name': 'encode_support',
       'type': 'none',
-
-      'sources':
-			[
-				'src/Support.java',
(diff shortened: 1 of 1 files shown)


montegoulding:
@ali I'm wondering if this is the issue on travis as it breaks lcidlc builds here but not on vulcan

livecodeali:
@montegoulding Just pushed over my branch to check

livecodeali:
@montegoulding Heh, oops - these are already removed in develop it seems

montegoulding:
Aha! woops... I bet I did it too ;-)

PR #7097, opened 2019-06-10
[[ StandaloneBuilder ]] Use semver lib when parsing google play deps

livecodemichael:
The standalone builder has been updated to use semantic version
checking when including play aar dependencies. If multiple
extensions depend on the the same play service, the standalone
builder will use the semver library to find the correct version
(if any) to use.

Relies on https://github.com/livecode/livecode/pull/7096

### DIFF
--- a/Installer/package.txt
+++ b/Installer/package.txt
@@ -488,6 +488,8 @@ component Extensions
 		rfolder macosx:packaged_extensions/com.livecode.library.objectrepository
 	into [[ToolsFolder]]/Extensions place
 		rfolder macosx:packaged_extensions/com.livecode.library.native.mac.statusmenu
+	into [[ToolsFolder]]/Extensions place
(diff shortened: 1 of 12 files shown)


Installer/package.txt
```diff
@@ -488,6 +488,8 @@ component Extensions
 		rfolder macosx:packaged_extensions/com.livecode.library.objectrepository
 	into [[ToolsFolder]]/Extensions place
 		rfolder macosx:packaged_extensions/com.livecode.library.native.mac.statusmenu
+	into [[ToolsFolder]]/Extensions place
+		rfolder macosx:packaged_extensions/com.livecode.library.getsemver
```
montegoulding: @livecodemichael I think there's an extra `get` in the name here

livecodemichael: @montegoulding Oops! No idea where that came from.

PR #7100, opened 2019-06-12
[[ Bug ]] Fix project section capsule for mac x86_64

montegoulding:
This patch uses `MCcapsule` directly rather than using `MCExecutableFindSection`.


livecodepanos:
Tested the build from the `fixinstallercapsule` branch and the problem is fixed.

@livecode-vulcan review ok c2fd1e6

PR #7101, opened 2019-06-12
[[ SemverLibrary ]] Range intersection

livecodemichael:
The semever library has been extended to include range intersection functions. You can now check to see if any number of ranges intersect as well as get the range that defines the intersection.

As part of the changes, the range parsing command has been extended to support optionally expanding caret, tilde and x ranges into range conditions (e.g. ~1.2.3 --> >=1.2.3 <1.3.0).

### DIFF
--- a/extensions/script-libraries/semver/semver.livecodescript
+++ b/extensions/script-libraries/semver/semver.livecodescript
@@ -649,38 +649,298 @@ private function __SatisfiesRange pVersion, pRange
    return tResult
 end __SatisfiesRange
 
-private command __SemverSplitRange @xRange
-   replace "||" with " || " in xRange
-   repeat while xRange contains "  "
-      replace "  " with " " in xRange
-   end repeat
-   split xRange by space
-end __SemverSplitRange
+private function __IsX pVal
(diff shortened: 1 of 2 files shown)


extensions/script-libraries/semver/semver.livecodescript
```diff
@@ -649,38 +649,298 @@ private function __SatisfiesRange pVersion, pRange
...
+      put 0 into tExpandedRange[2]["patch"]
+      put "<" into tExpandedRange[2]["operator"]
+   end if
+   
+   return tExpandedRange
+end __ExpandXRanges
 
-command semverParseRange @xRange
-   __SemverSplitRange xRange
+command semverParseRange @xRange, pExpand
```
montegoulding, line 878: @livecodemichael this command should be documented if it's public

montegoulding: Aha! looks like that's my fault originally! Sorry do you mind fixing my error here and adding the docs?

montegoulding:
@livecodemichael could you bump the minor version. Would be somewhat ironic to not do so on a semver library ;-)

PR #7102, opened 2019-06-13
Revert "[[ Bug ]] Fix project section capsule for mac x86_64"

livecodepanos:
Reverts livecode/livecode#7100


livecodepanos:
Reverting this as it broke installers on Linux and Windows

PR #7119, opened 2019-06-24
[[ Bug 21183 ]] specialFolderPath("resources") for stacks with empty filename is not documented

researchware:


livecodepanos:
@researchware This needs a release note as well.

BTW the release notes should have a "positive" tone, and describe the fix, for example for this one you can use:

`# Ensure specialFolderPath("resources") for stacks with empty filename is documented`

instead of

`# specialFolderPath("resources") for stacks with empty filename is not documented`

PR #7125, opened 2019-07-18
[Bug 19157] Prevent treeview widget hang on non-printable characters

BerndN:


bwmilby:
The note should be renamed to `19157.md` to conform to the naming convention.

BerndN:
Thank you Brian, will do

PR #7127, opened 2019-07-23
Fixes the dbsqlite gyp file for building on linux

mwieder:
Fixes the dbsqlite gyp file for building on linux. Or more properly to avoid crashing on launch after building from source. This fix was proposed by @livecodefraser and it solves the problem for me.


mwieder:
From my limited understanding of this, it would appear that the team's build server is using a gcc version pre-5.1? I'm currently on 7.4.0. And that this patch instructs the compiler that the dbsqlite libraries are in C++03 ABI rather than C++11?

mwieder:
Should I resubmit this against develop-9.5 instead?

livecodepanos:
Oh I missed this PR! Nice, so this was the reason LC crashed for me when I built from source on Linux recently!

Thanks @mwieder and @livecodefraser :)

livecodepanos:
@mwieder I have rebased this against `develop-9.5`, luckily `develop` and `develop-9.5` have not diverged yet, so I was able to change the base from the GitHub web interface without conflicts

mwieder:
Am I really the only one attempting to build from source on a linux distro that hasn't already reached its end of life? I have to apply this patch every time I build, and then try to remember to remove it afterwards.

PR #7157, opened 2019-09-07
answer dialog should throw exception if no dialog stack exists

mwieder:
If there is no ask or answer stack in a standalone and an ask or answer command is issued, the command is ignored silently. This patch fixes that situation so that an exception is thrown and can be caught in a try/catch construct. Currently at the point of this patch t_success is still true from the previous tests and t_stack is nil, thus the next section of code never gets executed, t_success never gets set to false, and so the function exits without throwing the exception.


mwieder:
Seems to have failed for not having environment variables?

PR #7188, opened 2019-10-09
[TreeView] Allow case-insensitive sorting of tree view keys

livecodepanos:
This patch adds a new boolean property `sortCaseSensitive` to the treeView widget, which defaults to true. 

When true, the keys `AA`,`a`,`BB`,`b`,`CC`,`c` are sorted as : `AA`,`BB`,`CC`,`a`,`b`,`c`.

When false, they are sorted as : `a`,`AA`,`b`,`BB`,`c`,`CC`


extensions/widgets/treeview/treeview.lcb
```diff
@@ -1948,22 +1969,49 @@ private handler convertArrayToList(in pArray as Array, in pLevel as Integer, in
...
+
+	if not mSortCaseSensitive then
+		repeat for each element tKeyLower in tKeysTemp
+			repeat for each element tKey in tKeys
+				if the lower of tKey is tKeyLower then
+					push tKey onto back of tKeysSorted
+				end if
+			end repeat
+		end repeat
+	else
```
livecodeian, line 2008: Is this intended to be used where there may be values that differ only in the case of their characters? (e.g. "Hello" vs "hello") If so, this method won't work correctly as it will replace both values with whichever appears first in the list

PR #7192, opened 2019-10-11
[22103] Ensure effective textFont of chunk reports correct value

livecodepanos:


engine/src/exec-interface-field-chunk.cpp
```diff
@@ -2438,6 +2438,10 @@ void MCField::GetEffectiveTextFontOfCharChunk(MCExecContext& ctxt, uint32_t p_pa
 
     if (r_mixed)
         return;
+    
+    if(!MCStringIsEmpty(*t_value))
```
runrevmark, line 2442: Why is the check for the non-empty string necessary?

livecodepanos: I think the effective textFont of a *chunk* cannot be empty. It `*t_value` is nil or `kMCEmptyString` then the effective textFont of a chunk should be the effective textFont of the field (i.e. `GetEffectiveTextFont(ctxt, r_value);` from above)

runrevmark: Okay so - I don't think t_value can be nil :)

The problem here is an incorrect use of the GetCharPropCharChunk method I think... It is being called in 'effective'; mode - so the method should be:
```
MCAutoStringRef t_default;
GetEffectiveTextFont(ctxt, &t_default);
GetCharPropOfCharChunk< PodFieldPropType<MCStringRef> >(ctxt, this, p_part_id, si, ei, &MCBlock::GetTextFont, true, *t_default, r_mixed, r_value);
```
Here we are calling them method in question in 'effective' mode, which means it uses t_default as the value to use when a style run has the property unset. Of course, appropriate tests will determine whether this is correct :)

livecodepanos: I have tried this ^^^^ with the sample stack attached to the bug report and for the first line (where its textFont is not set - it is just the textfont of the field that is set) it returns empty for the effective textFont (and r_value is NULL in this case):

Screenshot 2019-10-14 at 12 31 31

PR #7195, opened 2019-10-15
[[ Bug 13085 ]] Add field equivalents of some paragraph properties

montegoulding:
This patch makes `leftIndent`, `rightIndent`, `spaceAbove` and `spaceBelow`
which were previously only line chunk properties properties of fields similar
to `firstIndent` etc. This allows them to be set for all paragraphs in a
field at once rather than setting the properties on individual paragraphs.


montegoulding:
Round trip to disk tests will merge up from [this PR](https://github.com/livecode/livecode/pull/7196) so we may want to merge that in and up before reviewing this. They pass locally though.

trevordevore:
@montegoulding and @runrevmark - When I saw this PR I thought of enhancement request [#11934](https://quality.livecode.com/show_bug.cgi?id=11934). Could an argument be made that the `spaceAbove`/`spaceBelow` of a paragraph should affect just the paragraph, and not every formatted line in the paragraph?

PR #7206, opened 2019-10-28
Update clickCharChunk.lcdoc

PaulMcClernan:
Removed extra "as"


PaulMcClernan:
Ah, I just realized that the problem is actually that the word "well" is missing, it should read "cleared at the next mouse click, as **well** as after some editing actions."

PR #7214, opened 2019-10-31
Correct Mac OS X native line endings to LF

bwmilby:


mwieder:
Thanks, @bwmilby. Looks good.

PR #7242, opened 2019-12-19
[Bugfix-22502] Clarify `CodeOfChar` docs

trevordevore:


peter-b (changes requested):
Hi @trevordevore,

Unicode codeunits are different from codepoints, and depend on the encoding of the string.

In LiveCode, strings are stored internally in (roughly) UTF-16 format.  UTF-16 is an encoding where each codeunit is 16 bits. Codepoints in the Basic Multilingual Plane fit into a single UTF-16 codeunit. For example, U+0046 LATIN CAPITAL LETTER F is UTF-16 encoded as `0x0046`.

However, codepoints in higher planes need multiple UTF-16 codeunits to represent them.  For example, U+1F1EB REGIONAL INDICATOR SYMBOL LETTER F is in the Supplementary Multilingual plane and is UTF-16 encoded as `0xD83C 0xDDEB`. In this case, the codeunit and codepoint are different.

So, unfortunately this is not correct:

> Use the `codeunit` property to target a specific `codepoint` within a multi-codepoint string.

Does `the code of codepoint 1 of tChar` do the right thing?

Best regards,

Peter

trevordevore:
@peter-b I made this update based on a conversation with @runrevmark. It looks like I misunderstood, however. I would suggest that someone from LiveCode with a better understanding of what the documentation should say submit their own PR.

PR #7260, opened 2020-02-06
make backslash appear in chrome browser

LCHeather:
Backslash in the "Wrapping" section is not escaped and currently appears as blank in Chrome. Have added the escape backslash to the existing backslash.

### DIFF
--- a/docs/dictionary/function/within.lcdoc
+++ b/docs/dictionary/function/within.lcdoc
@@ -64,13 +64,13 @@ borders.
 
 The expression
 
-    <point> is within the rect of <object(glossary)> 
+    point is within the rect of object 
 
 is equivalent to
 
-    within(<object(glossary)>, <point>)
+    within(object, point)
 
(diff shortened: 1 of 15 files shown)


runrevmark:
@LCHeather On what page online does this appear malformed?

I don't think the issue here is in the markdown - its in whatever we are using to turn the markdown into the HTML output (so fixing it here is the wrong place).

LCHeather:
Its in the docs, its not malformed its just blank :) Doesn't show the slash at all. Adding the escape slash fixes it in the main text but not in the code (when i tested in preview).

https://livecode.com/docs/9-5-0/extending-livecode/livecode-builder-style-guide/ <https://livecode.com/docs/9-5-0/extending-livecode/livecode-builder-style-guide/>

In the "Wrapping" section.


Heather Laine
Customer Services Manager
LiveCode Ltd
www.livecode.com

runrevmark:
@LCHeather I probably should have checked markdown syntax docs before commenting :) It seems you always need to escape backslashes - so this is correct :)

PR #7269, opened 2020-02-28
[bugfix-7948] Add docs note that revDeleteFolder can't delete defaultFolder

livecodealice:
Added a note in the documentation for revDeleteFolder that choosing a folder that is or contains the defaultFolder will fail to delete it.


livecodealice:
https://github.com/livecode/livecode/pull/7272

PR #7270, opened 2020-02-28
[Bugfix-11554] Remove inaccuracies from mobilePickDate docs

livecodealice:
Removed all mention of mobilePickDate taking input and outputting in seconds on Android as this is not the case.


livecodealice:
https://github.com/livecode/livecode/pull/7273

PR #7312, opened 2020-04-02
[OpenSSL] Update to 1.1.1f

livecodepanos:
With this patch, the prebuilts are built successfully in all platforms but Android.

Note: They are *NOT* tested yet, it is just that the builds complete with no error.

On Android, the error I get is:
`no NDK -gcc on $PATH at (eval 5) line 124.`

which seems to come from here:
https://github.com/openssl/openssl/blob/OpenSSL_1_1_1f/Configurations/15-android.conf#L124
Looks like the `$triarch` var is empty.
Note the this ^^ file is from the (most recent) OpenSSL 1.1.1f

In the version of OpenSSL we use (1.1.0g), this file (`Configurations/15-android.conf`) does not exist.
In fact, it seems that the android part of the file `Configurations/10-main.conf` which we use to configure openssl in 1.1.0g is removed, and a new file `Configurations/15-android.conf` is created.

Here is the diff when this happened:
https://github.com/openssl/openssl/commit/ebea0f3014c5f19573afc7a8746278b557e5a1f5


livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/7322

PR #7378, opened 2020-06-12
[22780] Ensure the compositorType property can be set to "none"

livecodepanos:
This patch ensures the compositorType property can be set to "none".
Previously this enumeration value was set to read-only, thus an error
was thrown when attempting to set the compositorType to "none"

Closes https://quality.livecode.com/show_bug.cgi?id=22780

### DIFF
--- a/docs/notes/bugfix-22780.md
+++ b/docs/notes/bugfix-22780.md
@@ -1,0 +1,1 @@
+# Ensure no error is thrown when setting the compositorType property to "none"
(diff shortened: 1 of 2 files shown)


runrevmark (commented):
So we need to check whether this was a 'bug' introduced in the refactor - the docs state `none` is returned if no compositor... And this change allows setting to none certainly, but it will still return empty in that case.

It might be just a docs bug if none has never figured as a compositor type!

PR #7401, opened 2020-07-20
[[ Bug 22840 ]] Activate IME in redraw

montegoulding:
This patch moves IME activation to screen redraw. This has the benefit of
limiting repeated IME activation/deactivation when setting field properties. It
additionally allows `lock screen` to be used to control IME activation.


runrevmark:
This is a change in behavior rather than a bugfix, so should be targetting develop.

PR #7434, opened 2020-09-16
Bugfix 22891

mwieder:


mwieder:
Ack! Not sure what happened here with the rebase.

PR #7437, opened 2020-09-16
Bugfix 15094

mwieder:
Implement specialFolderPath(documents) on Linux as "~/Documents"


mwieder:
@runrevmark I managed to wrangle git (at no small risk to life and limb) into fixing up the intermediate commits. I think this is ready now.

PR #7452, opened 2020-10-14
Fixed a bunch of typos

PaulMcClernan:


PaulMcClernan (commented):
Looks good to me

docs/dictionary/command/edit.lcdoc
```diff
@@ -25,7 +25,7 @@ edit the script of player "Show Me"
 
 Parameters:
 object:
-Any object reference. Usint the optional at syntax will pass the
+Any object reference. Use the optional at syntax will pass the
```
livecodepanos: This should be `Using`

livecodepanos:
Hello @PaulMcClernan 

This PR is almost ready. Some things to change:

Would it be possible to revert the last commit (since it is not just a correction of a typo as the other commits), and submit a new PR (separate) with this change?

Also, regarding the change itself, you should use `>*Note:*` instead of just `Note`, just after the `Description` part. For example see the description and the note of the `OnSave` handler:
```
Description:
Handler the <OnSave> message to execute code when a stack containing the widget is saved.
>*Note:* Access to most script object operations is not allowed whilst an <OnSave>
handler is running.
```

You need to do something similar in the new PR. 
 
Finally, the commit msg in the new PR should *just* say:

`Add note that the 'current' form of TheClickButton has not yet been implemented`

Thanks :)

livecodepanos:
Also, for the new PR make sure you branch off (and submit the PR against) `develop-9.6`

PaulMcClernan (commented):
Refixified

PaulMcClernan:
Git is such a p.i.a. 
Not sure how to remove a commit from a PR (using the web interface).
I guess I'll close this out, and resubmit another PR

PR #7455, opened 2020-10-19
Make 'working screenrect' to return the safe area on iOS

livecodepanos:


livecodepanos:
WIP because this results in black screen, even though the correct safe area is reported:

Screenshot 2020-10-19 at 13 14 57

livecodepanos:
Note that the safe area is actually the screen rect excluding the area of the top notch and the bottom area of the soft "home" button:

livecodepanos:
I believe the "black screen" problem is caused because the value of the member variable `workarea` (of struct `MCDisplay`) is used in other places in the engine as well, so these places do not like the `workarea` having the value of the safe area. Or because `bool MCScreenDC::platform_getdisplays(bool p_effective, MCDisplay *&r_displays, uint32_t &r_count)` is used in several other places apart from `MCInterfaceGetScreenRect`, and again, these places probably do not like the new value of `r_displays[0] -> workarea` which is now the safe area.

livecodepanos:
I have pushed some changes that fix the black screen issue. The `working screenrect` now reports the safe area, at any point it is called (i.e. either on `preopenstack` or later after the `opencard` is sent.

Screenshot 2020-10-20 at 15 33 33

PR #7492, opened 2020-12-11
[[ Bug 22994 ]] Fix crash when setting the startTime of a video on iOS

livecodepanos:
This patch fixes a crash when setting the startTime of a video on iOS


engine/src/mbliphoneplayer.mm
```diff
@@ -506,7 +506,7 @@ static void content_to_url(MCStringRef p_file, NSURL*& r_url)
 		m_start_time = CMTimeMake(p_time, 1000);
 
     if (m_player != nil)
-		m_player.currentItem.reversePlaybackEndTime = m_start_time;
+		SetCurrentTime(ctxt, p_time);
```
livecodeian, line 509: @livecodepanos `SetCurrentTime` just seeks to that position. the startTime should specify the position that the video begins playback from, including when looping

PR #7527, opened 2021-02-18
[[ Bug 22400 ]] Allow http connections on Android browser

livecodepanos:
This patch allows http connections if the user checks the appropriate checkbox in the Android standalone settings.

PR #7571, opened 2021-06-01
Revert "[[ Bug 22400 ]] Allow http connections on Android browser"

livecodepanos:
Reverts livecode/livecode#7527


livecodepanos:
Reverting this PR, since it was accidentally merged into `develop` instead of develop-9.6

PR #7579, opened 2021-06-09
[[ Bug 23193 ]] Ensure AppTrackingTransparency framework is included on iOS 14+

livecodepanos:
This patch ensures the AppTrackingTransparency framework is included if the version of the target_sdk is 14+
not only on simulator builds but also on device builds.

PR #7580, opened 2021-06-10
[[ Bug 23234 ]] Fix iOS standalone build error when running from source

livecodepanos:
This patch fixes the error:

```
Failed to open macho file at 
/../Xcodes/Xcode_12_5_0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ 
for reading: Too many levels of symbolic links
```

It seems there is a chance `g++` is not symlinked correctly, so use the full path to `clang++` instead.

Closes https://quality.livecode.com/show_bug.cgi?id=23234

### DIFF
--- a/ide-support/revsaveasiosstandalone.livecodescript
+++ b/ide-support/revsaveasiosstandalone.livecodescript
@@ -535,7 +535,11 @@ private command revSaveAsMobileStandaloneMain pStack, pAppBundle, pTarget, pSett
             -- MM-2013-09-23: [[ iOS7 Support ]] Use g++ instead of llvm-g++-4.2. XCode 5.0 uses llvm 5.0.
             -- g++ appears to be sym-linked to the appropriate compiler in all SDKS.
             -- SN-2015-02-19: [[ Bug 14625 ]] The minimum iOS version is bound to the architecture
-            get shell("g++ -arch " & tArch && "-miphoneos-version-min=" & tiPhoneMinVersion[tArch] && " -w " & quote & "@" & tLinkOptionsFile & quote)
+            
+            -- PM-2021-06-10: [[ Bug 23234 ]] g++ might not be symlinked correctly, so use full path to clang++ instead
+            local tCommand
+            put tSdkRoot & "/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++" into tCommand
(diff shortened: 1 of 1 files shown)


runrevmark (changes requested):
Why is this only an issue when running from source?

In any case, please file a BZ issue so it has an issue number and update commit message / PR appropriately.

livecodepanos:
@runrevmark Done :)

Does it need a release note, or is it considered "internal"?

runrevmark:
I think its esoteric enough that it doesn't need a release note - although how is g++ not symlinked correctly? Is it something about your setup? Something one of our internals setup scripts does?

livecodepanos:
I do not think it is my setup, since @montegoulding experienced the same issue a couple of months ago with Xcode 12.4. It might be because of a change in recent Xcodes, since I see this error with Xcode 12.5, but not with Xcode 11.3 in a different Mac.

My understanding is that somehow `g++` symlinks to `Xcode_12_5_0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++`, but the latter is a symlink as well (to `Xcode_12_5_0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang`), hence the `too many symbolic links` error.

runrevmark:
Okay so it sounds like it isn't necessarily a build from source issue then - isn't the modified code in code that runs in a built IDE too?

livecodepanos:
The modified code is in the S/B, which does run in a built IDE too. It does not seem to happen with Xcode 12.1, otherwise people would have noticed. We have not released a build that supports Xcode 12.4 or Xcode 12.5 so far, so I am not entirely sure if this is a build from source issue or a general Xcode12.4+ issue. I think I will be able to tell if this is the case when #7579 is merged and we kick off a new build. So I am marking this PR as WIP until then

runrevmark:
Ah! Okay well if it is an xcode 12.4/12.5 issue then it doesn't need a release note as it won't have affected anyone :)

runrevmark:
@livecodepanos Is this definitely an issue we need to resolve for the new Xcode update?

livecodepanos:
@runrevmark I have checked with a built version of 9.6.3 rc1 from staging, with Xcode 12.5, and this issue is not present. So it seems it happens only when building from source.

PR #7589, opened 2021-06-28
[Bugfix 23247] Remove references to HTML5 in post docs

livecodealice:
Removed HTML5 notes and HTML5 from the platforms in the post dictionary entry as the post command does not work on HTML5.


livecodepanos:
Replaced by https://github.com/livecode/livecode/pull/7590

